Issue #120, opened 2018-02-11
System requirements detection


Sherlouk:
I'll start of by saying I love what you're building here and I can't wait to see it progress! 🙌 

In the same way the launch script informs the user if certain environment variables are not defined, it would be handy if the program could ensure the system meets the necessary requirements

Currently I get the following error when trying to run locally:

```
bundler: failed to load command: rackup (/usr/local/bin/rackup)
NoMethodError: undefined method `provider_credentials' for nil:NilClass
```

This is in fact caused by the fact I don't have Ruby 2.3 (or newer), an easy check to run!

Same goes for validating OS, Xcode Version, Command Line Tools and any other requirements

**Update**

Follow up work around detecting Xcode installation, and more

taquitos:
We're super focused on getting this up and running that we haven't spent a ton of time validating that sort of thing. Is that something you'd like to help us with?

Sherlouk:
I'm not super hot on ruby but I'll definitely give it a go!

taquitos:
Hopefully you'll start to ❤️it 😄

KrauseFx:
This is merged, nice work 👍

Sherlouk:
For what it's worth, I had left it open intentionally to handle further requirements such as Xcode/what not - but we can recreate an issue when it becomes more of an issue (Ruby was throwing weird errors hence prioritised that)

Thanks though!

KrauseFx:
Alright, good call, I re-opened, and added a placeholder for more information + context on what's required to get this card to be shipped 👍

PR #121, opened 2018-02-11
Add pre-check for Ruby version

Sherlouk:
Thought I'd keep it simple to dip my toes in the water, took logic from the [existing Fastlane action](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/ruby_version.rb) and what was already in the launch script

Not sure if there's a better location to store the constant?

### DIFF
--- a/launch.rb
+++ b/launch.rb
@@ -6,6 +6,7 @@ module FastlaneCI
   class Launch
     def self.take_off
       verify_dependencies
+      verify_system_requirements
       load_dot_env
       verify_env_variables
       setup_threads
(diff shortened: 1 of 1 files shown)


taquitos:
This seems totally reasonable to me. I don't think we use this constant in other places yet, so keeping it inline is 👌 
🎈:goat:

Sherlouk:
Thanks for merging! 🙌 

Linking #120 as I forgot to, should leave it open though to track other system requirements 😄

taquitos:
Thank you for helping us out here!

KrauseFx:
🚀

Issue #89, opened 2018-01-30
`rspec` tests are failing, as they assume `~/.fastlane/ci` to exist and have the repos checked out it seems


KrauseFx:
Running `bundle exec rspec` fails, as it seems like it assumes the config folders and files to already exist
```
Git::GitExecuteError:
  git  clone '--recursive' '--' '' '/Users/fkrause/.fastlane/ci/fastlane-ci-config'  2>&1:fatal: The empty string is not a valid path
# ./shared/models/git_repo.rb:220:in `clone'
# ./shared/models/git_repo.rb:81:in `initialize'
# ./services/config_data_sources/git_config_data_source.rb:29:in `new'
# ./services/config_data_sources/git_config_data_source.rb:29:in `initialize'
# ./services/user_service.rb:64:in `new'
# ./services/user_service.rb:64:in `trigger_initial_ci_setup'
# ./services/user_service.rb:46:in `login'
# ./fastlane_app.rb:65:in `<class:FastlaneApp>'
# ./fastlane_app.rb:31:in `<module:FastlaneCI>'
# ./fastlane_app.rb:23:in `<top (required)>'
# ./spec/spec_helper.rb:6:in `require'
# ./spec/spec_helper.rb:6:in `<top (required)>'
# ./spec/app_spec.rb:1:in `require'
# ./spec/app_spec.rb:1:in `<top (required)>'
```

KrauseFx:
Fixed with https://github.com/fastlane/ci/pull/122

PR #122, opened 2018-02-12
Update string_encrypter_spec to stub encryption key

KrauseFx:
Is that by design that the encryption key can be anything for the tests to be green?

If so, feel free to merge. If not, we need to investigate

Fixes https://github.com/fastlane/ci/issues/89


minuscorp:
Yep, that's the requirement, the missing ENV.

PR #126, opened 2018-02-13
Create a single queue for all git repo things

taquitos:
the git gem is not meant to be threadsafe
Use a queue


KrauseFx:
is that a mono queue? :trollface:

taquitos:
best queue

PR #127, opened 2018-02-13
Implement ensure blocks for repo actions

taquitos:


shared/models/git_repo.rb
```diff
@@ -166,13 +188,12 @@ def temporary_git_storage
     # Responsible for using the auth token to be able to push/pull changes
     # from git remote
     def setup_auth(repo_auth: self.repo_auth)
+      self.temporary_storage_path = File.join(self.temporary_git_storage, "git-auth-#{SecureRandom.uuid}")
```
taquitos: @KrauseFx want to double check this is ok. Instead returning a path from `setup_auth` I'm setting it as a temporary instance variable, this way I can reference the variable in an `ensure_block`

taquitos: I prefer to not use attributes for this sort of thing, but in order to use ensure blocks, I had to know where to get the value of the variable before executing the block, so referencing an attribute works.

PR #133, opened 2018-02-14
ProjectService setup.

minuscorp:
So this PR tries to cover the following objectives:

- Generalize the `JSONDataSource` using a mixin that can be included by any `SomeModelJSONDataSource`.

- Using this pattern, all models that have a `JSONDataSource`, should include the mixin.

- At the end of this, we can achieve that `ProjectService` is created, using `ProjectJSONDataSource`. 

I have some questions regarding how this can be achieved but we can break this in little pieces so it’s easier to work on.

### DIFF
--- a/features/project/views/project.erb
+++ b/features/project/views/project.erb
@@ -40,7 +40,7 @@
               <% end %>
             </td>
             <td style="text-align: center;"><%= build.number %></td>
-            <td><%= build.timestamp %></td>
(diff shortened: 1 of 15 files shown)


minuscorp:
There's some issue with CLA @KrauseFx ... But after this is solved I think this is good to go into review! @taquitos 🚀

KrauseFx:
Don't worry about the CLA, you signed it, and I'm the co-author (one commit of mine is included), so we're all good

PR #138, opened 2018-02-15
Convert DateTime to seconds_since_epoch and back.

### LINKED ISSUE #110: build JSON file, use timestamp in seconds since 1970 instead of full date format
In the `[build_number].json` it seems like currently we store `"2018-01-30 16:24:23 -0500"` for time stamps, however I think we should use `1516639255` instead 👍

minuscorp:
Closes #110

### DIFF
--- a/services/data_sources/json_build_data_source.rb
+++ b/services/data_sources/json_build_data_source.rb
@@ -8,6 +8,20 @@ module FastlaneCI
   # Mixin the JSONConvertible class for Build
   class Build
     include FastlaneCI::JSONConvertible
+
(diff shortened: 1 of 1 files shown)


minuscorp:
I couldn't make a local build yet :(

minuscorp:
As far as I can get, the first build (I don't know why) it's always set with the old timestamp format. The following builds are written correctly using the desired format. I don't know if somewhere on the code the first build is hard-coded with a custom hash that does not use `JSONConvertible` but I couldn't find it.

PR #139, opened 2018-02-15
[WIP] fastlane integration

KrauseFx:
Early work in progress, this PR will contain the communication between `fastlane.ci` and good old [fastlane](https://fastlane.tools).

The idea is to use the custom UI layer, that was introduced a [few years ago](https://github.com/fastlane-old/fastlane_core/pull/53), to think ahead for this exact scenario, future much?

Open tasks
- [x] Connect the Ruby classes
- [ ] Verify it works with plugins
- [ ] Platform support, how do we store and pass the platform
- [ ] Plan and build architecture on how to render output to the user (more info https://github.com/fastlane/ci/issues/147)

Fixes https://github.com/fastlane/ci/issues/24
Depends on https://github.com/fastlane/fastlane/pull/11857


minuscorp:
What if, idea is coming, we generalize the output so (using a mixin) (sorry I come from the iOS world and I love protocol oriented programming) there's no need for a *file_path* so there is a simple pipe out of strings with some metadata (defined by an interface or other mixin so we can't mess up with the attributes or the methods) so that output might be a terminal, a YML file, a JSON, an HTML... Don't know if you see where I'm going😜. I can help on this is you want during the coming week, as I have finished what I had on queue!

KrauseFx:
Great idea. I figured we kind of have that with https://github.com/fastlane/ci/pull/139/files#diff-d98d72f5d471fb004eb4477c06539ba2R12 also, and have the log file as an extra. What do you think? 

The block would get a real-time call on every new line

minuscorp:
My main concern was that there's a `file_path` dependency. So it kind of reflects a file dependency, which I think that it wouldn't be required always (I'm thinking of, for example, a socket output, or a redis, mongodb, etc storages) so the pipe is just a output of some data (I.e., a string or an HTML, etc) to a source.
For example:
1) data log string comes in
2) HTML transformer generates the data from 1)
3) the _outputter_ takes the HTML data and pushes it to a S3 instance.
(Just take it as an example, but the design into one source, one transformer and one storer would be super nice in order to adapt to future requirements or different needs that'd come in the future).

KrauseFx:
Great point, will refactor to account for any kind of "output" 👍 Thanks @minuscorp

Issue #84, opened 2018-01-30
general Xcode/tool problems affecting the whole server


taquitos:
If there are Xcode or other tool issues that are impacting the server as a whole, we need to figure out the best way to notify people
- [x] Dashboard banner?
- [x] Pager-duty?
- [x] Github Issue creation?
- [x] Attempt to fix common problems (xcode-install)?

KrauseFx:
I think email would probably the most generic, most common one we could start with. And then we can go from there. And agreed, the first integration (e.g. email) should already be built in a way to allow developers to hook in their own systems, maybe just allowing running a shell script in the beginning

taquitos:
I was even thinking, just for a MVP of this issue, persisting the notifications somewhere and getting a dashboard banner that you can acknowledge/dismiss

KrauseFx:
Yeah, that's a great idea 👍 Sending emails is still a pain nowadays, so your approach might be the best one to get started with

PR #142, opened 2018-02-15
#84 Notifications Dashboard.

AndrewMcBurney:
The following is a PR for issue https://github.com/fastlane/ci/issues/84.

### Notification Structure

Notifications have the following attributes:

| Attribute    | Type     | Description                                                                         |
| :---         | :---     | :---                                                                                |
| `id`         | `String` | primary key for the notification                                                    |
| `priority`   | `String` | notification priority (i.e., `LOW`, `MEDIUM`, `HIGH`)                               |
| `type`       | `String` | the type of the notification                                                    |
| `name`       | `String` | name for the notification                                                    |
| `message`    | `String` | notification message body                                                           |
| `created_at` | `String` | a timestamp cast as ``String`` pertaining to when the notification was created      |
| `updated_at` | `String` | a timestamp cast as ``String`` pertaining to when the notification was last updated |

### Usage
_Assuming the requests are authenticated._

**Creating a notification**
```bash
curl -d '{"priority":"HIGH", "name":"Build failure on host 127.0.0.1", "message": "ERROR: A build failure occured due to:\n- not enough emojis 🚀🚀🚀"}' -H "Content-Type: application/json" -X POST http://localhost:8080/notifications/create
```

**Updating a notification**
```bash
curl -d '{"id":"test_id", "priority":"LOW", "name":"Build failure on host 127.0.0.1", "message": "WARNING: Build at risk of failing due to:\n- not enough emojis 🚀🚀🚀"}' -H "Content-Type: application/json" -X POST http://localhost:8080/notifications/update
```

### Dashboard

screen shot 2018-02-25 at 3 26 14 pm

> A user may delete notifications from the notifications dashboard UI

### Enhancements

1) Add a way to delete notifications without modifying the JSON file manually
- [x] `POST` action to delete notification by `name` (i.e., the primary key to uniquely specify the notification)

2) UI love :heart:
- [x] Better formatting of notifications table
- [x] Add button on UI to delete notification


services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,137 @@
...
+    # @return [String]
+    def notifications_file_path(path: "notifications.json")
+      File.join(json_folder_path, path)
+    end
+
+    # Reloads the notifications from the data source
+    #
+    # @return [nil]
+    def reload_notifications
+      JSONNOtificationDataSource.file_semaphore.synchronize do
```
KrauseFx: `JSONNO` should be `JSONNo`

shared/models/notification.rb
```diff
@@ -0,0 +1,69 @@
...
+    # A way to uniquely specify a notification
+    #
+    # @return [String]
+    attr_accessor :primary_key
+
+    # Instantiates a new `Notification` model object
+    #
+    # @param  [String] id
+    # @param  [String] priority
+    # @param  [String] name
```
KrauseFx: Could a message be user specific also? I could imagine things like:
- Invalid scope for your GitHub user account
- [Maybe] a build you triggered has failed
Maybe there is more in the future?

spec/services/notification_service_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+    }
+  end
+end
+
+require "ripper"
+
+# Monkey patch to fix issue with HEREDOC in rspec for Ruby 2.3
+#
+# https://github.com/rspec/rspec-core/issues/2163#issuecomment-193657248
+Ripper::Lexer.class_eval do
```
AndrewMcBurney: I can use the `<<-` version of the HEREDOC if you'd prefer I don't monkey patch in the test suite too. Please let me know if you want me to make changes.

shared/models/notification.rb
AndrewMcBurney: I was thinking of associating notifications to users by persisting the user `email` in `notifications.json`. That way the `email` attribute acts as a foreign key. Does this sound like an okay approach, @taquitos. Felix said you wrote the user-specific logic, so I wanted to run the idea by you.

I saw `email.downcase` was used for comparisons when determining if a user already exists - so it acts as a primary key for the `User` model.

snatchev:
This is pretty sweet! I just wanted to raise up the advantage of using a slightly different file encoding for the notifications JSON file. The problem with have a single JSON Array object in a file is that only one process may add records and it must replace the entire contents every time it does so. Another simple alternative would be to have each JSON record be newline-delimited such as these examples https://en.wikipedia.org/wiki/JSON_streaming

The benefit is that multiple process can do appending operations to file without having to replace the entire contents.

KrauseFx:
> I also wonder if it makes sense to introduce a dependency on sqlite if we are going to be creating any kind of persistent data storage.

We already use a file-based persistence layer currently, which could potentially be replaced by a database. I haven't had a chance to review this particular PR, but I'd think it stores the notification in the same git repo where we store the build information, project metadata, etc.

minuscorp:
I think you'd benefit of the proposal https://github.com/fastlane/TaskQueue/issues/3, where you'll have actual feedback from your tasks, whether they finished correctly or not. And maybe, this requests should have REST representation. We haven't thought of adding a `API` to the application, so js and other async processes can use it.

taquitos:
Heyo @AndrewMcBurney, how is this PR going? Anything more you need before we can :shipit:? Is it ready for review?

AndrewMcBurney:
@taquitos I think we're good for review once I make one minor fix. I'll do that now and let you know once it's 💯

features/notifications/notifications_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+module FastlaneCI
+  # CRUD controller for handling and displaying notifications to user
+  class NotificationsController < AuthenticatedControllerBase
+    HOME = "/notifications"
+
+    # Instantiates a new `Notification Controller` with a task queue to process
+    # requests
+    #
+    # @param  [Sinatra::Base] app
+    # @return [nil]
```
taquitos: we probably don't need all these `@return [nil]`, unless there is something that is enabled by having it here.

features/notifications/notifications_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+    def initialize(app = nil)
+      @task_queue = TaskQueue::TaskQueue.new(name: "notifications")
```
taquitos: I know we're not super consistent on this, but maybe we should standardize the order in which we call `super()`
Generally speaking, I would opt for calling `super()` before changing any state in the current initializer.

services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,162 @@
...
+        git_repo.commit_changes!
+      end
+    end
+
+    # Returns `true` if the notification exists in the in-memory notifications object
+    #
+    # @param  [String] name
+    # @return [Boolean]
+    def notification_exist?(name: nil)
+      notification = @notifications.select { |n| n.primary_key == Notification.make_primary_key(name) }.first
```
taquitos: we might need to synchronize on this too, depending on the behavior of `.select`, I know `@notifications` could be updated while this is iterating.

services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,162 @@
...
+    # the notification exists
+    #
+    # @param  [Notification] notification
+    # @return [nil]
+    def update_notification!(notification: nil)
+      notification.updated_at = Time.now
+      notification_index = nil
+      existing_notification = nil
+
+      @notifications.each.with_index do |old_notification, index|
```
taquitos: same here re: synchronization, what do you think?

services/notification_service.rb
```diff
@@ -0,0 +1,65 @@
...
+    # @return [nil]
+    def update_notification!(notification: nil)
+      notification_data_source.update_notification!(notification: notification)
+    end
+
+    # Deletes a notification if the 'primary key' exists
+    #
+    # @param  [String] name
+    # @return [nil]
+    def delete_notification!(name: nil)
```
taquitos: Shouldn't this be using the `id` instead of `name`? (`so we're not open to ID guessing attacks`)

shared/models/notification.rb
taquitos: Yeah, that's good for determining if a user already exists, compare all the emails they want to use.
But, for comparing anywhere else, we should use the `id` as the foreign key.

shared/models/notification.rb
```diff
@@ -0,0 +1,70 @@
...
+      self.name = name
+      self.message = message
+      self.created_at = created_at || Time.now.to_s
+      self.updated_at = updated_at || Time.now.to_s
+    end
+
+    # A way to uniquely specify a notification
+    #
+    # @return [String]
+    def primary_key
```
taquitos: This is interesting. Instead of `make_primary_key` maybe we need to define a contract for notifications. As in, if you wish to send a notification that should be deduplicated, you need to figure out how to scope it so that we can apply deduplication logic to it.

I say this because I'm scared that we're hiding the logic of how a primary key works behind the `name` variable. If you create two notifications named the same thing, they are seen as the same, but they might not be.

Instead, I'd rather register my own primary-key generator with the notification system for each module I build, `iff` I need deduplication.

Also, I think notifications should have types.
`acknowledgement_required` and `disappearing` (like a Toast in android)

And then, because of this they should also have `acknowledged` state, so that if we have deduplication logic, it will only dedupe records that are still `acknowledged? == false`, that way if we do have a reoccurring notification, it will continue to notify when the user dismisses the notification.

Does this make sense?

Also, I think it would really help to make `use cases` around how we expect to use notifications so we can explore the situations we want to support 😄

taquitos: BTW, I'm happy to work on those use cases, it might be good experience for you to flesh them out, just let me know  what you want to do 🚀

spec/fixtures/files/notifications.json
```diff
@@ -0,0 +1,2 @@
+[
```
taquitos: best fixture

features/notifications/notifications_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+      super(app)
+    end
+
+    # Renders the notifications dashboard, displaying a table of all notifications
+    #
+    # @return [nil]
+    get HOME do
+      notifications = Services.notification_service.notification_data_source.notifications
```
taquitos: We can see all the notifications here, maybe we should scope them by user

spec/services/notification_service_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+
+      it "writes to the `notifications.json` file" do
+        File.should_receive(:write)
+        notification_service.update_notification!(notification: notification)
+      end
+    end
+  end
+
+  def json_notification_string
+    <<~JSON
```
taquitos: I think this should go in a fixture file

spec/services/notification_service_spec.rb
taquitos: This comment: https://github.com/rspec/rspec-core/issues/2163#issuecomment-214955288
says you might be able to do something like:
```
str = <<~SQUIGGLY.strip
  Your String
SQUIGGLY
```
if it works, I'd prefer that over the monkey patch. If it doesn't work, then I think keeping the monkey patch is ok.

spec/stub_helpers.rb
```diff
@@ -4,6 +4,8 @@ module StubHelpers
   include HelperFunctions
 
   def stub_git_repos
+    # TODO: refactor GitRepo so that initialize is non-blocking if repo isn't valid
```
taquitos: I think you can remove this comment and turn it into a github issue 🚀

features/notifications/notifications_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+      redirect HOME
+    end
+
+    private
+
+    # Adds a block of code to the notifications task queue
+    #
+    # @param  [Proc] block
+    # @return [nil]
+    def add_to_task_queue(&block)
```
taquitos: ❤️

services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,162 @@
...
+  # Data source for notifications backed by JSON
+  class JSONNotificationDataSource < NotificationDataSource
+    include FastlaneCI::JSONDataSource
+    include FastlaneCI::Logging
+
+    class << self
+      attr_accessor :file_semaphore
+    end
+
+    # Can't have us reading and writing to a file at the same time
```
taquitos: best comment

services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,162 @@
...
+    # @return [nil]
+    def delete_notification!(name: nil)
+      primary_key = Notification.make_primary_key(name)
+      self.notifications = @notifications.delete_if { |notification| notification.primary_key == primary_key }
+    end
+
+    private
+
+    # @return [FastlaneCI::GitRepo]
+    attr_accessor :git_repo
```
taquitos: let's move the `attr_accessor` declarations up to the top

minuscorp:
What about the `commit_changes` method, whose reasonability is it? The service or the data source? I bring it here but it is widely used throughout the app

services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,162 @@
...
+
+        return JSON.parse(File.read(path)).map(&Notification.method(:from_json!))
+      end
+    end
+
+    # Writes the notifications array to the git repo as JSON and commit them as
+    # the CI user
+    #
+    # @param  [Array[Notification]] notifications
+    def notifications=(notifications)
```
snatchev: Small API nitpick: As an API consumer of this class, I would be surprised that calling a setter method results in an IO operation. I would suggest an alternative of having a `persist_notifications` (or something like that) private method that can be called at the end of methods such as `update_notification!`. I would feel more idiomatic ruby to have `notification` getter and setter operate on the `@notification` instance variable. Just a suggestion, tho!

services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,162 @@
...
+      end
+    end
+
+    # Returns `true` if the notification exists in the in-memory notifications object
+    #
+    # @param  [String] id
+    # @return [Boolean]
+    def notification_exist?(id: nil)
+      JSONNotificationDataSource.file_semaphore.synchronize do
+        existing_notification = @notifications.select { |notification| notification.id == id }.first
```
snatchev: This can be re-written using `@notifications.any?` instead of `select`+`first`+`nil?`

services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,162 @@
...
+            notification_index = index
+            existing_notification = old_notification
+            break
+          end
+        end
+      end
+
+      if existing_notification.nil?
+        error_message = "Couldn't update notification #{notification.name} because it doesn't exist"
+        logger.debug(error_message)
```
snatchev: looks like a perfect case for `logger.error` :)

services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,162 @@
...
+
+      JSONNotificationDataSource.file_semaphore.synchronize do
+        @notifications.each.with_index do |old_notification, index|
+          if old_notification.id == notification.id
+            notification_index = index
+            existing_notification = old_notification
+            break
+          end
+        end
+      end
```
snatchev: Not clear why this semaphore exists here. It doesn't look like it's wrapping any IO

services/data_sources/json_notification_data_source.rb
taquitos: This is a great point. We might want to consider putting together a best practices guide for this sort of thing because I've actually run into it a couple of times.

services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,162 @@
...
+      if existing_notification.nil?
+        error_message = "Couldn't update notification #{notification.name} because it doesn't exist"
+        logger.debug(error_message)
+        raise error_message
+      else
+        @notifications[notification_index] = notification
+        self.notifications = @notifications
+        logger.debug("Updating notification #{existing_notification.name}, writing out notifications.json to #{notifications_file_path}")
+      end
+    end
```
snatchev: I think this method is a little hard to read. I have a _personal_ aversion to conditionals and temporary variables :) so this is more of a suggestion. I think it can be distilled into something like this:

```ruby
# guard against a non-operation
unless @notifications.any?{|n| n.id == notification.id}
  raise "Couldn't update notification #{notification.name} because it doesn't exist"
end

JSONNotificationDataSource.file_semaphore.synchronize do
  @notifications.map! do |old|
    if old.id == notification.id
      notification
    else
      old
    end
  end
  # todo: persist @notifications here
end
```

services/data_sources/json_notification_data_source.rb
AndrewMcBurney: Ahh, yeah. I tried to keep it consistent with the code we currently have in the `json_project_data_source.rb` [1](https://github.com/fastlane/ci/blob/master/services/config_data_sources/json_project_data_source.rb#L84-L94), [2](https://github.com/fastlane/ci/blob/master/services/config_data_sources/json_project_data_source.rb#L116-L120).

That being said, I am a fan of the private `persist_notifications` method you're suggesting. I agree it feels more like idiomatic ruby.

There's a lot of shared logic in both the `json_project_data_source.rb` and `json_notification_data_source.rb`, so I think it might also be beneficial to abstract the shared logic in both of those classes to an abstract base class. However, the thing keeping me from making those changes is that it might be a premature abstraction, or might be suited for another PR.

Thoughts?

services/data_sources/json_notification_data_source.rb
AndrewMcBurney: Ah good call 👍

services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,162 @@
...
+    # @param  [String] path
+    # @return [String]
+    def notifications_file_path(path: "notifications.json")
+      git_repo.file_path("notifications.json")
+    end
+
+    # Reloads the notifications from the data source
+    def reload_notifications
+      JSONNotificationDataSource.file_semaphore.synchronize do
+        @notifications = []
```
snatchev: this looks un-necessary

services/data_sources/json_notification_data_source.rb
AndrewMcBurney: Again, I was trying to keep consistency with the [`json_project_data_source`](https://github.com/fastlane/ci/blob/master/services/config_data_sources/json_project_data_source.rb#L176-L184) implementation.

If you'd like, I could change it in both locations.

snatchev: @AndrewMcBurney Thanks for pointing that out. A very valid reason! My vote would be to keep consistency.

services/data_sources/json_notification_data_source.rb
AndrewMcBurney: Oops, that was my bad 😳

features/notifications/notifications_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+  class NotificationsController < AuthenticatedControllerBase
+    HOME = "/notifications"
+
+    # Instantiates a new `Notification Controller` with a task queue to process
+    # requests
+    #
+    # @param  [Sinatra::Base] app
+    def initialize(app = nil)
+      super(app)
+      @task_queue = TaskQueue::TaskQueue.new(name: "notifications")
```
taquitos: Let's push this down one level into a service. I think it's possible that `@task_queue` could be garbage collected due to the lifecycle of `Sinatra::Base`
Generally speaking, we shouldn't be using `TaskQueue` in these controllers because of that. Also, any time we want a task queue in a controller, it might be a sign that we actually just want to call some async api on a server 😄

features/notifications/notifications_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      user_notifications = notifications.select { |notification| notification.user_id == user.id }
+      locals = { notifications: user_notifications, title: "Notifications" }
+      erb(:dashboard, locals: locals, layout: FastlaneCI.default_layout)
+    end
+
+    # Creates a new notification
+    post "#{HOME}/create" do
+      add_to_task_queue do
+        payload = notification_params(request)
+        Services.notification_service.create_notification!(payload)
```
taquitos: Yeah, like here, I think we actually want ` Services.notification_service.create_notification!` to handle the async dispatching

services/data_sources/json_notification_data_source.rb
```diff
@@ -0,0 +1,159 @@
...
+    # @param  [Any] params
+    def after_creation(**params)
+      if params.nil?
+        raise "Either user or a provider credential is mandatory."
+      else
+        if !params[:user] && !params[:provider_credential]
+          raise "Either user or a provider credential is mandatory."
+        else
+          params[:provider_credential] ||= params[:user].provider_credential(type: ProviderCredential::PROVIDER_CREDENTIAL_TYPES[:github])
+          @git_repo = FastlaneCI::GitRepo.new(git_config: self.json_folder_path, provider_credential: params[:provider_credential])
```
taquitos: Thinking about this, I don't think we even want to store this in git.
Here's why, I'm going through the use cases of why we would want persisted notifications:
1. Server restarts
1. User doesn't acknowledge notifications and logs out

So that's great, we just need persistence to handle these cases. The reasons why we want to store it in git?
1. Server crashes and we want to replace it

We haven't really discussed all the places where we want notifications, but the most common ones I can think of:
1. something wrong with tooling (Xcode, whatever)
1. Bot's access to git is messed up
1. Build failed
1. Yey, something good happened (success message)

All of these are either specific to the build machine or, will be triggered again in the future so a new notification will be created. 

So, tldr; I don't think there is any utility in storing our notifications in git. We should just keep it locally somewhere and be ok with losing the notifications.

KrauseFx: 👍 Especially for the MVP, this should be good enough

services/data_sources/json_notification_data_source.rb
taquitos: I would put that in another PR, and keep the focus of this one 💯

services/data_sources/json_user_data_source.rb
```diff
@@ -100,12 +100,7 @@ def login(email: nil, password: nil)
 
     # just check to see if we have a user with that email...
     def user_exist?(email: nil)
-      user = self.users.select { |existing_user| existing_user.email.casecmp(email.downcase).zero? }.first
-      if user.nil?
-        return false
-      else
-        return true
-      end
+      return self.users.any? { |existing_user| existing_user.email.casecmp(email.downcase).zero? }
```
taquitos: much nicer looking here :)

shared/models/notification.rb
```diff
@@ -0,0 +1,66 @@
...
+  class Notification
+    # Is a UDID so we're not open to ID guessing attacks
+    #
+    # @return [String]
+    attr_accessor :id
+
+    # The relative priority of a notification
+    #
+    # @return [String]
+    attr_accessor :priority
```
taquitos: Maybe we should define priorities using an enum, kinda like:
```ruby
PRIORITIES = {
      urgent: "urgent",
      normal: "normal",
      success: "success"
    }
```
Would we want any other priorities?
I can see `urgent` being used for stuff that must be fixed (and could potentially page people)
vs `normal` which is like `this build failed`, and `success` which is reserved for success messages.

shared/models/git_repo.rb
```diff
@@ -223,7 +223,13 @@ def setup_auth(repo_auth: self.repo_auth)
       self.temporary_storage_path = File.join(self.temporary_git_storage, "git-auth-#{SecureRandom.uuid}")
       # More details: https://git-scm.com/book/en/v2/Git-Tools-Credential-Storage
       local_repo_path = self.git_config.local_repo_path
-      FileUtils.mkdir_p(local_repo_path) unless File.directory?(local_repo_path)
+
+      # Creates the `local_repo_path` directory and `notifications/` directory
+      # if they do not exist
+      unless File.directory?(local_repo_path)
+        FileUtils.mkdir_p(local_repo_path)
+        FileUtils.mkdir_p(File.join(local_repo_path, "notifications"))
```
taquitos: I don't know if `notifications` should go here, that directory is totally separate from any `git_repo`, right?

AndrewMcBurney: Ahh, yep! The `local_repo_path` is `~/fastlane/ci/`. So the notifications would be stored in `~/fastlane/ci/notifications`, separate from the configuration repo. 🙂

AndrewMcBurney: Oh shoot, I just lied. I want the `@containing_path` 😳

services/services.rb
```diff
@@ -67,6 +67,15 @@ def self.user_service
       )
     end
 
+    # Start up a UserService from our JSONUserDataSource
+    def self.notification_service
+      @_notification_service ||= FastlaneCI::NotificationService.new(
+        notification_data_source: JSONNotificationDataSource.create(
+          File.expand_path("..", ci_config_git_repo_path)
```
taquitos: In the future, I think we should create like a `ci_data_home_path` which so we don't need to use `File.expand_path("..", ci_config_git_repo_path)`
but not now, just in the future :)

launch.rb
```diff
@@ -16,6 +16,7 @@ def self.take_off
       verify_system_requirements
       load_dot_env
       verify_env_variables
+      write_configuration_directories
```
taquitos: 💯

PR #151, opened 2018-02-20
Create, update, delete and read Projects.

minuscorp:
How do we handle the ci-config repo changes? (discussion here?)

- [x] Refactor endpoints to be wherever they must be.

- [ ] Provide an API to delete existing projects (and their already done builds?).

> To make it open, we think we're going to do the following: there're two ways of deleting a project, 1) virtually, the project gets disabled (or another param) and all its build data is preserved. 2) physically, the project gets deleted from the data source and all build data and artifacts are deleted.

- [x] Provide an API to commit and push changes on the `ci-config` repo easily.

> Whose responsibility should be the operations over the `ci-config` repo? the data source? the service?

PR #153, opened 2018-02-21
Changed `project_controller.rb` to use named parameters over `*`.

AndrewMcBurney:
Uses named parameters instead of `*` glob.


minuscorp:
As I have refactored the whole controller, would you mind to commit on my PR (can you do that?) or I just use your changes on my PR #151

AndrewMcBurney:
@minuscorp - sounds good, I'll create a PR to your branch with the changes 🙂

Issue #52, opened 2018-01-25
Find a way to resume or re-run pending builds


KrauseFx:
Right now, when a build runs, and the CI server either gets killed or restarted, a Build will stay stuck in `pending` for the rest of time of the universe. We gotta build something to see if the worker is actually still running, and if not, mark the build as failed. 

e.g. https://github.com/KrauseFx/ci-config/commit/6167cdac97f96fc34d1277683652539701cbc638#diff-2d77cf50a00bf53694d062b016527aa0R3

Sherlouk:
I wonder if this is relevant to this but another idea I had was the potential of, when retrying a build, if it could skip actions that have already happened successfully and somehow reuse the artifacts?

As we know Xcode tests have a tendency to randomly fail, but if I've already successfully built the app (using `build-for-testing`) then I shouldn't have to redo all of that to just retry the test action?

(Hope I explained that well enough!)

taquitos:
I think that makes sense, but we have to be super judicious about what artifacts are reusable. Having a list of things that are safe to reuse might be the best way to go about this.

PR #154, opened 2018-02-21
#52 - re-run pending builds on server initialization.

AndrewMcBurney:
The following is a PR for issue #52.

### Changes

**`launch.rb`**
- On [`Launch::take_off`](), pending builds will be run after all other configuration takes place
- [`Launch::run_pending_builds`](https://github.com/AndrewMcBurney/ci/blob/rerun_pending_builds/launch.rb#L143) iterates through all projects and appends a worker task dedicated to running pending builds corresponding to each project asynchronously
- `provider_credential` is now a [memoized class method](https://github.com/AndrewMcBurney/ci/blob/rerun_pending_builds/launch.rb#L191), since the `provider_credentials` are now being used in multiple class methods

**`json_build_data_source.rb`**
- I made the methods `JSONBuildDataSource::json_folder_path` and `JSONBuildDataSource::builds_path` [private](https://github.com/AndrewMcBurney/ci/blob/rerun_pending_builds/services/data_sources/json_build_data_source.rb#L57-L65), because I was worried about potential violation of the [dependency inversion principle](http://deviq.com/dependency-inversion-principle/) - since the abstract class doesn't have these public methods

### Use case

1. Have some builds with `pending` status in your Fastlane CI config repo:

screen shot 2018-02-25 at 1 36 18 pm

> I updated `1.json` and `3.json` to be `pending` builds

2. Boot up the Fastlane CI server

screen shot 2018-02-25 at 1 36 59 pm

3. _Voila_, the builds are rerun!

screen shot 2018-02-25 at 1 37 29 pm
screen shot 2018-02-25 at 1 37 40 pm

### Enhancements

- [x] Add tests
- [x] Show use case
- [x] Validate everything works :joy:

### DIFF
--- a/services/code_hosting/git_hub_service.rb
+++ b/services/code_hosting/git_hub_service.rb
@@ -44,7 +44,7 @@ module FastlaneCI
     # Does the client with the associated credentials have access to the specified repo?
     # @repo [String] Repo URL as string
     def access_to_repo?(repo_url: nil)
-      client.repository?(repo_url.sub("https://github.com/",""))
+      client.repository?(repo_url.sub("https://github.com/", ""))
     end
 
     # The `target_url`, `description` and `context` parameters are optional
--- a/launch.rb
+++ b/launch.rb
@@ -1,3 +1,5 @@
+require_relative "taskqueue/task_queue"
+
 module FastlaneCI
   # Launch is responsible for spawning up the whole
   # fastlane.ci server, this includes all needed classes
@@ -14,6 +16,7 @@ module FastlaneCI
       check_for_existing_setup
       prepare_server
       launch_workers
+      run_pending_builds
     end
 
     def self.require_fastlane_ci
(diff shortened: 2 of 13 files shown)


KrauseFx:
This looks amazing 👍

minuscorp:
It'd be great if you add the conformance to the `JSONDataSource` mixin to the `build_json_data_source`. 😁 I can help if you have any problems with it, you have examples on the user and project data sources!

services/code_hosting/git_hub_service.rb
```diff
@@ -44,7 +44,7 @@ def repos
     # Does the client with the associated credentials have access to the specified repo?
     # @repo [String] Repo URL as string
     def access_to_repo?(repo_url: nil)
-      client.repository?(repo_url.sub("https://github.com/",""))
+      client.repository?(repo_url.sub("https://github.com/", ""))
```
minuscorp: I was concerned when this was introduced. This raw substitution instead of a regex, are there any cases where `https://github.com/` is not used? (http?, ssh?, short url?)

taquitos: Right now we're focused on using https github links (most common). In the future, we will 💯 want to support all the things.

PR #155, opened 2018-02-21
Update ruby version to 2.3.3

AndrewMcBurney:
### Motivation

macOS 10.13.3 has ruby 2.3.3p222 [installed by default](https://apple.stackexchange.com/questions/257627/ruby-version-on-macos-sierra#answer-257629).


KrauseFx:
And https://github.com/fastlane/ci#system-requirements needs to be updated also 👍

PR #156, opened 2018-02-22
Initialize git_repo with callback when async.

minuscorp:
I've added it with a method with the `git_repo` created as the parameter.

### DIFF
--- a/shared/models/git_repo.rb
+++ b/shared/models/git_repo.rb
@@ -72,9 +72,9 @@ module FastlaneCI
       logger.debug("Adding task to setup repo #{self.git_config.git_url} at: #{self.git_config.local_repo_path}")
 
       setup_task = git_action_with_queue do
-        super_verbose("starting setup_repo #{self.git_config.git_url}".freeze)
+        logger.debug("starting setup_repo #{self.git_config.git_url}".freeze)
         self.setup_repo
-        super_verbose("done setup_repo #{self.git_config.git_url}".freeze)
+        logger.debug("done setup_repo #{self.git_config.git_url}".freeze)
       end
 
       # if we're starting asynchronously, we can return now.
@@ -107,10 +107,10 @@ module FastlaneCI
           # Now we have to check if the repo is actually from the
           # same repo URL
           if repo.remote("origin").url.casecmp(self.git_config.git_url.downcase).zero?
-            super_verbose("Resetting #{self.git_config.git_url}")
+            logger.debug("Resetting #{self.git_config.git_url}")
             self.git.reset_hard
 
-            super_verbose("Pulling #{self.git_config.git_url}")
+            logger.debug("Pulling #{self.git_config.git_url}")
             self.pull
           else
             logger.debug("[#{self.git_config.id}] Repo URL seems to have changed... deleting the old directory and cloning again")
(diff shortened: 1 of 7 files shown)


shared/models/git_repo.rb
```diff
@@ -43,17 +43,21 @@ class GitRepo
     attr_accessor :repo_auth # whatever pieces of information that can change between git users
 
     attr_accessor :temporary_storage_path
+    # return [proc]
```
KrauseFx: Please add a comment here describing what this is used for

shared/models/git_repo.rb
```diff
@@ -322,6 +326,13 @@ def clone(repo_auth: self.repo_auth, async: false)
       end
     end
 
+    def callback_block(async_start)
+      # How do we know that the task was successfully finished?
```
KrauseFx: You could make that a little easier to read:
```ruby
return if self.callback.nil?
return unless async_start

self.callback.call(self)
```

shared/models/git_repo.rb
```diff
@@ -43,17 +43,21 @@ class GitRepo
...
+    attr_accessor :callback
 
     class << self
       attr_accessor :git_action_queue
     end
 
     GitRepo.git_action_queue = TaskQueue::TaskQueue.new(name: "GitRepo task queue")
 
-    def initialize(git_config: nil, provider_credential: nil, async_start: false, sync_setup_timeout_seconds: 120)
+    def initialize(git_config: nil, provider_credential: nil, async_start: false, sync_setup_timeout_seconds: 120, callback: nil)
```
KrauseFx: I know this isn't added by this PR, but it would be good to have comments on what each of the parameters does here, e.g. `async_start` and `callback` 👍

minuscorp:
I think that's it, ready to merge? :D

taquitos:
It's weird it's picking up the other commits. Maybe you can pull master and rebase on that?

shared/models/git_repo.rb
```diff
@@ -161,13 +177,13 @@ def git
     # call like you would, but you also get the git repo involved, so it's  .each { |git, branch| branch.yolo; git.yolo }
     def git_and_remote_branches_each(&each_block)
       git_action_with_queue do
-        super_verbose("iterating through all remote branches of #{self.git_config.git_url}")
+        logger.debug("iterating through all remote branches of #{self.git_config.git_url}")
         branch_count = 0
         self.git.branches.remote.each do |branch|
-          each_block.call(self.git, branch)
-          branch_count = branch_count + 1
+          yield(self.git, branch)
```
taquitos: I actually prefer `each_block` here to be a little more explicit about what is going on. I think `yield` may be more preferred, but in cases like this I prefer clarity.

shared/models/git_repo.rb
```diff
@@ -293,43 +309,45 @@ def git_action_with_queue(ensure_block: nil, &block)
...
         end
       else
         clone_synchronously(repo_auth: repo_auth)
       end
     end
 
-    private
+    def callback_block(async_start)
+      # How do we know that the task was successfully finished?
+      return if self.callback.nil?
```
taquitos: It looks like if `self.callback` exists and `async_start` if false, we return without calling anything. 
Maybe we should crash in development mode since this shouldn't happen
Also, maybe we should check at initialization time for this bad state and error out there too.
Maybe in `validate_initialization_params`

taquitos:
Just a few comments I'd like your input on before we merge

shared/models/git_repo.rb
minuscorp: With the same thought, if we don't provide any callback when `async_start` we should raise another error?

taquitos: I think it's ok to not have a callback if you want to fire and forget, but I don't have a strong opinion one way or the other.

minuscorp: :+1: !

PR #157, opened 2018-02-22
Minor rubocop fix in `shared/models/git_repo.rb`.

AndrewMcBurney:
I found a minor rubocop violation, and ran the formatting command `bundle exec rubocop -a`.

### DIFF
--- a/shared/models/git_repo.rb
+++ b/shared/models/git_repo.rb
@@ -164,8 +164,8 @@ module FastlaneCI
         logger.debug("iterating through all remote branches of #{self.git_config.git_url}")
         branch_count = 0
         self.git.branches.remote.each do |branch|
-          each_block.call(self.git, branch)
-          branch_count = branch_count + 1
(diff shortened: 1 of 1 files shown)


shared/models/git_repo.rb
```diff
@@ -164,8 +164,8 @@ def git_and_remote_branches_each(&each_block)
         logger.debug("iterating through all remote branches of #{self.git_config.git_url}")
         branch_count = 0
         self.git.branches.remote.each do |branch|
-          each_block.call(self.git, branch)
-          branch_count = branch_count + 1
+          yield(self.git, branch)
```
taquitos: I don't agree with rubocop on this one for `each_block.call` to `yield`

AndrewMcBurney: Okay, I can change the `.rubocop.yml` file 👍

taquitos: Awesome, thank you!

PR #158, opened 2018-02-22
Added my own public ios app to the list of git repos allowed

nakhbari:


KrauseFx:
Instead, couldn't we just not provide a list of `all_repos`, but instead only offer different methods, e.g.:

- Does current user have access to specific repo
- Get list of all orgs
- Get list of all repos inside an org

taquitos:
Yeah, I think ultimately, though, we still need to address corner cases like one org having 1000+ repos.

KrauseFx:
That's right, but I feel like by selecting the org first, and then the repo, it should be enough for the MVP

PR #160, opened 2018-02-22
Check if the provided user has access to the individual projects 

nakhbari:
Use octokit client API for github to check if the repository exists. If the repository is private and the specified user does not have access, it will return false.

TESTED_LOCALLY

### DIFF
--- a/services/code_hosting/code_hosting_service.rb
+++ b/services/code_hosting/code_hosting_service.rb
@@ -19,7 +19,7 @@ module FastlaneCI
       not_implemented(__method__)
     end
 
-    def access_to_repo?(repo: nil)
+    def access_to_repo?(repo_url: nil)
       not_implemented(__method__)
     end
   end
--- a/services/code_hosting/code_hosting_service.rb
+++ b/services/code_hosting/code_hosting_service.rb
@@ -18,5 +18,9 @@ module FastlaneCI
     def repos
       not_implemented(__method__)
     end
+
+    def access_to_repo?(repo: nil)
+      not_implemented(__method__)
+    end
   end
 end
--- a/services/code_hosting/git_hub_service.rb
+++ b/services/code_hosting/git_hub_service.rb
@@ -43,8 +43,8 @@ module FastlaneCI
 
     # Does the client with the associated credentials have access to the specified repo?
     # @repo [String] Repo URL as string
-    def access_to_repo?(repo: nil)
-      client.repository?(repo.sub("https://github.com/",""))
+    def access_to_repo?(repo_url: nil)
+      client.repository?(repo_url.sub("https://github.com/",""))
     end
 
     # The `target_url`, `description` and `context` parameters are optional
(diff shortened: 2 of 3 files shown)


services/config_service.rb
```diff
@@ -54,18 +54,9 @@ def octokit_projects(provider_credential: nil)
...
-      # proper paging, or we ask for specific repos instead
-      # Either way, my account has access to too many repos, so for now, let's just workaround using this
-      current_repo_git_url_set << "https://github.com/nakhbari/Hello"
-      current_repo_git_url_set << "https://github.com/taquitos/ci-sample-repo"
-      current_repo_git_url_set << "https://github.com/fastlane/ci"
-
       logger.debug("Finding projects we have access to with #{provider_credential.ci_user.email}, #{provider_credential.type}")
       projects = self.project_service.projects.select do |project|
-        current_repo_git_url_set.include?(project.repo_config.git_url)
+        current_code_hosting_service.access_to_repo?({repo: project.repo_config.git_url})
```
taquitos: `access_to_repo?({repo: project.repo_config.git_url})` can be:
`access_to_repo?(repo: project.repo_config.git_url)`

services/code_hosting/code_hosting_service.rb
```diff
@@ -18,5 +18,9 @@ def provider_type
     def repos
       not_implemented(__method__)
     end
+
+    def access_to_repo?(repo: nil)
```
taquitos: in this case, I think we actually want it be maybe
`access_to_repo?(repo_config: nil)`
or even 
`access_to_repo?(repo_url: nil)`
The reason being, we have a `GitRepo` object already and that is very heavyweight.

services/code_hosting/git_hub_service.rb
```diff
@@ -41,6 +41,12 @@ def repos
       client.repos({}, query: { sort: "asc" })
     end
 
+    # Does the client with the associated credentials have access to the specified repo?
+    # @repo [String] Repo URL as string
+    def access_to_repo?(repo: nil)
+      client.repository?(repo.sub("https://github.com/",""))
```
taquitos: yeah, in this case, it's not a `repo` it's just a string. You could either rename it `git_url` or pass in a `repo_config` item and extract the `git_url` from there.

nakhbari: if i name the parent `repo_url` i'll still have to use that name correct? Can the child change the name of the param?

taquitos: you'll need to use `repo_url` if you change it to that. The child cannot change the name of a named param.

services/code_hosting/code_hosting_service.rb
nakhbari: changed to repo_url, to be repo type agnostic.

services/config_service.rb
nakhbari: done

services/code_hosting/git_hub_service.rb
nakhbari: done

PR #175, opened 2018-02-26
Fix the user authentication to properly assert the decoded password

nakhbari:


services/data_sources/json_user_data_source.rb
```diff
@@ -88,12 +88,13 @@ def login(email: nil, password: nil)
...
 
       # sweet, it matches, return the user
-      if user_password.to_s == password
+      if user_password == password
         logger.debug("user #{email} authenticated")
         return user
       end
 
       # nope, wrong password
+      logger.debug("user #{email} authentication failed")
```
taquitos: good idea 🚀

Issue #80, opened 2018-01-30
Wizard for creating a private configuration repository


taquitos:
- [ ] Sanity test all configuration before moving on

KrauseFx:
When launching and using fastlane.ci for the first time, we should help the user create a new private repo (right now, we're gonna focus just on GitHub, see fastlane.ci vision). 

Basically right now, fastlane.ci won't start at all when not having a git config repo, as we haven't planned the onboarding experience yet. For the time being, I guess it would be enough to have the doc available as part of the main README, and then telling the user to set it as the ENV variable for the `FASTLANE_CI_REPO_URL`, and then as Josh mentioned, do some checks. In the beginning those checks could be as simple as:
- Can I read and write on that repo
- Is the repo empty

@taquitos let me know if there was anything else you were thinking about on this one 👍

taquitos:
I think for the time being that is ok. 

We definitely want to keep this in the MVP though, so anybody who wants to work on a wizard for creating a private config repo is totally 💯 free to pick this up and make the first-use experience awesome because we can't ship the MVP without it!

AndrewMcBurney:
Closed in #275

KrauseFx:
yesss 🚀

PR #194, opened 2018-02-27
#80 Add first-time configuration wizard for setting up configuration repo.

AndrewMcBurney:
This PR addresses issue #80, and introduces some breaking changes to the way we currently configure FastlaneCI.

### Changes

**[features/configuration/configuration_controller.rb](https://github.com/AndrewMcBurney/ci/blob/new-configuration-repo-wizard/features/configuration/configuration_controller.rb)**
- `ConfigurationController` is a controller with the purpose of managing FastlaneCI configuration. It has two main purposes:
  - Lets the user configure the environment specific variables found in the `~/.fastlane/ci/.keys` file
  - Lets the user create a private remote repository, and populates it with the `users.json` and `projects.json` files

**[services/configuration_repository_service.rb](https://github.com/AndrewMcBurney/ci/blob/new-configuration-repo-wizard/services/configuration_repository_service.rb)**
- `ConfigurationRepositoryService` provides operations to create and mutate the FastlaneCI configuration repository.
- It creates the `user.json` file with the CI user, its corresponding provider credentials, and the empty JSON array `projects.json` file.

**[services/environment_variable_service.rb](https://github.com/AndrewMcBurney/ci/blob/new-configuration-repo-wizard/services/environment_variable_service.rb)**
- ` EnvironmentVariableService` provides methods for managnig the `~/.fastlane/ci/.keys` environment variables file and its corresponding environment variables

**Other changes**

- The server can now run without the local `~/.fastlane/ci/fastlane-ci-config` repository
- However, when the user runs the server without the `~/.fastlane/ci/fastlane-ci-config` repository, and logs into the server with their CI credentials (i.e., email => `FASTLANE_CI_USER`, password => `FASTLANE_CI_PASSWORD`) they will be redirected to the `configuration/` page where they'll be asked to create the remote repository.

**IMPORTANT NOTE**
- Any developers working on FastlaneCI must move their `.keys` file from the FastlaneCI project root (i.e., where the `config.ru` file is located) to `~/.fastlane/ci/.keys`.


features/configuration/configuration_controller.rb
```diff
@@ -0,0 +1,207 @@
...
+    end
+
+    # Updates a project existing in the configuration repository `projects.json`
+    post "#{HOME}/projects/update/:id" do
+    end
+
+    # Removes a project from the configuration git repo `projects.json`, but
+    # does not actually delete the project repository
+    post "#{HOME}/projects/delete/:id" do
+      # TODO: consider using projects_controller delete action instead? - probably better
```
minuscorp: 👍

services/file_writers/projects_writer.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require_relative "file_writer"
+
+module FastlaneCI
```
AndrewMcBurney: I'm going to use the JSONDataSource that we currently have instead of this - this part is currently pretty bad 😳

taquitos: Also, if you want to use file templates, you might want to consider using ERBs (like how we have them for our UI). That being said, if you're using JSON then yah, JSONDataSource might be the way to go 👍

AndrewMcBurney: Good call! For the `keys` template I'll use ERB :D

features/configuration/configuration_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+    private
+
+    #####################################################
+    # @!group Data: View-specific data
+    #####################################################
+
+    # @return [Hash]
+    def keys
+      {
+        encryption_key: ENV["FASTLANE_CI_ENCRYPTION_KEY"],
```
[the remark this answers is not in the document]
AndrewMcBurney: That's a great idea!

launch.rb
```diff
@@ -149,7 +112,21 @@ def self.prepare_server
...
+
+      github_projects = Services.config_service.projects(provider_credential: self.provider_credential)
+      github_service = FastlaneCI::GitHubService.new(provider_credential: self.provider_credential)
+
+      run_pending_github_builds(projects: github_projects, github_service: github_service)
+      enqueue_builds_for_open_github_prs_with_no_status(projects: github_projects, github_service: github_service)
+    end
+
     def self.launch_workers
+      return if first_time_user?
```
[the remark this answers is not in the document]
AndrewMcBurney: Ahh, yeah - that's a good point. The reason I had the guard was in case someone wants to call `Launch.launch_workers` outside of the context of `Launch.run_github_workers`.

Do you think it's worth having the guard statement, or should I just have the one guard in `Launch.run_github_workers`?

features/configuration/configuration_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+    #
+    # 2) Redirect back to `/configuration`
+    post "#{HOME}/git_repo" do
+      variables =
+        if keys.none?(&:nil?)
+          Services.configuration_repository_service.create_private_remote_configuration_repo
+          Services.reset_services!
+          Launch.trigger_initial_ci_setup
+          Launch.run_github_workers
+          { status: STATUS[:success], message: "Remote repo successfully created" }
```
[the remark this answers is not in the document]
AndrewMcBurney: Yeah, that's a great idea! 👍 We can probably put it in the `features/partials/` directory (and perhaps rename the directory to `features/shared` so it's more apparent it's to be used in multiple feature views.

features/partials/navigation.erb
```diff
@@ -3,6 +3,7 @@
   <nav class="mdl-navigation">
     <a class="mdl-navigation__link" href="/">Dashboard</a>
     <% if defined?(user) && user %>
+      <a class="mdl-navigation__link" href="/configuration">Configuration</a>
```
KrauseFx: Feel free to remove `Settings here` then 👍

PR #195, opened 2018-02-27
Add Nightly Build Worker

nakhbari:
Added:

- Nightly Build Worker
- Scheduler being initialized on initialize for each worker
- shutdown scheduler if the thread is told to stop
- stop from scheduling the cron job again if the scheduler has already scheduled something
- only init scheduler if cron_schedule is provided

TESTED_LOCALLY with a closer time than midnight 😉

### DIFF
--- a/services/worker_service.rb
+++ b/services/worker_service.rb
@@ -1,4 +1,5 @@
 require_relative "../workers/check_for_new_commits_on_github_worker"
+require_relative "../workers/nightly_build_github_worker"
 require_relative "../shared/logging_module"
 require_relative "../shared/models/provider_credential"
 
@@ -27,15 +28,18 @@ module FastlaneCI
       raise "incompatible repo_config and provider_credential" if provider_credential.type != repo_config.provider_credential_type_needed
 
       logger.debug("Starting worker for #{project.project_name}, on behalf of #{user_responsible.email}")
-      new_worker = nil
+      new_workers = []
       case provider_credential.type
       when FastlaneCI::ProviderCredential::PROVIDER_CREDENTIAL_TYPES[:github]
-        new_worker = FastlaneCI::CheckForNewCommitsOnGithubWorker.new(provider_credential: provider_credential, project: project)
+        new_workers.push(FastlaneCI::CheckForNewCommitsOnGithubWorker.new(provider_credential: provider_credential, project: project))
+        new_workers.push(FastlaneCI::NightlyBuildGithubWorker.new(provider_credential: provider_credential, project: project))
       else
         raise "unrecognized provider_type: #{provider_credential.type}"
       end
 
-      self.resource_workers_dictionary[worker_key] = new_worker
+      new_workers.each do |new_worker|
+        self.resource_workers_dictionary[worker_key] = new_worker
+      end
     end
 
     def stop_worker(project: nil, user_responsible: nil)
--- a/workers/nightly_build_github_worker.rb
+++ b/workers/nightly_build_github_worker.rb
@@ -1,0 +1,113 @@
+require_relative "worker_base"
+require_relative "../services/build_service"
+require_relative "../shared/models/job_trigger"
+require_relative "../shared/models/provider_credential"
+require_relative "../shared/logging_module"
+require_relative "../services/test_runner_service"
+require_relative "../services/code_hosting/git_hub_service"
+
+module FastlaneCI
+  # Responsible for starting off builds nightly
+  # TODO: determine what timezone to use
+  class NightlyBuildGithubWorker < WorkerBase
+    include FastlaneCI::Logging
+    NIGHTLY_CRON_TIME = "0 0 * * *"
+
+    attr_accessor :provider_credential
+    attr_accessor :project
+    attr_accessor :user_config_service
+    attr_accessor :github_service
+    attr_accessor :serial_task_queue
+    attr_accessor :project_full_name
+    attr_accessor :scheduler
+    attr_accessor :target_branches_set
+
+    attr_writer :git_repo
+
+    def provider_type
+      return FastlaneCI::ProviderCredential::PROVIDER_CREDENTIAL_TYPES[:github]
+    end
+
+    def initialize(provider_credential: nil, project: nil)
+      self.provider_credential = provider_credential
+      self.github_service = FastlaneCI::GitHubService.new(provider_credential: provider_credential)
+      self.project = project
+      self.scheduler = WorkerScheduler.new(cron_schedule: NIGHTLY_CRON_TIME)
+
+      self.target_branches_set = Set.new
+      project.job_triggers.each do |trigger|
+        if trigger.type == FastlaneCI::JobTrigger::TRIGGER_TYPE[:nightly]
+          self.target_branches_set.add(trigger.branch)
+        end
+      end
+
+      self.project_full_name = project.repo_config.git_url
+
+      if project.repo_config.kind_of?(FastlaneCI::GitRepoConfig)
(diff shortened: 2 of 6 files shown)


workers/nightly_build_github_worker.rb
```diff
@@ -0,0 +1,113 @@
...
+require_relative "../shared/models/job_trigger"
+require_relative "../shared/models/provider_credential"
+require_relative "../shared/logging_module"
+require_relative "../services/test_runner_service"
+require_relative "../services/code_hosting/git_hub_service"
+
+module FastlaneCI
+  # Responsible for starting off builds nightly
+  # TODO: determine what timezone to use
+  class NightlyBuildGithubWorker < WorkerBase
```
nakhbari: This is branched from CheckForNewCommitGithubWorker
http://www.mergely.com/Tr0JT2n4/

services/worker_service.rb
```diff
@@ -27,15 +28,18 @@ def start_worker_for_provider_credential_and_config(project: nil, provider_crede
...
+        new_workers.push(FastlaneCI::CheckForNewCommitsOnGithubWorker.new(provider_credential: provider_credential, project: project))
+        new_workers.push(FastlaneCI::NightlyBuildGithubWorker.new(provider_credential: provider_credential, project: project))
       else
         raise "unrecognized provider_type: #{provider_credential.type}"
       end
 
-      self.resource_workers_dictionary[worker_key] = new_worker
+      new_workers.each do |new_worker|
+        self.resource_workers_dictionary[worker_key] = new_worker
+      end
```
taquitos: Take a look at `launch.rb:130` at `def self.launch_workers` 
I think calculating `number_of_workers_started` is now inaccurate

taquitos:
I think it's time to rename `start_worker_for_provider_credential_and_config` to something more accurate, like `start_workers_for_project_and_credential`

services/worker_service.rb
```diff
@@ -27,15 +28,18 @@ def start_worker_for_provider_credential_and_config(project: nil, provider_crede
       raise "incompatible repo_config and provider_credential" if provider_credential.type != repo_config.provider_credential_type_needed
 
       logger.debug("Starting worker for #{project.project_name}, on behalf of #{user_responsible.email}")
-      new_worker = nil
+      new_workers = []
       case provider_credential.type
       when FastlaneCI::ProviderCredential::PROVIDER_CREDENTIAL_TYPES[:github]
-        new_worker = FastlaneCI::CheckForNewCommitsOnGithubWorker.new(provider_credential: provider_credential, project: project)
+        new_workers.push(FastlaneCI::CheckForNewCommitsOnGithubWorker.new(provider_credential: provider_credential, project: project))
```
taquitos: instead of `.push` (which is totally legit) we prefer ` << `

services/worker_service.rb
```diff
@@ -27,15 +28,18 @@ def start_worker_for_provider_credential_and_config(project: nil, provider_crede
...
+        new_workers.push(FastlaneCI::NightlyBuildGithubWorker.new(provider_credential: provider_credential, project: project))
       else
         raise "unrecognized provider_type: #{provider_credential.type}"
       end
 
-      self.resource_workers_dictionary[worker_key] = new_worker
+      new_workers.each do |new_worker|
+        self.resource_workers_dictionary[worker_key] = new_worker
```
taquitos: if there are two workers in `new_workers`, this will only store the last one since we're using the same `worker_key`. I think you need to generate a new key for each worker

taquitos: Alternatively, you could make it so that `resource_workers_dictionary` is a hash of `{ key => workers_array }` instead of `{key => single_worker}`

services/worker_service.rb
```diff
@@ -27,15 +28,18 @@ def start_worker_for_provider_credential_and_config(project: nil, provider_crede
       raise "incompatible repo_config and provider_credential" if provider_credential.type != repo_config.provider_credential_type_needed
 
       logger.debug("Starting worker for #{project.project_name}, on behalf of #{user_responsible.email}")
-      new_worker = nil
+      new_workers = []
       case provider_credential.type
       when FastlaneCI::ProviderCredential::PROVIDER_CREDENTIAL_TYPES[:github]
-        new_worker = FastlaneCI::CheckForNewCommitsOnGithubWorker.new(provider_credential: provider_credential, project: project)
+        new_workers.push(FastlaneCI::CheckForNewCommitsOnGithubWorker.new(provider_credential: provider_credential, project: project))
+        new_workers.push(FastlaneCI::NightlyBuildGithubWorker.new(provider_credential: provider_credential, project: project))
```
taquitos: I think we also need to check: 
if the project has any `NightlyJobTriggers` before creating a worker for that
if the project has any `CommitJobTriggers` before creating a worker for that

workers/nightly_build_github_worker.rb
```diff
@@ -0,0 +1,113 @@
...
+    # since it will be run on the "main" thread if it were
+    # part of the #initialize method
+    def git_repo
+      @git_repo ||= GitRepo.new(
+        git_config: project.repo_config,
+        provider_credential: provider_credential
+      )
+    end
+
+    def work
```
taquitos: A lot of this stuff is copy/paste from the other worker. How about you extract out the common stuff into `GitHubWorkerBase`, and create 2 new subclasses, `NightlyGitHubWorker` and `NewCommitGitHubWorker` (where `NewCommitGitHubWorker` is the new name for `CheckForNewCommitGithubWorker`)

workers/worker_base.rb
```diff
@@ -25,7 +25,11 @@ def initialize
           # We have the `work` inside a `begin rescue`
           # so that if something fails, the thread still is alive
           begin
-            self.scheduler.schedule { self.work } unless self.should_stop
+            if self.should_stop
+              self.scheduler.shutdown
+            else
+              self.scheduler.schedule { self.work }
```
taquitos: because this is happening in a loop, does this mean `self.scheduler.schedule { self.work }` will be called many times? In this case, we might want to consider removing the thread and just using `self.scheduler`

workers/worker_scheduler.rb
```diff
@@ -25,11 +32,25 @@ def initialize(sleep_interval: nil, cron_schedule: nil)
...
+        self.scheduled_cron_job = self.scheduler.job(job_id)
+        logger.debug("Scheduling cron job for #{self.cron_schedule}.")
+        logger.debug("Next time #{self.scheduled_cron_job.next_time} or #{(self.scheduled_cron_job.next_time - Time.now) / (60 * 60)} hours from now.")
+      end
+    end
+
+    # Shuts down the scheduler, ceases any scheduler/triggering activity.
+    def shutdown
+      if self.scheduler
+        self.scheduler.shutdown
```
taquitos: can be rewritten as `self.scheduler.shutdown if self.scheduler` ❤️ ruby for this

workers/worker_scheduler.rb
```diff
@@ -25,11 +32,25 @@ def initialize(sleep_interval: nil, cron_schedule: nil)
...
     def schedule(&block)
-      if !self.sleep_interval.nil?
+      if self.sleep_interval
         block.call
         Kernel.sleep(self.sleep_interval)
-      elsif !self.cron_schedule.nil?
-        self.scheduler.cron(self.cron_schedule) { block.call }
+      elsif self.cron_schedule && self.scheduled_cron_job.nil?
+        job_id = self.scheduler.cron(self.cron_schedule) do
+          self.scheduled_cron_job = nil
```
taquitos: Here is the cron loop, right? As in, every time the cron job is run, this block is executed?

nakhbari:
did the changes but this CL is getting chunky. Gonna break this up

PR #198, opened 2018-03-02
Create worker base for workers that use github.

nakhbari:
TESTED_LOCALLY

### DIFF
--- a/workers/check_for_new_commits_on_github_worker.rb
+++ b/workers/check_for_new_commits_on_github_worker.rb
@@ -9,7 +9,7 @@ module FastlaneCI
   # We have to poll, as there is no easy way to hear about
   # new commits from web events, as the CI system might be behind
   # firewalls
-  class CheckForNewCommitsOnGithubWorker < GithubWorkerBase
+  class CheckForNewCommitsOnGithubWorker < GitHubWorkerBase
     include FastlaneCI::Logging
 
     attr_accessor :trigger_type
@@ -22,7 +22,7 @@ module FastlaneCI
 
       self.current_tasks = []
 
-      super # This starts the work by calling `work`
+      super(provider_credential: provider_credential, project: project) # This starts the work by calling `work`
     end
 
     def wait_for_previous_tasks?
--- a/workers/check_for_new_commits_on_github_worker.rb
+++ b/workers/check_for_new_commits_on_github_worker.rb
@@ -1,78 +1,28 @@
-require_relative "worker_base"
+require_relative "github_worker_base"
 require_relative "worker_scheduler"
 require_relative "../services/build_service"
-require_relative "../shared/models/provider_credential"
 require_relative "../shared/models/job_trigger"
 require_relative "../shared/logging_module"
-require_relative "../services/test_runner_service"
-require_relative "../services/code_hosting/git_hub_service"
 
 module FastlaneCI
   # Responsible for checking if there have been new commits
   # We have to poll, as there is no easy way to hear about
   # new commits from web events, as the CI system might be behind
   # firewalls
-  class CheckForNewCommitsOnGithubWorker < WorkerBase
+  class CheckForNewCommitsOnGithubWorker < GithubWorkerBase
     include FastlaneCI::Logging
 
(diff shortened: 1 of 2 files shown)


workers/github_worker_base.rb
```diff
@@ -0,0 +1,112 @@
+require_relative "worker_base"
+require_relative "../shared/models/provider_credential"
+require_relative "../shared/logging_module"
+require_relative "../services/test_runner_service"
+require_relative "../services/code_hosting/git_hub_service"
+
+module FastlaneCI
+  # Base class for GitHub workers
+  class GithubWorkerBase < WorkerBase
```
taquitos: GitHub (i know we get the casing wrong in a few places too)

workers/check_for_new_commits_on_github_worker.rb
```diff
@@ -1,78 +1,28 @@
...
-      project_full_name = project.repo_config.git_url
-
-      if project.repo_config.kind_of?(FastlaneCI::GitRepoConfig)
-        project_full_name = project.repo_config.full_name unless project.repo_config.full_name.nil?
-      end
-
-      self.serial_task_queue = TaskQueue::TaskQueue.new(name: "#{project_full_name}:#{provider_credential.ci_user.email}")
       self.current_tasks = []
 
-      super() # This starts the work by calling `work`
```
taquitos: will still need to call super with `super(provider_credential: provider_credential, project: project)`

workers/check_for_new_commits_on_github_worker.rb
```diff
@@ -1,78 +1,28 @@
...
-    # part of the #initialize method
-    def git_repo
-      @git_repo ||= GitRepo.new(
-        git_config: project.repo_config,
-        provider_credential: provider_credential
-      )
+      super # This starts the work by calling `work`
     end
 
     def wait_for_previous_tasks?
```
taquitos: We might end up wanting this in the base class too, but I'm not sure. Just something I'm wondering about.

workers/github_worker_base.rb
```diff
@@ -0,0 +1,112 @@
...
+      super() # This starts the work by calling `work`
+    end
+
+    def thread_id
+      if @thread_id
+        return @thread_id
+      end
+
+      time_nano = Time.now.nsec
+      @thread_id = "GithubWorker:#{time_nano}: #{self.serial_task_queue.name}"
```
taquitos: Let's use the current class name instead of `GithubWorker` so the logging is more precise

workers/check_for_new_commits_on_github_worker.rb
nakhbari: Yes. I should do this. 
I did this originally but then found out without the () super gets called with all the same args

workers/check_for_new_commits_on_github_worker.rb
nakhbari: Yeah i guess i can move this to the base class. This isn't really needed for nightly/timed builds so i didn't add it.

Maybe we can do this if another child requires it

workers/check_for_new_commits_on_github_worker.rb
taquitos: If you want to keep it as `super` I'd be ok with that. Originally I made this comment because it didn't look like you called `super` at all due to how GitHub had everything highlighted for me.
I usually prefer explicitly calling methods with named params, but I think the way you have it might be more idiomatic to ruby, so either way 👍

workers/check_for_new_commits_on_github_worker.rb
taquitos: 💯 that makes sense

workers/github_worker_base.rb
nakhbari: good catch!

workers/check_for_new_commits_on_github_worker.rb
nakhbari: done

workers/github_worker_base.rb
nakhbari: done

PR #199, opened 2018-03-02
Add Nightly GitHub worker, but not using it yet

nakhbari:


workers/worker_base.rb
```diff
@@ -25,7 +25,11 @@ def initialize
           # We have the `work` inside a `begin rescue`
           # so that if something fails, the thread still is alive
           begin
-            self.scheduler.schedule { self.work } unless self.should_stop
+            if self.should_stop
+              self.scheduler.shutdown
+            else
+              self.scheduler.schedule { self.work }
```
nakhbari: Copying @taquitos comment 
>because this is happening in a loop, does this mean self.scheduler.schedule { self.work } will be called many times? In this case, we might want to consider removing the thread and just using self.scheduler. I think this is the way to move forward because the cron schedule already takes care of repetition, right? So there's no need to call this in a loop.

This was my thought earlier about having a differentiation between cron jobs and workers. What do you think about adding these as types to the worker?

taquitos: I wonder if we can just replace the normal workers the rely on sleep intervals with a scheduler instead? So if somebody wants to run this every 10 seconds, we just create a cron job that repeats every 10 seconds.

nakhbari: YES!

workers/check_for_new_commits_on_github_worker.rb
```diff
@@ -68,4 +68,4 @@ def work
       end
     end
   end
-end
\ No newline at end of file
+end
```
taquitos: Gotta make sure we have a new line at the end of the file

workers/github_worker_base.rb
```diff
@@ -110,4 +110,4 @@ def trigger_type
       not_implemented(__method__)
     end
   end
-end
\ No newline at end of file
+end
```
taquitos: new line needed at end

workers/nightly_build_github_worker.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      self.target_branches do |git, branch|
+        current_sha = repo.most_recent_commit.sha
+        logger.debug("Running Nightly build on branch #{branch.name} with sha #{current_sha}")
+
+        self.create_and_queue_build_task
+      end
+    end
+  end
+end
```
taquitos: new line needed at end

taquitos:
What do you think about fixing that `sleep` vs `cron` stuff before merging? I'm thinking we should do it because of the scheduler stuff I pointed out.

workers/github_worker_base.rb
nakhbari: done

workers/nightly_build_github_worker.rb
nakhbari: done

workers/worker_base.rb
nakhbari: done.

workers/check_for_new_commits_on_github_worker.rb
nakhbari: done.

workers/worker_scheduler.rb
```diff
@@ -3,10 +3,13 @@
 module FastlaneCI
   # Class that handles the scheduling for fastlane.ci workers
   class WorkerScheduler
```
nakhbari: Might just remove this class later and do this directly in the scheduler method.

workers/worker_scheduler.rb
```diff
@@ -25,12 +28,19 @@ def initialize(sleep_interval: nil, cron_schedule: nil)
     end
 
     def schedule(&block)
-      if !self.sleep_interval.nil?
-        block.call
-        Kernel.sleep(self.sleep_interval)
-      elsif !self.cron_schedule.nil?
-        self.scheduler.cron(self.cron_schedule) { block.call }
+      if self.sleep_interval
+        self.scheduler.every(self.sleep_interval) { block.call }
```
taquitos: :heart:

Issue #147, opened 2018-02-16
[Proposal] Build standalone JS class to convert fastlane output from generic format to prettified HTML


KrauseFx:
So I'm thinking a bit more about how we render the real-time log of fastlane.ci in the user's browser. We haven't decided on front-end frameworks and stuff yet, so things are still in the open
but having HTML based output for fastlane output is something that would be interesting outside of fastlane.ci also
What I mean, imagine you can run `fastlane beta --html` (a regular lane) and it would additionally generate a HTML file with colored output, which can then be used for other CI systems, etc.

so the way we could architect is could be that it's a single `.js` file that converts JSON data to beautiful HTML based log files. The JSON data could look something like this

```json
[
  {
      "type": "message",
      "text": "fastlane is great"
  },
  {
      "type": "header",
      "text": "Step: ensure_git_status"
  },
  {
      "type": "error",
      "text": "Ugh, git isn't clean"
  }
]
```

generating HTML + CSS to something like this (very simplified of course)

```html
<p>fastlane is great</p>
<h1>Step: ensure_git_status</h1>
<p class="error">Ugh, git isn't clean</p>
```

by having a single `.js` file, we could use it for existing `fastlane` setups by just running that `js` file after running a full build
and then we could use the same `.js` file for real-time output rendering for `fastlane.ci`

- So in the case of `fastlane` it would run on the user's machine locally using the local JS engine that's pre-installed on the Mac
- In the case of `fastlane.ci` we'd run it on the user's browser, piping over the JSON metadata through a web socket, updating the HTML dom for the user in real-time

revolter:
Firstly, I don't think there is a JS engine pre-installed in macOS.
And secondly, where would the CSS be? Or we require the user to provide one, being able to customize it how they want?

KrauseFx:
The CSS would be at 2 locations:
- for `fastlane.ci` it would be part of `fastlane.ci`
- for `fastlane` it would be bundled in the HTML file, probably a more basic style sheet

KrauseFx:
I'm gonna have to look into more detail on the JavaScript side of things, but it looks like it's pre-installed on every Mac
```
/System/Library/Frameworks/JavaScriptCore.framework/Versions/A/Resources/jsc
```

revolter:
Oh, ok, didn't know about that

KrauseFx:
I'm of course also open for other suggestion. I guess the best alternative would be a Ruby based implementation:

[...]

PR #200, opened 2018-03-02
Real-time build log stream

KrauseFx:
Fixes https://github.com/fastlane/ci/issues/201
Fixes https://github.com/fastlane/ci/issues/147

**To test this change:**
1. open a project view (e.g. `http://localhost:9292/projects/details/db799377-aaa3-4605-ba43-c91a13c8f83`)
1. Click on `Trigger job now`
1. See how half the build output is there instantly (shows that fastlane.ci "caches" the output)
1. See how more messages are being loaded as _fastlane_ runs the `Fastfile` (has a few sleeps in there to proof the real-time-ness)

**Notes**
1. Thin seems to require the use of `localhost` instead of `127.0.0.1`, I didn't look into this yet if we can support both
1. Main thing that's missing to get this in a rather solid state is the identification of a build. As raised the question in Slack (`I see the Build object doesn't have an ID like we have for all other objects. Is there a reason for that? Did we decide you can't run multiple builds for one sha? Is project + sha unique?)`). My proposal is to add a unique ID to builds also, so that this ID in combination with the project is unique (or even without the project?). Once that's done, multiple `# TODO` items from this PR can be resolved. (e.g. see https://github.com/fastlane/ci/pull/200/files#diff-40d0c75f55437cd62b7224bd7d90d275R14)
1. The custom middleware holds a reference to each web socket connection (https://github.com/fastlane/ci/pull/200/files#diff-7db6fa3652598445f8cb2adaed28ae88R40) grouped by the Build ID, so once the build ID is unique, without the project ideally, this will work nicely.
1. The frontend is just a basic proof of concept of course
1. We discussed that we don't want to generate the HTML on the backend long term. I'll change this with a follow-up PR
1. The `fastlane.log` file is hard coded for now, until we have the artifact system ready. 
1. I get those weird `lib/git/lib.rb:460: warning: conflicting chdir during another chdir block` warnings again. I assume this happens due to running _fastlane_ and we do some `Dir.chdir` in that code base also. I haven't verified yet if this PR adds this issue, or if that's already the case in `master`.
1. As seen in https://github.com/fastlane/ci/pull/200/files#diff-62e64a60028b7cd0fdf6248bbc3f9bfcR13, we currently have to store references to all active `test_runner_services` to keep them alive, and to be able to access them when the user wants to see their progress. I added a comment with a proposal on how we can properly architect it

### DIFF
--- a/features/build/build_controller.rb
+++ b/features/build/build_controller.rb
@@ -9,7 +9,7 @@ module FastlaneCI
 
     use(FastlaneCI::BuildWebsocketBackend)
 
-    get "/projects/*/builds/*" do |project_id, build_id|
+    get "#{HOME}/*" do |project_id, build_id|
       project = self.user_project_with_id(project_id: project_id)
       build = project.builds.find { |b| b.sha == build_id } # TODO: We need a build ID, sha isn't enough
 
--- a/features/build/build_controller.rb
+++ b/features/build/build_controller.rb
@@ -13,7 +13,7 @@ module FastlaneCI
       project = self.user_project_with_id(project_id: project_id)
       build = project.builds.find { |b| b.sha == build_id } # TODO: We need a build ID, sha isn't enough
 
-      current_runner_service = TestRunnerService.test_runner_services.find do |t| 
+      current_runner_service = TestRunnerService.test_runner_services.find do |t|
         t.project.id == project_id && t.sha == build_id # TODO: just as above, use actual identification
       end
 
--- a/features/build/build_controller.rb
+++ b/features/build/build_controller.rb
@@ -13,6 +13,17 @@ module FastlaneCI
       project = self.user_project_with_id(project_id: project_id)
       build = project.builds.find { |b| b.sha == build_id } # TODO: We need a build ID, sha isn't enough
 
+      # TODO: currently we just trigger a build here
+      service = FastlaneCI::TestRunnerService.new(
+        project: project,
+        sha: build_id
+      )
+      BuildWebsocketBackend.test_runner_services ||= []
+      BuildWebsocketBackend.test_runner_services << service
+      Thread.new do
+        service.run
+      end
+
       locals = {
         project: project,
         build: build,
--- a/features/build/build_controller.rb
+++ b/features/build/build_controller.rb
@@ -21,7 +21,7 @@ module FastlaneCI
         project: project,
         build: build,
         title: "Project #{project.project_name}, Build #{build.sha}",
-        existing_rows: current_runner_service.all_lines.collect { |a| a[:html] }
+        existing_rows: current_runner_service.all_build_output_log_lines.collect { |a| a[:html] }
       }
       erb(:build, locals: locals, layout: FastlaneCI.default_layout)
     end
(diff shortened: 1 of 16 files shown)


services/test_runner_service.rb
```diff
@@ -11,14 +27,57 @@ class TestRunnerService
     attr_accessor :current_build
     attr_accessor :sha
 
-    def initialize(project: nil, sha: nil, github_service: nil)
+    # All lines that were generated so far, this might not be a complete run
+    # This is an array of hashes
+    # TODO: have a class representing a Row (has to offer dynamic values though, as we might have non fastlane runners in the future)
+    attr_accessor :all_lines
```
taquitos: `all_build_output_log_lines` might be clearer

services/test_runner_service.rb
```diff
@@ -11,14 +27,57 @@ class TestRunnerService
...
+
+    # All blocks listening to changes for this build
+    attr_accessor :all_blocks
```
taquitos: `build_change_observer_blocks`

features/build/build_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+module FastlaneCI
+  # Controller for a single project view. Responsible for updates, triggering builds, and displaying project info
+  class BuildController < AuthenticatedControllerBase
+    HOME = "/projects/*/builds"
+
+    use(FastlaneCI::BuildWebsocketBackend)
+
+    get "#{HOME}/*" do |project_id, build_id|
+      project = self.user_project_with_id(project_id: project_id)
+      build = project.builds.find { |b| b.sha == build_id } # TODO: We need a build ID, sha isn't enough
```
minuscorp: This should be just the number?

features/test_runner/fastlane_test_runner.rb
```diff
@@ -4,13 +4,35 @@ module FastlaneCI
...
+          #
+          # {:type=>:success, :message=>"Everything worked"}
+          #
+          # we append the HTML code that should be used in the `html` key
+          # the result looks like this
+          #
+          # {"type":"success","message":"Driving the lane 'ios beta'","html":"<p class=\"success\">Driving the lane 'ios beta'</p>"}
+          #
+          row[:html] = FastlaneOutputToHtml.convert_row(row)
           yield(row)
```
minuscorp: This was failing for me (on `master`) `yield` misses a callable, where is it? 😕

KrauseFx: Sorry, my bad that it was failing on master, it's over here now https://github.com/fastlane/ci/blob/e1edb77c5904036f8b8ba6280fadc04dd0df9242/services/test_runner_service.rb#L108-L110

features/build/build_controller.rb
KrauseFx: Yes, once we resolved the question/proposal around having a unique Build ID

minuscorp:
Can we get this merged? So I can follow with #205 😄

KrauseFx:
Sorry for the delay @minuscorp, this is merged now 👍

PR #203, opened 2018-03-02
Start spinning Nightly build workers when the trigger exists.

nakhbari:


services/worker_service.rb
```diff
@@ -8,44 +9,61 @@ module FastlaneCI
...
+      self.project_to_workers_dictionary[workers_key] = nil
+    end
+
+    def num_workers
+      self.project_to_workers_dictionary.values.reduce(0) do |sum, workers|
+        sum + workers.length
+      end
+    end
+
+    def project_has_trigger_type(project: nil, trigger_type: nil)
```
snatchev: Methods that return booleans should end in `?`

services/worker_service.rb
```diff
@@ -8,44 +9,61 @@ module FastlaneCI
...
+      raise "Need both project and trigger_type" unless project && trigger_type
```
snatchev: Did we settle on a minimum ruby version? If it was > 2.2 then declaring the named parameters as `(project:, trigger_type:)` will make them required by runtime.

services/worker_service.rb
KrauseFx: Nice catch 👍

services/worker_service.rb
nakhbari: Cool! i'm not sure if we have a minimum ruby version. @KrauseFx is there a minimum ruby version set?

services/worker_service.rb
nakhbari: fixed.

services/worker_service.rb
taquitos: 2.3.3p222 🎉

PR #204, opened 2018-03-02
Do not run build if the manual trigger is not set

nakhbari:
Note: The "Trigger Job Now" button in the view has this request as the href. So if we return an error, it will just display it. The button needs to be fixed.

### DIFF
--- a/features/project/project_controller.rb
+++ b/features/project/project_controller.rb
@@ -18,6 +18,13 @@ module FastlaneCI
 
       repo = FastlaneCI::GitRepo.new(git_config: project.repo_config, provider_credential: current_github_provider_credential)
       current_sha = repo.most_recent_commit.sha
+      manual_triggers_allowed = project.job_triggers.any? { |trigger| trigger.type == FastlaneCI::JobTrigger::TRIGGER_TYPE[:manual] }
+
+      unless manual_triggers_allowed
+        status(403) # Forbidden
+        body("Cannot build. There is no manual build trigger, for this branch, associated with this project.")
+        return
+      end
 
       # TODO: not the best approach to spawn a thread
       Thread.new do


KrauseFx:
I remember there was some discussion around this, but could you remind me: does that mean a project that misses this given trigger, can never be triggered manually? What's the use-case for not being able to trigger a build manually? Sorry for probably repeating a question that's already been answered

taquitos:
a `JobTrigger` is a capability with associated metadata needed to perform that trigger. If you have a `ManualJobTrigger` on the project, that means the job can be triggered by hand. There might be times when we don't want that, and in that case, a `ManualJobTrigger` won't exist. This is so we have all our possible triggers in one place, so if the user choses to not allow manual triggering, it just doesn't exist as a `JobTrigger` 

It also allows for an easier time coding up a UI. If you want to display all possible ways to trigger a job, you just have to iterate through a list of triggers. If we didn't have a `ManualJobTrigger` we'd have to iterate and then, check the Project for `can_manually_trigger?` and then add that to the UI.
Maybe `ManualJobTrigger` could be renamed to `OneOffJobTrigger`

taquitos:
Some people only want builds to be triggered when a commit has been made to a `release/` branch
That way, nobody can accidentally trigger a release.

features/project/project_controller.rb
```diff
@@ -18,6 +18,13 @@ class ProjectController < AuthenticatedControllerBase
 
       repo = FastlaneCI::GitRepo.new(git_config: project.repo_config, provider_credential: current_github_provider_credential)
       current_sha = repo.most_recent_commit.sha
+      manual_triggers_allowed = project.job_triggers.any? { |trigger| trigger.type == FastlaneCI::JobTrigger::TRIGGER_TYPE[:manual] }
```
snatchev: Might be nice to (dynamically?) add some query methods to `JobTrigger` such as `job_trigger.manual?` so that this can be re-written as `project.job_triggers.any?(&:manual?)`

nakhbari: That would be cool! Mind if i tackle this in a separate PR?

snatchev: 👍 sounds good!

features/project/project_controller.rb
```diff
@@ -18,6 +18,13 @@ class ProjectController < AuthenticatedControllerBase
...
+
+      unless manual_triggers_allowed
+        status(403) # Forbidden
```
taquitos: I don't think we want to return any statuses, but rather, use the notification pipeline to send the message. For right now, this is totally OK, though.

KrauseFx: Yeah, good point, I guess this all very much depends on what we decide on for the frontend

taquitos: I think we should continue to explore single page webapps, but still go full steam ahead with our current design for MVP. Remembering that the M in MVP is Minimum 😄 
It would be trivial to migrate off ERBs and into a single-page webapp. I just don't want to do that right now because then we'd be committing to building a `client` on top of the server, and having to handle all the complexities that come along with that.

nakhbari: @taquitos could you add that to the web app issue ^

PR #205, opened 2018-03-02
Artifacts and ArtifactProvider

minuscorp:
- [x] `Project` should store its current `artifact_provider` information.

- [x] `Build` should have an array of `Artifact`.

- [x] `Artifact` should have a reference to the `ArtifactProvider` which created it.

- [x] `ArtifactProvider` provides a simple `store!` `retrieve!` interface.

- [x] `LocalArtifactProvider` stores an `Artifact` into the server.

- [ ] ~`GoogleCloudArtifactProvider` may be done in this PR?~

- [x] Communicate the `TestFastlaneRunner` with the `lane_context` action in order to retrieve the `SharedValues` that might represent an `Artifact`. (help needed @KrauseFx )

- [x] Add documentation.

- [ ] ~Add tests, if possible.~

- [x] Drop the beat.


features/test_runner/fastlane_test_runner.rb
```diff
@@ -1,24 +1,29 @@
 require_relative "./fastlane_test_runner_helpers/fastlane_ci_output"
+require_relative "test_runner"
 
 module FastlaneCI
   # Represents the test runner responsible for loading and running
   # fastlane Fastfile configurations
+  # TODO: run method *should* return an array of artifacts
```
minuscorp: Confirm that the `run` method should always return an array of artifacts.

features/test_runner/fastlane_test_runner.rb
```diff
@@ -1,24 +1,29 @@
...
+          # WARN: This crashes because (not block provided)
+          #yield(row)
         end
       )
       FastlaneCore::UI.ui_object = ci_output
 
       # this only takes a few ms the first time being called
       Fastlane.load_actions
 
+      # TODO: How do I access fastlane actions (like lane_context) from here?
```
minuscorp: I'm kind of stuck here.

workers/github_worker_base.rb
```diff
@@ -27,6 +27,7 @@ def initialize(provider_credential: nil, project: nil)
       self.project = project
 
       self.target_branches_set = Set.new
+      project.job_triggers = [] # WARN: DELETE
```
taquitos: TODO: 😄 We search for these, but not # WARN

minuscorp: Those WARN are for myself. There're a lot of TODO, so I use WARN to find code I have to delete before merge 😅

features/test_runner/fastlane_test_runner.rb
KrauseFx: Alright, just took a look, and @taquitos is gonna love this. The `lane_context` isn't attached to the fastlane runner, but instead, you can access it right after the `fast_file.runner.execute(self.lane, self.platform, self.parameters)` call using

```ruby
Actions.lane_context
```
this prints out something like
```ruby
=> {:DEFAULT_PLATFORM=>:ios, :PLATFORM_NAME=>:ios, :LANE_NAME=>"ios beta"}
```

as mentioned before though, for now, please just assume some dummy/test values, as I'm currently working on this file with my current PR of https://github.com/fastlane/ci/pull/200

minuscorp: That's great! I'll try to get some work on that, is the last piece of the puzzle, maybe it'd be great to wait to your #200 to finish stablishing how the `FastlaneRunner` will work. What do you think?

features/test_runner/fastlane_test_runner.rb
```diff
@@ -42,8 +47,10 @@ def run
...
```
KrauseFx: You already have that in the `ensure` block, right?

features/test_runner/fastlane_test_runner_helpers/fastlane_ci_output.rb
```diff
@@ -62,7 +62,10 @@ def success(message)
       )
     end
 
-    def message(message)
+    # TODO: Getting recurrent fails on this method due to wrong argument number
```
KrauseFx: If you have the stack trace, can you send it over?

launch.rb
```diff
@@ -31,8 +31,10 @@ def self.take_off
       github_projects = Services.config_service.projects(provider_credential: self.provider_credential)
       github_service = FastlaneCI::GitHubService.new(provider_credential: self.provider_credential)
 
-      run_pending_github_builds(projects: github_projects, github_service: github_service)
-      enqueue_builds_for_open_github_prs_with_no_status(projects: github_projects, github_service: github_service)
+      # WARN: Don't want any automatic build.
```
KrauseFx: Ha, I do the same locally for me to be able to launch the server 😂

services/config_data_sources/json_project_data_source.rb
```diff
@@ -40,6 +40,27 @@ def self.map_enumerable_type(enumerable_property_name: nil, current_json_object:
         job_trigger
       end
     end
+
+    def self.json_to_attribute_name_proc_map
```
KrauseFx: Could you add a comment above each of these methods with a description of what it's used for?

services/data_sources/json_build_data_source.rb
```diff
@@ -22,6 +25,42 @@ def self.json_to_attribute_name_proc_map
...
+    end
+  end
+
+  # Mixin the JSONConvertible class for Artifact
+  class Artifact
+    include FastlaneCI::JSONConvertible
+
+    def self.json_to_attribute_name_proc_map
+      provider_object_to_provider = proc { |object|
+        nil if object.nil?
```
KrauseFx: What does that line do? It's not 100% clear

services/test_runner_service.rb
```diff
@@ -198,5 +215,24 @@ def update_build_status_source!
       logger.error(ex.to_s)
       logger.error(ex.backtrace.join("\n"))
     end
+
+    def fastfile_path_from_project!(project: nil)
```
KrauseFx: Why is that code needed? I thought we already detect the Fastfile, because we already have the lane selection

shared/models/artifact.rb
```diff
@@ -0,0 +1,25 @@
+require "securerandom"
+
+module FastlaneCI
+  # Represents a single unit of data generated by a build
+  class Artifact
+    # @return [UUID] The id of the artifact.
+    attr_accessor :id
+
+    # @return [String] The type or kind of artifact.
+    attr_accessor :type
```
KrauseFx: Example values would be good here as part of the comment

shared/models/local_artifact_provider.rb
```diff
@@ -0,0 +1,69 @@
...
+require "tmpdir"
+require "pathname"
+
+module FastlaneCI
+  # ArtifactProvider backed by a local filesystem.
+  class LocalArtifactProvider < ArtifactProvider
+    class << self
+      # Provide a simple default root_path for users that don't want much configuration.
+      def default_root_path
+        Pathname.new(Dir.tmpdir).join("fastlane-artifacts")
```
KrauseFx: Instead, it might make sense to make use of `~/.fastlane/artifacts` (we already create `~/.fastlane`) for the user

services/test_runner_service.rb
minuscorp: The lane detection is made from the `project_controller` itself with exactly the same code defined there, and this has relationship with what I commented to you in slack, let me extend the thought in the comment below.

minuscorp:
So we are at this point:

https://github.com/fastlane/ci/blob/658b171d257f9c938a9089b0d8c2341910253783/features/test_runner/fastlane_test_runner.rb#L46

Where we load the fastfile of the (presumably) `project` we're building on. If we debug that line, we get that `FastlaneCore::FastlaneFolder.fastfile_path` evaluates to (for example): `~/Documents/ci/fastlane/Fastfile`. The local folder of the `ci` itself, instead of the fastfile of the `project` we're building. For this reason, every action executed by the `fastfile.runner` is the scope of the `ci`, instead of the `project` local path. This is why _fastlane_ does not find the `xcodeproj` or other files needed for the action execution.
So, in order to get into the `project` scope, should I just `Dir.chdir` to the `project` `git_repo_path` in the `run` method, or this is going to be achieved in other ways?

minuscorp:
I also add here a design rationale behind this PR:

## Artifact
The `Artifact` model provides a simple interface to store a reference to any kind of data, but it is a dummy class. It does not know how to access to its resource.
Also, it holds a reference to the `ArtifactProvider` (explained below) which created the `Artifact` itself.
`type` is just a `String` that refers to the _kind_ of data that the `Artifact` represents (i.e., `ipa`,  `apk`, `log`, etc.) 

## Artifact Provider
The `ArtifactProvider` is a simple interface class that ensures that, at least `store!` and `retrieve!` methods are implemented by child classes.
Also, it creates the `class_name` attribute, key element in the implementation of the system.

### Implementation
- The `Project` now holds a reference to its `ArtifactProvider`, it must be just one at a given moment but it can change between builds without affecting any build run.

- In order to instantiate the correct `ArtifactProvider` subclass, here's where `class_name` comes handy. When in the project's JSON we find the `artifact_provider` key, we make a custom mapping where we obtain a new instance of `class_name` and (**important, the subclasses must include `JSONConvertible`**) set its attributes.

- The `fastlane_runner` returns an array of hashes like `{ type: , path: }`, where the local artifacts were created by the run of a certain build in a project. With that information, _dummy_ `Artifact` are created, and later on `store!`d by the `project.artifact_provider`, which returns the actual `Artifact` which is valid to be encoded to JSON.

- Each `Build`, holds a reference for its artifacts, which are responsible (via their `provider`) of being `retrieve!`d.

So.. I think that's it. Wanted to explain the decisions made in order to achieve the objective in the best way I could think of.

features/test_runner/fastlane_test_runner.rb
minuscorp: Yeah! Deleting it in next commit 😄

services/config_data_sources/json_project_data_source.rb
minuscorp: These are `JSONConvertible` methods, the module includes documentation, should I add more context in those implementations or leave as is?

services/data_sources/json_build_data_source.rb
minuscorp: So, there might be some chances (but it shouldn't ever happen) that the `Artifact`'s `ArtifactProvider` is `null` in the JSON, so it must be `nil` in the object. But that means that the `Artifact` is unrecoverable.

KrauseFx:
The last comment with the description is great, this should be part of the PR (the code), in the form of markdown files, or comments on top of each class to describe what they do 👍

minuscorp:
`GoogleCloudArtifactProvider` may be done in this PR?

KrauseFx:
It's your call re `GoogleCloudArtifactProvider`, it's fine to have a separate PR and get this merged sooner

PR #207, opened 2018-03-02
Proposal for documenting consistent URL routes

KrauseFx:
There we go 🚀


KrauseFx:
Alright, let's merge this and create a follow-up card to actually implement it 👍

Issue #219, opened 2018-03-05
Support running multiple builds for one commit sha


KrauseFx:
Proposal is: e.g. if you manually trigger a build, the most recent commit from a given branch (e.g. master) is being used to run tests. If there is an existing build, we don't want to overwrite it, as the other build is still valid. Instead it should be possible to run a build for the same commit

Issue #211, opened 2018-03-05
Add a unique ID for each build


KrauseFx:
Main thing that's missing to get this in a rather solid state is the identification of a build. As raised the question in Slack (`I see the Build object doesn't have an ID like we have for all other objects. Is there a reason for that? Did we decide you can't run multiple builds for one sha? Is project + sha unique?)`). My proposal is to add a unique ID to builds also, so that this ID in combination with the project is unique (or even without the project?). Once that's done, multiple `# TODO` items from this PR can be resolved. (e.g. see https://github.com/fastlane/ci/pull/200/files#diff-40d0c75f55437cd62b7224bd7d90d275R14)

PR #220, opened 2018-03-05
Migrate from sha-based identification to build number, allow multiple builds per sha

KrauseFx:
Fixes https://github.com/fastlane/ci/issues/219
Fixes https://github.com/fastlane/ci/issues/211

This allows us to
- Identify each build with `project.id` + `build.number`
- Properly identify a build for real time streaming (access control missing)
- Properly fetch existing running test runners and get their output
- Automatically redirect to the build output right after triggering a new build manually
- Run multiple builds for one commit sha

### DIFF
--- a/features/build/build_controller.rb
+++ b/features/build/build_controller.rb
@@ -11,16 +11,16 @@ module FastlaneCI
 
     get "#{HOME}/*" do |project_id, build_id|
       project = self.user_project_with_id(project_id: project_id)
-      build = project.builds.find { |b| b.sha == build_id } # TODO: We need a build ID, sha isn't enough
+      build = project.builds.find { |b| b.id == build_id }
 
       current_runner_service = TestRunnerService.test_runner_services.find do |t|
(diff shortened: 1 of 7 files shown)


minuscorp:
Does the file JSON file name change? Or we'll keep it as is?

KrauseFx:
@minuscorp ah interesting. I guess `project.id` + `build.number` is also unique, right? Should we still have the `build.id` then, or should we leave the combination of project + build number? I'm ok either way.

If we want the `build.id` (just like with `project.id`), should we rename the JSON files to `build.id` instead of `build.number`? cc @taquitos

taquitos:
I think we should have the build filename and ID be auto-incrementing like 1, 2, 3...
Because the uniqueness of a build comes from `project.id+build.id` I think it is sufficient that each build log have a simple number and not a `SecureRandom.uuid`

KrauseFx:
Alright, thanks @taquitos, I'm on it 👍

PR #221, opened 2018-03-05
`users_controller` and `provider_credentials` controllers for managing and displaying the `users` resource

AndrewMcBurney:
### Motivation

- To allow a way for authenticated users to edit the `users.json` configuration values from the CI server.
- This was originally intended to be part of #80. However,  the PR was becoming bloated, and the original intentions were muddled with multiple, disjoint purposes.

### Changes

- Adds two controllers (`users_controller` and `provider_credentials_controller`) for creating, updating, and displaying users:
  - Adds view-partial forms for creating and mutating users and provider credentials
  - Form submission changes will be persisted to the `users.json` file in the remote git configuration repo

### TODO

- [x] Write tests for the `provider_credentials_service`
- [x] Make the UI view partials for `users` and `provider_credentials` nicer

### Updates For Future PRs

- Currently, the `JSONUserDataSource` persists changes locally. However, it doesn't persist the changes to the git configuration repo. In a future pull request, I'd like to make it so the `JSONUserDataSource` pushes changes to the private git configuration repo (just like the `JSONProjectDataSource` does)
- Form validations for creating resources, and handling error cases
- Add `delete` actions for deleting `users` and `provider_credentials` (to make resources fully CRUD)


shared/json_convertible.rb
```diff
@@ -29,7 +29,7 @@ def to_object_dictionary(ignore_instance_variables: [])
           if self.instance_variable_get(var).kind_of?(Array)
             object_array = []
             self.instance_variable_get(var).each do |obj|
-              object_array << obj.to_object_dictionary
+              object_array << (obj.kind_of?(Hash) ? obj : obj.to_object_dictionary)
```
minuscorp: Is this change still necessary?

AndrewMcBurney: Ahh, yeah - the change is still necessary because I'm serializing the `ProviderCredential`s to a `Hash` first, and then serializing the `User`. I documented the changes in my [most recent commit](https://github.com/fastlane/ci/pull/221/commits/3c3d7e3ba9ad9c6e0c894658c35342e0d5c6c02f). When you get a chance, could you please let me know if the documentation is good, or if there is anything I should change with the documentation or the code? Thank you!

minuscorp: I'm kind of confused, because it can happen that many types does not include that method. So I think the smartest way to proceed is checking if the `obj.class.include?(JSONConvertible)`. What do you think?

AndrewMcBurney: Ahh, I like that suggestion much better than what I implemented. Good call!

minuscorp: Is like I forgot the safe fallback to types that does not include `JSONConvertible`. Could you add a test case that covers this case? 😄

AndrewMcBurney: Will do! :)

features/credentials/provider_credentials_controller.rb
```diff
@@ -0,0 +1,88 @@
+require "set"
+
+require_relative "../../shared/authenticated_controller_base"
+require_relative "../../services/services"
+
+module FastlaneCI
+  # A CRUD controller to manage provider credentials associated with a user
```
KrauseFx: A comment here would be a good on what a provider credential is

spec/shared/json_convertible_spec.rb
```diff
@@ -210,6 +217,14 @@ module FastlaneCI
                                              ] })
     end
 
+    it "Allows to encode objects with array properties of a type that does not include the JSONConvertible mixin" do
```
minuscorp: ❤️

services/user_service.rb
```diff
@@ -24,13 +24,13 @@ def initialize(user_data_source: nil)
...
 
-    def create_user!(email: nil, password: nil)
+    def create_user!(id: nil, email: nil, password: nil)
       email = email.strip
 
       unless self.user_data_source.user_exist?(email: email)
         logger.debug("creating account #{email}")
         provider_credential = GitHubProviderCredential.new(email: email)
-        return self.user_data_source.create_user!(email: email, password: password, provider_credential: provider_credential)
+        return self.user_data_source.create_user!(id: nil, email: email, password: password, provider_credential: provider_credential)
```
minuscorp: `id: nil` -> `id: id`?

AndrewMcBurney: oops, great catch! 😳

features/credentials/provider_credentials_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    private
+
+    #####################################################
+    # @!group Data: View-specific data
+    #####################################################
+
+    # @return [Array[User]]
+    def users
+      Services.user_service.user_data_source.users
```
taquitos: services are meant to abstract away the details of their datasources. So for this one, let's not reach in to the service's properties(`user_data_source`), instead, let's expose a method like `def users`

features/credentials/provider_credentials_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+    def user_credentials
+      users
+        .map { |user| [user.id, user.provider_credentials] }
+        .to_h
+    end
+
+    # Empty provider credential
+    #
+    # @return [GitHubProviderCredential]
+    def new_credential
```
taquitos: `new_credential` method name is confusing, I would assume a credential is created every time I call it. Maybe you could make this a little more descriptive. I'm totally down with longer method/property names if it adds clarity.

features/credentials/provider_credentials_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+      @new_credential ||= GitHubProviderCredential.new
+    end
+
+    #####################################################
+    # @!group Locals: View-specific locals
+    #####################################################
+
+    # @return [Set[Symbol]]
+    def provider_credential_params
```
taquitos: Let's make a more descriptive name + comment about how it's used. I think I understand this is used for parameter validation, right? I might even go for `post_parameter_list_for_validation` or something like.

features/credentials/provider_credentials_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+      Set.new(%w(user_id id email api_token provider_name type full_name))
+    end
+
+    #####################################################
+    # @!group Validations: Controller validations
+    #####################################################
+
+    # Validates that the user you're trying to associate the credential with
```
taquitos: This comment isn't totally accurate, this code only checks if the list of `users` has a user with an `id` matching the `user_id`. Where it is used, I'm sure this is the case 😄 I'd move that comment and rename this method to `user_exists_with_id?` or something like that

features/credentials/views/provider_credentials.erb
```diff
@@ -0,0 +1,69 @@
...
+  <%= erb :"../../partials/navigation" %>
+
+  <main class="mdl-layout__content" style="padding: 40px">
+    <!-- List all existing provider credentials for a user -->
+
+    <h3>Provider Credentials</h3>
+    <table>
+      <thead>
+        <tr>
+          <th>id</th>
```
taquitos: I think for the headers, we could make this a little less ruby and more user friendly, like instead of `api_token` it could just be `API Token`. No big deal either way, just thinking that since we're displaying this, it might be nicer to look at.

features/users/users_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    private
+
+    #####################################################
+    # @!group Data: View-specific data
+    #####################################################
+
+    # @return [Array[User]]
+    def users
+      Services.user_service.user_data_source.users
```
taquitos: same comment here about accessing what should be private properties in the `user_service`

features/users/users_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    # Empty user object for new user form
+    #
+    # @return [User]
+    def new_user
```
taquitos: same here, name is confusing to me.

features/users/users_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      @new_user ||= User.new(provider_credentials: [GitHubProviderCredential.new])
```
taquitos: Also, does this support adding multiple new users?

features/users/users_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    #####################################################
+    # @!group Params: View parameters required
+    #####################################################
+
+    # @return [Set[Symbol]]
+    def users_params
```
taquitos: Same naming comment

features/users/views/forms/_users.erb
```diff
@@ -0,0 +1,42 @@
...
+
+  <td>
+    <input type="text" name="password_hash" value=<%= user.password_hash %>>
+  </td>
+
+  <td>
+    <table>
+      <thead>
+        <tr>
+          <th><i>id</i></th>
```
taquitos: same header comment

services/data_sources/json_user_data_source.rb
```diff
@@ -60,6 +64,25 @@ def users
       end
     end
 
+    def users=(users)
```
taquitos: I think I would prefer this to be `save_users!` instead of a property

services/provider_credential_service.rb
```diff
@@ -0,0 +1,59 @@
+require_relative "./services"
+require_relative "../shared/logging_module"
+
+module FastlaneCI
+  # Provides access to provider credential related methods
+  class ProviderCredentialService
```
taquitos: I'm not sold on the idea that `ProviderCredentialService` should be separate from the `UserService`. I think because `User` owns `ProviderCredential`s, it might make sense to continue to manage that in the actual `UserService`. 
What do you think?

services/provider_credential_service.rb
```diff
@@ -0,0 +1,59 @@
...
+          id: user.id,
+          email: user.email,
+          password_hash: user.password_hash,
+          provider_credentials: user.provider_credentials.push(provider_credential)
+        )
+        Services.user_service.update_user!(user: new_user)
+      end
+    end
+
+    def update_provider_credential!(
```
taquitos: This is trickier than it seems...Provider credentials can have different email addresses from the user it is associated with. 
For example: I might use `taquitos@google.com` as my fastlane.ci account, but I want to use `taquitos@gmail.com` for my GitHub credentials.
This is why we have a back reference to `credential.ci_user`. I think the better move is to lookup the user you want to add a credential to, then create the credential with the proper credential bits, and then save that back.

services/provider_credential_service.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      if user.nil?
+        logger.error("Can't update provider credential for user, since user does not exist.")
+      else
+        new_user = User.new(
+          id: user.id,
+          email: user.email,
+          password_hash: user.password_hash,
+          provider_credentials: user.provider_credentials
+            .delete_if { |credential| credential.id == id }
```
taquitos: I think we should break this out into multiple steps. This is clever, for sure, but I think it will be more understandable if we make it a little more verbose

shared/controller_base.rb
```diff
@@ -79,5 +80,34 @@ def setup_common_controller_configuration(is_called_during_reload: false)
...
       # end
     end
+
+    protected
+
+    # View helper
+    #
+    # @return [String]
+    def new_uuid
+      SecureRandom.uuid
```
taquitos: I think this should live as API in a service, that ultimately asks the data_store, that would be the best move. This is because the backing data store might have a different idea about what a UUID should be. For example, a database might actually prefer auto-incrementing int64, so if we move this to datastore, it can do what it needs.

shared/models/github_provider_credential.rb
```diff
@@ -8,12 +8,13 @@ module FastlaneCI
   class GitHubProviderCredential < ProviderCredential
     attr_accessor :encrypted_api_token
 
-    def initialize(email: nil, full_name: nil, api_token: nil)
+    def initialize(id: nil, email: nil, full_name: nil, api_token: nil, provider_name: nil, type: nil)
```
taquitos: let's revert this change. If you let people provide different `provider_name` and `type` then this is no longer a `GitHubProviderCredential` it becomes more generic. We want to keep this as specific as possible. If we need to allow different `provider_name` and `type` let's create a new `ProviderCredential` subclass to represent that.

features/users/users_controller.rb
AndrewMcBurney: Yep! So the way this works is in the view form, this `new_user` is [passed into the form partial](https://github.com/AndrewMcBurney/ci/blob/users_resource/features/users/views/users.erb#L54) to populate the [`create` form](https://github.com/AndrewMcBurney/ci/blob/users_resource/features/users/views/forms/_users.erb#L1).

It's not actually the same `User` that's created in the `post "#{HOME}/create"` action, which is why I memoized it (to avoid creating a new user in memory each time a `/users` view is rendered). I agree the name is kinda confusing, so maybe something like naming this method `blank_form_user` as well as a descriptive comment about where it's used would clear things up a bit better.

My logic for this was that I'd be able to use the user form for both the `create` and `update` actions.

shared/controller_base.rb
AndrewMcBurney: Oh, oops. I forgot to remove this method. I'm not using it anymore. I really like the idea you suggested though. Would it be possible to do that in a separate PR?

services/provider_credential_service.rb
AndrewMcBurney: Hmm. That's a good point. My intention was to keep them separate for sake of Single Responsibility Principle (`UserService` is only responsible for users, `ProviderCredentialService` is only responsible for provider credentials), but that's a good point about the `User` owning the `ProviderCredential`. If you want, I can move these two methods to the `UserService`.

services/provider_credential_service.rb
AndrewMcBurney: I'm doing the user lookup in the method by [`user_id`](https://github.com/fastlane/ci/pull/221/files/37ab9934a7fcfa659deeda4d3bbc5064140d05e8#diff-a47cf038d988ee52e0479961109c6133R42) instead of by `email`. In the `update` view form for provider credentials, the `user_id` associated with the credential is [passed into the form](https://github.com/fastlane/ci/pull/221/files/37ab9934a7fcfa659deeda4d3bbc5064140d05e8#diff-c4965ee0be82e539c898652971f42e28R8).

taquitos: Ahh that makes sense. Carry on 😄

shared/controller_base.rb
taquitos: 👍

services/provider_credential_service.rb
taquitos: I like single responsible principle (in principle 🙃), but because users are ultimately the holders of credentials, I think it makes the most sense to bundle it together.

features/users/users_controller.rb
taquitos: Oh, that's great. I like that name too 💯

services/provider_credential_service.rb
AndrewMcBurney: Sounds good to me! :)

Issue #216, opened 2018-03-05
Refactor `TestRunnerService` to only have one instance of `TestRunnerService` that holds references to all test runners


KrauseFx:
As seen in https://github.com/fastlane/ci/pull/200/files#diff-62e64a60028b7cd0fdf6248bbc3f9bfcR13, we currently have to store references to all active `test_runner_services` to keep them alive, and to be able to access them when the user wants to see their progress. I added a comment with a proposal on how we can properly architect it

We probably want a single instance of `TestRunnerService` (like we do for other services), that holds references to all the active `TestRunners`

minuscorp:
Yep, I'm totally with you on this. A single `RunnerService` must live throughout the application, which holds references to each `Runner` that might be running at a given moment. One service, many runners.

taquitos:
@KrauseFx what are the responsibilities that the new refactored `TestRunneService` will have?

KrauseFx:
@taquitos 

So I was thinking of doing the following

### TestRunnerService
Responsible for
- Manage all `TestRunner` objects that are active (either in queue or actually running)
- Measures the time of a `TestRunner`'s execution
- Stores the `Build` information in version control and triggers the report of the build status on GitHub

### TestRunner
(e.g. `FastlaneTestRunner`) Responsible for
- Loading up _fastlane_ and running a lane with it, checking the return status
- Raise an exception if build fails, with information that can be handled by `TestRunnerService`
- Reporting back a list of artifacts  to `TestRunnerService`

Thoughts?

taquitos:
I love it. One question, and it's a minor one, should we rename them to `BuildRunnerService` and `BuildRunner`?

KrauseFx:
I don't feel strongly about the name `BuildRunnerService`, so that works. Any specific reason why it's not a service? I thought it behaves similarly to the other `fastlane.ci` services we already have, as in it stays alive and manages things (state :trollface:)

taquitos:
I was just saying `TestRunnerService` -> `BuildRunnerService` and also `TestRunner` -> `BuildRunner` to align us around the `build` terminology instead of `test` and `job`.

KrauseFx:
Ah right, sorry misunderstood your message, yeah that works 👍

KrauseFx:
Will work on that refactor today

PR #230, opened 2018-03-07
Refactor BuildRunnerService

KrauseFx:
Fixes https://github.com/fastlane/ci/issues/216

Follow-up tasks:
- [ ] Think about how we want to implement the `rerun` method, do we actually need it? Should we just trigger a new build instead, now that we can do that?

Open tasks, could be a separate PR to unblock peeps:
- Update `launch.rb#run_pending_github_builds` to use new architecture ([link](https://github.com/fastlane/ci/blob/7bf2ebebde73bd16f7fcb0f78b7d1d381a23a47b/launch.rb#L179))
- Update `github_worker_base.rb#create_and_queue_build_task` to use new architecture ([link](https://github.com/fastlane/ci/blob/7bf2ebebde73bd16f7fcb0f78b7d1d381a23a47b/workers/github_worker_base.rb#L96))


features/build_runner/build_runner.rb
```diff
@@ -1,11 +1,185 @@
...
+
+    def initialize(project: nil, sha: nil, github_service: nil)
+      self.project = project
+      self.sha = sha
+
+      self.all_build_output_log_lines = []
+      self.build_change_observer_blocks = []
+
+      # TODO: provider credential should determine what exact CodeHostingService gets instantiated
+      self.code_hosting_service = github_service
```
KrauseFx: @taquitos I'd need your advice on how/where we remember the `code_hosting_service`, which we use to set the build status. Currently it's part of each `BuildRunner` object, allowing us to use a different `code_hosting_service` (e.g. different projects have different code hostings)

features/build_runner/build_runner.rb
```diff
@@ -1,11 +1,185 @@
...
+      # For setting the build status, if that fails, it's fine
+      # as the source of truth is the git repo
+      raise ex
+    end
+
+    # Let GitHub know about the current state of the build
+    # Using a `rescue` block here is important
+    # As the build is still green, even though we couldn't set the GH status
+    def update_build_status_source!
+      return # TODO: enable again
```
KrauseFx: Remove this `return` once https://github.com/fastlane/ci/pull/230/files#r172956779 is resolved

features/build_runner/build_runner.rb
```diff
@@ -1,11 +1,190 @@
...
+  # - Offer a way to subscribe to new lines being added to the output (e.g. to stream them to the user's browser)
+  #
   class BuildRunner
+    include FastlaneCI::Logging
+
+    # Reference to the FastlaneCI::Project of this particular build run
+    attr_accessor :project
+
+    # The code hosting service we want to report the status back to
+    # TODO: this probably shouldn't be stored here, think about using
```
KrauseFx: Update this as https://github.com/fastlane/ci/pull/230/files#r172956779 is resolved

features/build_runner/build_runner.rb
taquitos: Can you explain the logic flow you expect? I imagine it's 1:1 `BuildRunner` to `CodeHostingService`, so in this case, I don't see a problem with keeping it in the `BuildRunner`

features/build_runner/build_runner.rb
```diff
@@ -1,11 +1,185 @@
...
+    # Responsible for updating the build status in our local config
+    # and on GitHub
+    def update_build_status!
+      update_build_status_locally!
+      update_build_status_source!
+    end
+
+    # Handle a new incoming row, and alert every stakeholder who is interested
+    def new_row(row)
+      logger.debug(row["message"])
```
taquitos: `logger.debug(row["message"]) if row["message"]`

features/build_runner/build_runner.rb
```diff
@@ -1,11 +1,185 @@
...
+      Services.build_service.add_build!(
+        project: self.project,
+        build: self.current_build
+      )
+
+      # Commit & Push the changes to git remote
+      FastlaneCI::Services.project_service.git_repo.commit_changes!
+    rescue StandardError => ex
+      logger.error("Error setting the build status as part of the config repo")
+      logger.error(ex.to_s)
```
taquitos: `logger.error(ex)` is ok here too

features/build_runner/build_runner.rb
```diff
@@ -1,11 +1,185 @@
...
+      logger.error(ex.backtrace.join("\n"))
```
taquitos: this isn't needed anymore.

features/build_runner/build_runner.rb
```diff
@@ -1,11 +1,185 @@
...
+      self.code_hosting_service.set_build_status!(
+        repo: self.project.repo_config.git_url,
+        sha: self.sha,
+        state: self.current_build.status,
+        status_context: self.project.project_name
+      )
+    rescue StandardError => ex
+      logger.error("Error setting the build status on remote service")
+      logger.error(ex.to_s)
+      logger.error(ex.backtrace.join("\n"))
```
taquitos: same here

features/project/project_controller.rb
```diff
@@ -26,18 +26,12 @@ class ProjectController < AuthenticatedControllerBase
...
-      Thread.new do
-        build_runner_service.run
-      end
+      # TODO: pass GitHub service?
+      build_runner = FastlaneBuildRunner.new(project: project, sha: current_sha, github_service: nil)
+      build_runner.setup(platform: "ios", lane: "beta", parameters: nil) # specific to fastlane
+      Services.build_runner_service.add_build_runner(build_runner: build_runner)
 
-      redirect("#{HOME}/#{project_id}/builds/#{build_runner_service.current_build.number}")
+      redirect("#{HOME}/#{project_id}/builds/#{build_runner.current_build.number}")
```
taquitos: let's not reach into the runner to grab a build, and then number. Instead I think it makes sense to call `build_runner.current_build_number` instead of exposing `current_build`. I'd rather keep `current_build` private if we can.

services/build_runner_service.rb
```diff
@@ -2,234 +2,41 @@
...
-          type: artifact[:type],
-          reference: artifact[:path],
-          provider: self.project.artifact_provider
-        )
-      end
-                                .map do |artifact|
-        self.project.artifact_provider.store!(artifact: artifact, build: self.current_build, project: self.project)
+      # TODO: not the best approach to spawn a thread
+      # Use TaskQueue instead
+      Thread.new do
```
taquitos: ⛔️

services/build_runner_service.rb
```diff
@@ -2,234 +2,41 @@
...
-      start_time = Time.now
-
-      self.current_build = build
-      update_build_status!
-
-      logger.debug("Running runner now")
-      artifact_paths = build_runner.run do |current_row|
-        new_row(current_row)
-      end
+    def find_build_runner(project_id: nil, build_number: nil)
```
taquitos: `find_build_runner(project_id:, build_number:)` since we are using 2.3.3 🎉

features/build/build_controller.rb
```diff
@@ -16,15 +16,18 @@ class BuildController < AuthenticatedControllerBase
...
+      )
+
+      raise "Couldn't find build runner for project #{project_id} with build_number #{build_number}" if current_build_runner.nil?
 
       locals = {
         project: project,
         build: build,
         title: "Project #{project.project_name}, Build #{build.number}",
-        existing_rows: current_runner_service.all_build_output_log_lines.collect { |a| a[:html] }
+        existing_rows: current_build_runner.all_build_output_log_lines.collect { |a| a[:html] }
```
taquitos: it might make sense to move this logic into the `current_build_runner` and expose it via `current_build_runner.all_build_output(format: :html)` or something like that

features/build/build_websocket_backend.rb
```diff
@@ -53,22 +53,17 @@ def call(env)
...
-          end)
-        end
+        current_build_runner = Services.build_runner_service.find_build_runner(
+          project_id: project_id,
+          build_number: build_number
+        )
+
+        # TODO: Think this through, do we properly add new listener, and notify them of line changes, etc.
+        #       Also how does the "offboarding" of runners work once the tests are finished
+        current_build_runner.add_listener(proc do |row|
```
taquitos: I think it's clearer as
`current_build_runner.add_listener(proc { |row| ws.send(row.to_json)})` or whatevs the syntax is if this isn't right

features/build/build_controller.rb
KrauseFx: Yeah, great idea. It might not be worth it for now, as we talked about getting rid of the server-side HTML rendering for build outputs though, should we keep it like this for now?

features/build_runner/build_runner.rb
```diff
@@ -1,11 +1,185 @@
...
+    end
+
+    # Starts the build, incrementing the build number from the number of builds
+    # for a given project
+    #
+    # @return [nil]
+    def start
+      start_time = Time.now
+
+      logger.debug("Running runner now")
```
taquitos: Let's add some identifiable information here so we know what runner this is

features/build_runner/build_runner.rb
```diff
@@ -1,11 +1,185 @@
...
+
+      duration = Time.now - start_time
+
+      current_build.duration = duration
+      current_build.status = :success
+
+      self.update_build_status!
+    rescue StandardError => ex
+      # TODO: better error handling, don't catch all Exception
+      puts(ex)
```
taquitos: logger.error

features/build_runner/build_runner.rb
```diff
@@ -1,11 +1,185 @@
...
+      self.build_change_observer_blocks << block
+    end
+
+    def prepare_build_object
+      builds = Services.build_service.list_builds(project: self.project)
+
+      if builds.count > 0
+        new_build_number = builds.sort_by(&:number).last.number + 1
+      else
+        new_build_number = 1 # do we start with 1?
```
taquitos: yes please

features/build/build_websocket_backend.rb
KrauseFx: That's what I have so far. Or do you refer to the one line approach vs multi-line? The reason for multi-line is that I need to add an additional auth check probably here, to avoid people accessing logs from builds they don't have access to

features/build_runner/build_runner.rb
KrauseFx: Yesss

features/project/project_controller.rb
KrauseFx: Nice! 👍

features/build_runner/build_runner.rb
KrauseFx: Ah yes, sorry totally my bad, I was confused about what this is being used, pushing a change now

features/build/build_websocket_backend.rb
taquitos: I was referring to the one-line approach if you have just 1 single statement. Since you're going to add stuff, you can ignore me here 👍

features/build/build_controller.rb
taquitos: Good point, let's keep it like this then 🐐

features/build_runner/build_runner.rb
```diff
@@ -62,7 +62,7 @@ def setup
     def start
       start_time = Time.now
 
-      logger.debug("Running runner now")
+      logger.debug("Starting runner #{self.class} now...")
```
taquitos: instead of self.class (already reported as part of logger.debug) let's use `self.project.name` or something like that so we can see what is happening easier when we have a bunch of runners

taquitos:
`Think about how we want to implement the rerun method, do we actually need it? Should we just trigger a new build instead, now that we can do that?`
Let's trigger a new build instead. Reduces complexity and I think it's the `right thing`™️

features/build_runner/build_runner.rb
KrauseFx: Ah yeah

KrauseFx:
It's merged 🎉

features/build_runner/build_runner.rb
```diff
@@ -65,7 +65,7 @@ def setup
     def start
       start_time = Time.now
 
-      logger.debug("Starting runner #{self.class} now...")
+      logger.debug("Starting runner #{self.class} for #{self.project.project_name} (self.project.id) now...")
```
taquitos: 😻

PR #232, opened 2018-03-07
Improve logger messages when they are arrays

taquitos:
Instead of printing out the array, we'll join the contents and print. That makes exception printing much nicer when we do logger.error(ex)

### DIFF
--- a/shared/logging_module.rb
+++ b/shared/logging_module.rb
@@ -46,6 +46,9 @@ module FastlaneCI
             thread_id = " #{Thread.current[:thread_id]}" unless Thread.current[:thread_id].nil?
           end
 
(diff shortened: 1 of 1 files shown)


shared/logging_module.rb
```diff
@@ -46,6 +46,9 @@ def logger
             thread_id = " #{Thread.current[:thread_id]}" unless Thread.current[:thread_id].nil?
           end
 
+          if msg.class == Array
```
KrauseFx: Types 😍

taquitos: In my email this comment looks way better:

screen shot 2018-03-07 at 11 58 38 am

Issue #226, opened 2018-03-06
Create GoogleCloudArtifactProvider


minuscorp:
No much to describe, I think this should be done for MVP, what do you think? 🚀

minuscorp:
Fixed by #235

PR #235, opened 2018-03-08
First approach to GCPStorageArtifactProvider

minuscorp:
- [x] Test that this actually works.

- [X] Refactor `ArtifactProvider` `retrieve!` method


minuscorp:
Fixes #226

features/build_runner/fastlane_build_runner.rb
```diff
@@ -89,21 +89,23 @@ def run
...
-        # artifact_paths = []
-        # artifact_paths << { type: "log", path: "fastlane.log" }
-        # constants_with_path = Fastlane::Actions::SharedValues.constants
-        #                                                      .select { |value| value.to_s.include?("PATH") } # Far from ideal, but meanwhile...
-        #                                                      .select { |value| !Fastlane::Actions.lane_context[value].nil? && !Fastlane::Actions.lane_context[value].empty? }
-        #                                                      .map { |value| { type: value.to_s, path: Fastlane::Actions.lane_context[value] } }
-        # artifact_paths.concat(constants_with_path)
+        artifact_paths = []
+        artifact_paths << { type: "log", path: "fastlane.log" }
+        constants_with_path = Fastlane::Actions::SharedValues.constants
```
taquitos: Let's add some comments about what is happening here and why we need to filter out stuff

features/project/project_controller.rb
```diff
@@ -116,8 +116,19 @@ class ProjectController < AuthenticatedControllerBase
       lane = params["selected_lane"]
       project_name = params["project_name"]
 
+      # Do this so we trigger the clone of the repo.
+      # TODO: Do this wherever it should be done, as we must redirect
+      # to the project details only when this task is finished.
+      _ = GitRepo.new(
```
taquitos: This should be moved into a service

shared/models/local_artifact_provider.rb
```diff
@@ -57,12 +57,9 @@ def store!(artifact: nil, build: nil, project: nil)
 
     def retrieve!(artifact: nil)
       raise "Artifact to store was not provided or wrong type provided" if artifact.nil? || artifact&.class&.is_a?(Artifact)
-      raise "Build was not provided or wrong type provided" if build.nil? || build&.class&.is_a?(Build)
-      raise "Project was not provided or wrong type provided" if project.nil? || project&.class&.is_a?(Project)
-
       raise "Artifact reference not found" unless File.exist?(artifact.reference)
 
-      return artifact.reference
+      artifact.reference
```
taquitos: explicit `return` is preferred

shared/models/gcp_artifact_provider.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    def initialize(cloud_project: nil, json_keyfile_path: nil, bucket_name: nil)
+      self.cloud_project = cloud_project
+      self.json_keyfile_path = json_keyfile_path
+      self.bucket_name = bucket_name
+      self.class_name = self.class.name.to_s
+    end
+
+    def init_storage!
+      # TODO: This will break when google-cloud-storage is updated.
```
taquitos: What do you mean this will break when ` google-cloud-storage is updated`? Why is that?

shared/models/gcp_artifact_provider.rb
```diff
@@ -0,0 +1,93 @@
...
+      self.storage = Google::Cloud::Storage.new(
+        project: self.cloud_project,
+        keyfile: File.expand_path(self.json_keyfile_path)
+      )
+      self.bucket = self.storage.bucket(self.bucket_name)
+      permissions = self.bucket.test_permissions("storage.objects.create", "storage.objects.get")
+      raise "Not enough permissions to the requested bucket" \
```
taquitos: Let's add more information in the exception message so we can help people figure out what exactly is wrong and how they can fix it.

shared/models/gcp_artifact_provider.rb
```diff
@@ -0,0 +1,93 @@
...
+        unless permissions.include?("storage.objects.create") && permissions.include?("storage.objects.get")
+    end
+
+    def store!(artifact: nil, build: nil, project: nil)
+      raise "Artifact to store was not provided or wrong type provided" if artifact.nil? || artifact&.class&.is_a?(Artifact)
```
taquitos: best messages ❤️

shared/models/gcp_artifact_provider.rb
```diff
@@ -0,0 +1,93 @@
...
+      raise "Build was not provided or wrong type provided" if build.nil? || build&.class&.is_a?(Build)
+      raise "Project was not provided or wrong type provided" if project.nil? || project&.class&.is_a?(Project)
+
+      init_storage!
+
+      original_artifact_reference = Pathname.new(artifact.reference)
+
+      raise "Artifact reference not found." unless File.exist?(original_artifact_reference)
```
taquitos: I think this error message could be improved. I don't understand what the problem is by reading it.

shared/models/gcp_artifact_provider.rb
```diff
@@ -0,0 +1,93 @@
...
+      new_artifact_reference = File.join(project.id, build.number.to_s, artifact.type, File.basename(original_artifact_reference))
+
+      file = self.bucket.create_file(original_artifact_reference.to_s,
+                                     new_artifact_reference)
+
+      raise "File couldn't be created" if file.nil?
+
+      artifact.reference = new_artifact_reference
+      artifact.provider = self
+      artifact # This is the Artifact that we will store in the build.
```
taquitos: return statement

shared/models/gcp_artifact_provider.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      init_storage!
+
+      file = self.bucket.file(artifact.reference)
+      raise "Artifact reference not found" if file.nil?
+
+      # TODO: For now is ok to return just the link to the browser in the GCP Console itself,
+      # this should be revisited later to address how we generate permanent public links to the artifact
+      # if the bucket is public (read-public) or expiring links for private buckets.
+      "#{self.class.root_browser}#{self.bucket_name}/#{artifacts.first.reference.split('/')[0..-2].join('/')}"
```
taquitos: return statement

shared/models/gcp_artifact_provider.rb
minuscorp: So _fastlane_ depends on a certain version of `google-api-client` pretty outdated, and `google-cloud-storage` too, so I had to downgrade the version to match the requirements of _fastlane_, there is a PR open to update it, but it might last long before getting merged.

features/project/project_controller.rb
minuscorp: I'm really thinking that in a follow up PR, refactoring all the logic between project's repos and ci-repo itself. They use the same underneath logic but their use cases are totally different. Bot I think is a little bit out of the scope of this PR, What do you think?

taquitos: I think that's a good idea. That being said, this shouldn't exist here 😄 controllers are meant to handle display logic, and to call into services which are meant to handle business logic. Mixing display with business logic, and in this case, relying on a side-effect to perform some business logic is extra dangerous.

shared/models/gcp_artifact_provider.rb
taquitos: Ahhh, let's add that comment into the code then, and you can specify exactly which versions you are talking about. That way if it does break we have some extra knowledge and context around it.

features/project/project_controller.rb
minuscorp: May I cover this in a separate PR as it involves pretty big refactoring?

taquitos: Yup, I just wouldn't want this to stick around and be forgotten 😄

minuscorp: Written #280 in order to tackle this soon!

taquitos: 👍 thanks

PR #237, opened 2018-03-08
Migrate remaining code to use new build architecture

KrauseFx:
Follow-up for https://github.com/fastlane/ci/pull/230


launch.rb
```diff
@@ -11,12 +11,8 @@ module FastlaneCI
   class Launch
     class << self
       include FastlaneCI::Logging
-
-      attr_accessor :build_queue
```
taquitos: neat, it looks like `add_build_runner` takes care of queuing up stuff, that's great!

Issue #241, opened 2018-03-08
[Proposal] Have separate class to access ENV variables


KrauseFx:
proposed by @taquitos, already implemented for fastlane by @hjanuschka 
better than just strings

KrauseFx:
cc @AndrewMcBurney

PR #244, opened 2018-03-08
Wrap FastlaneCI-specific environment variables in their own class

AndrewMcBurney:
https://github.com/fastlane/ci/issues/241

Wraps environment variables in their own class.

```ruby
FastlaneCI.env.repo_url
```
> Example usage

### DIFF
--- a/features/configuration/configuration_controller.rb
+++ b/features/configuration/configuration_controller.rb
@@ -73,14 +73,7 @@ module FastlaneCI
 
     # @return [Hash]
     def keys
-      {
-        encryption_key: ENV["FASTLANE_CI_ENCRYPTION_KEY"],
-        ci_user_email: ENV["FASTLANE_CI_USER"],
-        ci_user_api_token: ENV["FASTLANE_CI_PASSWORD"],
-        repo_url: ENV["FASTLANE_CI_REPO_URL"],
-        clone_user_email: ENV["FASTLANE_CI_INITIAL_CLONE_EMAIL"],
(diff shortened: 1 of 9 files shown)


features/configuration/configuration_controller.rb
```diff
@@ -73,14 +73,7 @@ class ConfigurationController < AuthenticatedControllerBase
...
     def keys
-      {
-        encryption_key: ENV["FASTLANE_CI_ENCRYPTION_KEY"],
-        ci_user_email: ENV["FASTLANE_CI_USER"],
-        ci_user_api_token: ENV["FASTLANE_CI_PASSWORD"],
-        repo_url: ENV["FASTLANE_CI_REPO_URL"],
-        clone_user_email: ENV["FASTLANE_CI_INITIAL_CLONE_EMAIL"],
-        clone_user_api_token: ENV["FASTLANE_CI_INITIAL_CLONE_API_TOKEN"]
-      }
+      FastlaneCI.env.all
```
KrauseFx: Eventually we probably want to white-list some of them, instead of just exposing all I'd think?

taquitos: What do you mean white-list some of them?

KrauseFx: Similar to how Rails handles parameters, instead of just taking all the `POST` parameters and trying to map it to a model, you define a list of allowed parameters, if they're there, good, if not also fine.

So instead of saying `.env.all`, you'd still have a list of env variables that we allow setting and try to map here, but haven't thought this through. actually feel free to ignore my comments here

PR #253, opened 2018-03-09
Fixes with startup, many edge-cases were solved.

minuscorp:


services/config_data_sources/json_project_data_source.rb
```diff
@@ -135,6 +135,8 @@ def job_triggers_from_hash_array(job_trigger_array: nil)
     end
 
     def projects=(projects)
+      require "pry"
```
KrauseFx: ☝️

services/config_data_sources/json_project_data_source.rb
```diff
@@ -164,6 +166,7 @@ def create_project!(name: nil, repo_config: nil, enabled: nil, lane: nil, artifa
                                 project_name: name,
                                 lane: lane,
                                 artifact_provider: artifact_provider)
+      _ = GitRepo
```
KrauseFx: What does this do

services/config_data_sources/json_project_data_source.rb
```diff
@@ -135,6 +135,8 @@ def job_triggers_from_hash_array(job_trigger_array: nil)
     end
 
     def projects=(projects)
+      require "pry"
+      binding.pry
```
AndrewMcBurney: Could you please remove the `binding.pry` please :)

shared/models/git_repo.rb
```diff
@@ -294,6 +294,7 @@ def commit_changes!(commit_message: nil, file_to_commit: nil, repo_auth: self.re
           logger.debug("done commit_changes! #{self.git_config.full_name}")
         end
       end
+      wait_for_task_to_complete(task: task) unless async
```
KrauseFx: Instead of this, could we make use of block based Ruby code? Providing a callback?

services/services.rb
```diff
@@ -105,6 +105,8 @@ def self.provider_credential
 
     # Start up a ProjectService from our JSONProjectDataSource
     def self.project_service
+      require "pry"
```
AndrewMcBurney: Same here, could you please remove the `binding.pry`?

services/data_sources/json_user_data_source.rb
```diff
@@ -19,6 +19,8 @@ def self.attribute_to_type_map
 
     def self.map_enumerable_type(enumerable_property_name: nil, current_json_object: nil)
       if enumerable_property_name == :@provider_credentials
+        require "pry"
```
AndrewMcBurney: Could you please remove the `binding.pry`?

taquitos:
Make sure you run rubocop before commit 😄 That being said, hopefully very soon we'll have ci for ci 🙃

minuscorp:
Me facepalm

services/config_data_sources/json_project_data_source.rb
minuscorp: This forces the clone of  the project, so we can (for the moment) retrieve the Fastfile and its lanes.

taquitos: Ahhh let's make that more explicit... like make a method that specifically says `force_clone` or something

shared/models/git_repo.rb
minuscorp: The problem is that this is being triggered by a web request, there're chances that the commit does not finish after the project being created and having `user minuscorp@gmail.com doesn't have access to a project with id`.
```ruby
authenticated_controller_base.rb in user_project_with_id
raise "user #{self.user.email} doesn't have access to a project with id `#{project_id}`" if project.nil?
```

features/project/project_controller.rb
```diff
@@ -98,6 +99,9 @@ class ProjectController < AuthenticatedControllerBase
           fastfile_path: fastfile_path
       }
 
+      # Delete the project
+      FileUtils.rm_rf(repo_path)
```
taquitos: I'm scared of this. What is `repo_path` is nil, or `` or accidentally `/` ?

minuscorp: Is safe to keep the cloned repo where it is? Or how can I check if this is safe to delete?

minuscorp: I think now is better 😄

services/config_data_sources/json_project_data_source.rb
```diff
@@ -158,16 +158,22 @@ def save_git_repo_configs!(git_repo_configs: nil)
...
                                 project_name: name,
                                 lane: lane,
                                 artifact_provider: artifact_provider)
       if !self.project_exist?(new_project.project_name)
-        projects.push(new_project)
+        projects << new_project
         self.projects = projects
         logger.debug("Added project #{new_project.project_name} to projets.json in #{self.json_folder_path}")
+        logger.debug("Trigger clone project #{new_project.project_name}")
+        _ = GitRepo.new(
```
taquitos: Sorry I didn't notice this until now, but I don't think this is the right location to be creating a GitRepo, or cloning. This data source is for managing the `Project` object.

minuscorp: Where should be the best place to clone the repo associated to the project? I mean, is necessary for a project to work to clone it in create time or it can be delayed?

minuscorp: Anyway, I really think that is not even necessary to clone the git repo of the project, as It'll be done when a build is triggered...

minuscorp: I'm afraid that on project settings, as the repo may not being cloned yet, we won't have access to the repo's lanes

minuscorp:
- [ ] `project_by_id` in `ProjectService` should shallow clone the `project` repo if not cloned yet. If cloned, pull latest changes (so we can have the latest Fastfile).

features/project/project_controller.rb
```diff
@@ -98,6 +99,9 @@ class ProjectController < AuthenticatedControllerBase
           fastfile_path: fastfile_path
       }
 
+      # Delete the project
+      FileUtils.rm_rf(repo_path) if File.directory?(File.join(repo_path, ".git"))
```
taquitos: Phew, this is better :)

services/config_data_sources/json_project_data_source.rb
```diff
@@ -158,14 +158,14 @@ def save_git_repo_configs!(git_repo_configs: nil)
     end
 
     def create_project!(name: nil, repo_config: nil, enabled: nil, lane: nil, artifact_provider: nil)
-      projects = self.projects
+      projects = self.projects.clone
```
taquitos: I think this will break, `self.projects` is an array, right?

services/config_data_sources/json_project_data_source.rb
```diff
@@ -158,14 +158,14 @@ def save_git_repo_configs!(git_repo_configs: nil)
...
       new_project = Project.new(repo_config: repo_config,
                                 enabled: enabled,
                                 project_name: name,
                                 lane: lane,
                                 artifact_provider: artifact_provider)
       if !self.project_exist?(new_project.project_name)
-        projects.push(new_project)
+        projects << new_project
```
taquitos: I like this syntax better 😄

services/configuration_repository_service.rb
```diff
@@ -118,7 +121,6 @@ def create_remote_json_file(file_path, json_string: "[]")
     # @return [Boolean]
     def remote_file_a_json_array?(file_path)
       return false unless configuration_repository_exists?
```
taquitos: You don't have to fix anything here, but I like having an empty line after precondition checks right at the start of methods. I think of these things as `guard` statements. They should be separated, from the actual business logic in my mind.

services/config_data_sources/json_project_data_source.rb
minuscorp: As far as I've tested it, doesn't seem to break, `.dup` just makes a light copy, but `.clone` makes a deep copy of the whole object (and its contents). That being said, just as far as I've tested so far. If we are not sure this can break at some moment let's just revert it 😄

services/configuration_repository_service.rb
minuscorp: I think @AndrewMcBurney is on this already, cc him to let him take note of it 😄

services/data_sources/json_user_data_source.rb
```diff
@@ -172,24 +172,23 @@ def update_user!(user: nil)
...
       if !self.user_exist?(email: email)
         users.push(new_user)
         self.users = users
         logger.debug("Added user #{new_user.email}, writing out users.json to #{user_file_path}")
-        return new_user
       else
         logger.debug("Couldn't add user #{new_user.email} because they already exist")
-        return nil
       end
+      return new_user
```
taquitos: So in this case we always return a user, even if we didn't create one. Let's change it back to only returning the `new_user` if we actually created a user, otherwise return `nil` to signal that creation didn't happen.

services/project_service.rb
```diff
@@ -35,6 +37,8 @@ def create_project!(name: nil, repo_config: nil, enabled: nil, lane: nil, artifa
       project = self.project_data_source.create_project!(name: name, repo_config: repo_config, enabled: enabled, lane: lane, artifact_provider: artifact_provider)
       raise "Project couldn't be created" if project.nil?
       self.commit_repo_changes!(message: "Created project #{project.project_name}.")
+      # We shallow clone the repo to have the information needed for retrieving lanes.
+      _ = self.git_repo_for_project(project: project)
```
taquitos: let's put this into a method called `shallow_clone_project(project: nil)`
I want to make it even more obvious with code (not just comments) that we're relying on side effects here.

services/data_sources/json_user_data_source.rb
minuscorp: You're totally right

services/config_data_sources/json_project_data_source.rb
taquitos: Ohhh, no no, I was confused. I was thinking you wanted to call `.clone` on a git repo, not the actual Ruby `Object.clone`. That being said, it looks like this is just a shallow clone too 😆 

```
Produces a shallow copy of obj—the instance variables of obj are copied, but not the objects they reference. clone copies the frozen 
```
(https://ruby-doc.org/core-2.3.3/Object.html#method-i-clone)

shared/models/git_repo.rb
```diff
@@ -370,7 +385,8 @@ def clone_synchronously(repo_auth: self.repo_auth)
       logger.debug("[#{self.git_config.id}]: Cloning git repo #{self.git_config.git_url}")
       Git.clone(self.git_config.git_url, self.git_config.id,
                 path: self.git_config.containing_path,
-                recursive: true)
+                recursive: true,
+                depth: 1)
```
taquitos: This makes it so remote branches aren't fetched 🙃

minuscorp: Yep, was my bad introducing this in this patch where there's no easy way to get remotes and operate with shallow cloning. My on progress PR #283 will fix this. In the meantime, if this is something is causing you issues, you can revert it. And sorry for the inconvenience

PR #255, opened 2018-03-09
Add section about not having to use the terminal

KrauseFx:
This isn't not gonna be the case for the first release, but should be a core principal we aim for when building fastlane.ci, this includes visual representation of a fastlane config file, fastlane onboarding via fastlane.ci etc.


taquitos:
Maybe we should hold off on merging this since we don't actually plan on doing anything for MVP? I dunno, either way, I'm ok 😄

KrauseFx:
Yah I was thinking about this also, not sure tbh. I'd b ok with merging and then re-factoring again in 1-3 months or so

Issue #238, opened 2018-03-08
Why do we not enqueue builds for new commits if we already have builds in the queue


KrauseFx:
Might be a stupid question, but maybe worth adding some comments to the code base (as I'm probably not the only one asking this in the future :))

https://github.com/fastlane/ci/blob/master/workers/check_for_new_commits_on_github_worker.rb#L39-L44
```ruby
      logger.info("Checking if we should wait for any previous previous tasks to complete")
      should_wait = self.wait_for_previous_tasks?
      if should_wait
        logger.info("We still have test runner tasks to finish, not enqueuing any more")
        return
      end

      logger.info("No old test runner tasks, enqueuing new runner tasks")
```

https://github.com/fastlane/ci/blob/master/workers/check_for_new_commits_on_github_worker.rb#L28-L34
```ruby
    def wait_for_previous_tasks?
      # see how many tasks have not completed
      not_finished_tasks = self.current_tasks.reject(&:completed)

      # if we have any tasks that are not complete, return true so that we don't enque more
      return not_finished_tasks.length > 0
    end
```

the piece I'm unsure about, why do we just not queue up the new builds when a new commit appears? Doesn't that mean that we don't create a new `Build` object, and therefore not report the `Pending` status back to GitHub? From my understanding, we should enqueue a new build as soon as we see the commit, and the TaskQueue will handle the waiting for us

KrauseFx:
via @taquitos 
> I’ll have to think about why I wrote that. You’re probably right that we should enqueue builds. Maybe we should cancel running builds that haven’t finished instead? Or we can keep those too. :man-tipping-hand:

KrauseFx:
From what I understand, even if we have those in the queue, and the server gets killed, we have those `pending` builds still there, and on server start up we should trigger new builds again. I think we don't re-run the builds, but we'd trigger new ones or so? This will probably need some more work (see https://github.com/fastlane/ci/issues/233)

PR #267, opened 2018-03-09
For #238, just enqueue it

taquitos:
Things have changed since I added this logic, it's time to remove it. Status changes happen right before enqueue, so if we skip this part, we might not set status to pending at the right time.


taquitos:
#238

Issue #280, opened 2018-03-11
Split responsibilities between ci-config repo operations and project's repo operations.


minuscorp:
For now, we use a unified interface to operate over any repo using `GitRepo`. Which is not the best way to proceed, because we have two separate kind of scopes concerning repo operations, the `ci-config` and every `project's repo`.
i.e., `ci-config` repo should handle conflicts and synchronization in the best manner possible, but `project's repo` does we should make shallow clones, for certain branches, never ever commit changes into it, etc. So I think we should have extended `ProjectService` to operate with project's repos and a `ci-config` repo service which handles their own operations separately.

minuscorp:
I'll try to elaborate a little bit further what I'm trying to tackle in this idea. ATM, we have a single class `GitRepo` which is responsible of doing everything related to Git operations, concretely with Github for now. Also we've got `GithubService` that, through a set of credentials make possible to operate or obtain information given a repo name.

The main goal of the PR from this issue is to achieve the following:

1. Every `User` has a number of `ProviderCredentials` of a certain `type`, as mean to communicate with the APIs of different code hosting services (Github, Bitbucket, Gitlab, etc).

2. For a given `User` and a given `ProviderCredential` a  `XXXCodeHostingService` must exist, normally under demand. None of this `CodeHostingServices` are meant to be persisted in any way.

3. A `CodeHostingService` provides a unified interface, independently of the `ProviderCredential` type.

4. Its responsibilities are:
    - Retrieve repositories accessible given the `ProviderCredential`.
    - Retrieve repositories **branches**.
    - Shallow clone repositories given a `Project` (the reachability of the `Project`'s repo will depend of the `ProviderCredential` which created the `CodeHostingService`.
    - The clone of the `Project`'s repo is done dependant of the `JobTrigger`'s branch attached to the `Project`.
    - Peek repository's (temporary shallow clone them) structure, in order to find the all branches and their `Fastfile`s **before** the project being created (given the full repo name) or with the created `Project`, in order to retrieve information to change its settings (branch, lane, etc.).
    - Shallow clone a `Project`'s repo with a certain branch every time a `Build` is being triggered.

5. A memoized `CIConfigurationService`, which responsibilities are:
    - Keep the `ci-config` repo up-to-date. Commit local changes, pushing them to remote, pulling possible remote changes, handle conflicts.
    - Keep the correct structure of all critical files.
[...]

PR #283, opened 2018-03-12
Rebuilding GithubService and GitRepo

minuscorp:
Follow up from https://github.com/fastlane/ci/issues/280#issuecomment-372314303

I'll try to elaborate a little bit further what I'm trying to tackle in this idea. ATM, we have a single class GitRepo which is responsible of doing everything related to Git operations, concretely with Github for now. Also we've got GithubService that, through a set of credentials make possible to operate or obtain information given a repo name.

The main goal of the PR from this issue is to achieve the following:

Every `User` has a number of `ProviderCredential`s of a certain type, as mean to communicate with the APIs of different code hosting services (Github, Bitbucket, Gitlab, etc).

For a given `User` and a given `ProviderCredential` a `CodeHostingService` must exist, normally under demand. None of this `CodeHostingService`s are meant to be persisted in any way.

- A `CodeHostingService` provides a unified interface, independently of the `ProviderCredential` type.

- [x] Retrieve repositories accessible given the `ProviderCredential`.
- [x] Retrieve repositories branches.
- [x] Shallow clone repositories given a `Project` (the reachability of the `Project`'s repo will depend of the `ProviderCredential` which created the `CodeHostingService`.
- [x] The clone of the `Project`'s repo is done dependant of the `JobTrigger`'s branch attached to the `Project`.
- [x] Peek repository's (temporary shallow clone them) structure, in order to find the all branches and their Fastfiles before the project being created (given the full repo name) or with the created `Project`, in order to retrieve information to change its settings (branch, lane, etc.).
- [x] Add a memoized cache to all repos being peeked at a certain moment in order to avoid several clones of the same repo, this cache is optional and can be avoided at any time on demand. 
- [x] Shallow clone a `Project`'s repo with a certain branch every time a `Build` is being triggered.

- A memoized `CIConfigurationService`, which responsibilities are:

- [x] Keep the ci-config repo up-to-date. Commit local changes, pushing them to remote, pulling possible remote changes, handle conflicts.
- [x] **Always** keep the correct structure of all critical files (`projects.json`, `users.json`).
> ^ This is being done by launch process so I have migrated the responsibility to my new code base. 
- ~`DataSources` may not be responsible of notifying changes to the `CIConfigurationService`, through [Watchman](https://github.com/facebook/watchman/tree/master/ruby/ruby-watchman), the service should be automatically notified of any change made in the directory and consequently make the required operations for keeping a nice and clean historical record of all operations made over the `ci-config` repo.~

## Achievements

- **Deprecate** `GitRepo` in favor of `CodeHostingService` which acts as an interface between
 `local_repo <-> project <-> CodeHostingService <-> CodeHosting`.

- Provide a `ci-config` decorator utilities for:
    - Pull before editing a file of the repo.
    - **[BETA]** Commit and push changes after changes being made to files of the repo.

- `ConfigurationRepositoryService` as an interface to interact with the `ci-config` repo independently of the Code Hosting Service behind it.

- `GitHubConfigurationRepositoryService` as a fully functional example on how (depending on the user's `ProviderCredential`) interact with the `ci-config` repo hosted in GitHub.

- `GitHubService` as a fully functional example on how (depending on the user's `ProviderCredential`) interact with any `Project` whose `RepoConfig` relies on GitHub.

- Always checkout the selected `branch` for a given `job_trigger` and always reset hard to a given `sha` when looking for new commits.

- Never enqueue several builds from the same project for a given `build_number` or `sha`.

- Cloning operations are synchronized, as it is required to set and unset authentication in a thread safe manner to avoid file locking fatal errors.

- **Everything** that worked by now in the application now should work with a cleaner API where the developer has the control over what operations are being made at what time.

### TODO In follow up:

- Cleanup repositories for certain `sha` that are already built to avoid disk overhead.


minuscorp:
Gotta want lot of feedback on this PR as this involves a lot of core changes and don't want to mess up everything (very early progress tho)

services/code_hosting/git_hub_service.rb
```diff
@@ -29,6 +37,17 @@ def initialize(provider_credential: nil)
...
 
+    # This method shallow clones the repo using the provider credential and looks for a valid _fastlane_
+    # configuration, parses the Fastfile (if found), and returns the lanes in the following way:
+    #     { 
+    #       :ios => [:lane_name, :other_lane_name],
+    #       :android => [:just_other_lane],
+    #       :no_platform => [:a_not_platform_lane]
+    #     }
+    # @param [String] repo_url
+    def self.peek_fastfile_configuration(repo_url: nil, provider_credential: nil)
```
taquitos: ❤️ this is gonna be so nice to have

services/code_hosting/git_hub_service.rb
```diff
@@ -12,6 +14,12 @@ class GitHubService < CodeHostingService
 
     class << self
       attr_accessor :status_context_prefix
+
+      def self.temp_path
```
taquitos: I wonder if it makes sense to have this live as some constant in the fastlane.ci app. I think we need access to the temp directory in a few places, so it might be a good thing to extract.

minuscorp: The temp path is live as long as any `GithubService` access to the class variable (as it creates the directory in case it doesn't exist). It comes in my mind the idea of implementing some kind of _global_ (between `GithubService` instances) state where some kind of information might be cached. So maybe we don't want to be cloning the same repo over and over again, and its less expensive to have some kind of metadata along with the cloned repo as long as the path exists (i.e., date of clone, cache of `Fastfile.json`).

services/code_hosting/git_hub_service.rb
KrauseFx: Great idea 👍

services/code_hosting/git_hub_service.rb
```diff
@@ -12,6 +17,148 @@ class GitHubService < CodeHostingService
...
+      #       :ios => [:lane_name, :other_lane_name],
+      #       :android => [:just_other_lane],
+      #       :no_platform => [:a_not_platform_lane]
+      #     }
+      # @param repo_url [String]
+      # @param branch [String]
+      # @param provider_credential [GithubProviderCredential]
+      def peek_fastfile_configuration(repo_url: nil, branch: "master", provider_credential: nil, cache: true)
+        repo = repo_from_url(repo_url)
+        return self.cache[repo + branch] if cache && self.cache && !self.cache[repo + branch].nil? && self.cache[repo + branch].kind_of?(Hash)
```
minuscorp: So I thought this could be a nice idea, an in-memory `repo` + `branch` cache, this is alive as long as the application exists, but I'm storing all `fastlane.json` from every peeked repo, so it can even be loaded between executions. The `cache` flag lets the user reload the whole thing just in case he made a recent change in the Fastfile.

features/global/layout.erb
```diff
@@ -4,6 +4,7 @@
     <title>fastlane CI</title>
     <link rel="stylesheet" href="/material.min.css">
     <script src="/material.min.js"></script>
+    <script src="/jquery-3.3.1.min.js"></script>
```
minuscorp: I've added `jQuery` for ease `DOM` manipulation, but this is not a change that affects the application architecture so I thought it was safe to add for development purposes.

KrauseFx: Sorry to be that person, but I'm pretty sure we want to avoid jQuery :)

minuscorp: Do you have another suggestion during early development stages besides raw JS for `DOM` manipulation? I'm totally open for other options, or just Vanilla JS if no other option is available

taquitos: I think because the only features we need are to query a json-serving endpoint and append items to the DOM, keeping it to vanilla js is the best way to move forward.  Eg: 

```js
<script type="text/javascript">
function loadXMLDoc() {
    var xmlhttp = new XMLHttpRequest();

    xmlhttp.onreadystatechange = function() {
        if (xmlhttp.readyState == XMLHttpRequest.DONE) {   // XMLHttpRequest.DONE == 4
           if (xmlhttp.status == 200) {
               document.getElementById("myDiv").innerHTML = xmlhttp.responseText;
           }
           else if (xmlhttp.status == 400) {
              alert('There was an error 400');
           }
           else {
               alert('something else other than 200 was returned');
           }
        }
    };

    xmlhttp.open("GET", "ajax_info.txt", true);
    xmlhttp.send();
}
</script>
```
(taken from https://stackoverflow.com/a/8567149)

minuscorp: Alright, gonna stick with vanilla for now, was using jQuery to avoid wasting time on things that doesn't really matter, as I have tested that the service itself works (I've really never worked with vanilla before😅)

taquitos: Awesome, thanks @minuscorp, I know it's a pain, but keeping as many external libraries out as possible helps us reduce the chances of running into support issues in the future. Also, you get to learn something 😄 (even if you might not ever use it again 🙃)

services/code_hosting/git_hub_service.rb
```diff
@@ -1,118 +1,352 @@
...
+        return fastfile_path
+      end
+
+      # Class method that setups the authentication needed for making git operations.
+      # @param [String] repo_url
+      # @param [GithubProviderCredential] provider_credential
+      # @param [String] path
+      def setup_auth(repo_url: nil, provider_credential: nil, path: nil)
+        repo = repo_from_url(repo_url)
+        self.temporary_storage_path = File.join(self.temporary_git_storage, "git-auth-#{SecureRandom.uuid}")
```
minuscorp: BTW, I've seen this https://github.com/fastlane/ci/pull/290 @taquitos, should I implement this as you've added in that PR? What that tries to achieve? to remove the duplication of the temporary_auths being created for each clone?

services/code_hosting/git_hub_service.rb
```diff
@@ -1,118 +1,352 @@
...
+        end
+        use_credentials_command = "git config --#{scope} credential.helper 'store --file #{self.temporary_storage_path.shellescape}' #{File.join(path, repo.split('/').last)}"
+
+        cmd = TTY::Command.new(printer: :quiet)
+        cmd.run(store_credentials_command, input: content)
+        cmd.run(use_credentials_command)
+      end
+
+      # Class method that removes the authentication stored for git operations.
+      def unset_auth
```
minuscorp: Maybe here I should to inject the random UUID generated in order to cleanup the directory between executions

services/code_hosting/git_hub_service.rb
```diff
@@ -1,118 +1,352 @@
...
+        path = File.join(path, repo, branch)
+        FileUtils.rm_rf(path) if File.directory?(path)
+        FileUtils.mkdir_p(path)
+        self.setup_auth(repo_url: repo_url, provider_credential: provider_credential, path: path)
+        Git.clone(url_from_repo(repo_url), repo.split("/").last,
+                  path: path,
+                  recursive: true,
+                  depth: 1)
+        git = Git.open(File.join(path, repo.split("/").last))
+        git.branch(branch).checkout
```
minuscorp: Also, @taquitos, correct me if I'm wrong, I'm creating a `/org/repo/branch/` under `File.join(Dir.tmpdir, ".fastlane")` dir, whenever a checkout is being made, the repo is fetched clean from origin, doesn't it? (So multiple peeks to the same repo and branch doesn't duplicate folders)

services/code_hosting/git_hub_service.rb
```diff
@@ -1,118 +1,352 @@
...
+
+      attr_writer :cache
+
+      def cache
+        @cache ||= {}
+        return @cache
+      end
+
+      def temp_path
+        temporary_path = File.join(Dir.tmpdir, ".fastlane")
```
minuscorp: I was also wondering how to manage this directory disk space, how does the system manages the size of this folders. AFAIK they're not deleted by default, but checking from dir size and cleanup everything sounds like a little risky for me.

features/global/layout.erb
minuscorp: I couldn't remember why I hated vanilla JS until now 🤣

services/code_hosting/git_hub_service.rb
taquitos: Yes, you should definitely take that change I made. We don't need tons of unique temporary auths being created. We just need 1 temp per checkout, so that's what my update did 😄

services/code_hosting/git_hub_service.rb
taquitos: yes, I think that's the best approach. Clean checkouts are the best 💯

services/code_hosting/git_hub_service.rb
taquitos: I don't think we need to solve that right now, it might add unneeded complexity. I think we can make an issue to address it we need to cross that bridge 😄

features/project/project_controller.rb
```diff
@@ -126,17 +105,24 @@ class ProjectController < AuthenticatedControllerBase
...
+        # TODO: Until we make a proper interface to attach JobTriggers to a Project, let's add a manual one for the selected branch.
+        job_triggers: [FastlaneCI::ManualJobTrigger.new(branch: branch)]
       )
 
       if !project.nil?
+        github_service = FastlaneCI::GitHubService.new(
+          provider_credential: provider_credential,
+          project: project
+        )
+        require "pry"
```
taquitos: 🙃

services/code_hosting/git_hub_service.rb
minuscorp: Gonna steal that for the next commit 😄

services/code_hosting/git_hub_service.rb
```diff
@@ -1,118 +1,352 @@
...
+        path = File.join(path, repo, branch)
+        begin
+          git_path = File.join(path, repo.split("/").last)
+          # This triggers the check of an existing repo in the given path,
+          # we recover from the error making the clone and checkout
+          Git.open(git_path)
+          fastfile_path = self.fastfile_path(root_path: git_path)
+          fastfile = Fastlane::FastfileParser.new(path: fastfile_path)
+          fastfile_config = {}
+          fastfile.tree.each_key do |key|
```
taquitos: Gonna want a lot more documentation around each of these steps 😄

services/code_hosting/git_hub_service.rb
```diff
@@ -1,118 +1,352 @@
...
+            else
+              fastfile_config[key.to_sym] = fastfile.tree[key]
+            end
+          end
+          fastfile_json_path = File.join(path, "fastfile.json")
+          FileUtils.touch(fastfile_json_path) unless File.exist?(fastfile_json_path) && !File.zero?(fastfile_json_path)
+          File.write(fastfile_json_path, JSON.pretty_generate(fastfile_config))
+          self.cache[repo + branch] = fastfile_config
+          return fastfile_config
+        rescue ArgumentError
```
taquitos: What would cause this?

features/project/project_controller.rb
minuscorp: 🤦‍♂️

services/code_hosting/git_hub_service.rb
taquitos: I don't know that is needs to be random. It could just be a unique hash of org, repo name, and committer

services/code_hosting/git_hub_service.rb
```diff
@@ -1,118 +1,352 @@
...
+      self.provider_credential = provider_credential
+      raise "Project instance not provided or wrong type parameter" if project.nil? || !project&.kind_of?(Project)
+      self.project = project
       @_client = Octokit::Client.new(access_token: provider_credential.api_token)
       Octokit.auto_paginate = true # TODO: just for now, we probably should do smart pagination in the future
       @task_queue = TaskQueue::TaskQueue.new(name: "#{provider_credential.type}-#{provider_credential.email}")
     end
 
+    def status_context
+      "fastlane.ci: " + @project.id
```
taquitos: don't forget those `return` statements ❤️

services/code_hosting/git_hub_service.rb
```diff
@@ -1,118 +1,352 @@
...
+    # @return [String] HTML URL for the given pull request query.
+    def pull_requests(branches: nil, state: "open")
+      all_open_pull_requests = client.pull_requests(self.project.repo_config.full_name, state: state)
 
       # if no specific branch, return all open prs
-      return all_open_pull_requests if branches.nil? || branches.count == 0
-
-      branch_set = branches.to_set
-      all_open_pull_requests_on_branch = all_open_pull_requests.select { |pull_request| branch_set.include?(pull_request.head.ref) }
+      return all_open_pull_requests.map(&:html_url) if branches&.count == 0
```
taquitos: Let's break this out to multiple statements, this is kinda confusing to read.

services/code_hosting/git_hub_service.rb
minuscorp: This is raised by `Git.open` when there's no valid `git` in that dir, so that allows me to trigger the whole process from a clean state. (It's a pity that they don't rise other error more exact)

services/code_hosting/git_hub_service.rb
```diff
@@ -1,118 +1,352 @@
...
+    end
+
+    # Returns the last sha for every pull request that targets a list of branches (or all if not given) and status (or open if not given)
+    # @param [Array<String>, nil] branches, Either an array of target branches names or nil.
+    # @param [String] state, Either open, closed, or all to filter by state. Default: open.
+    # @return [Array<String>] Array of the last commit SHA for the given repo and state.
+    def last_commit_sha_for_pull_requests(branches: nil, state: "open")
+      pull_requests_urls = self.pull_requests(branches: branches, state: state)
+      numbers = pull_requests_urls.map { |url| URI.parse(url) }.map(&:path).collect { |path| path.split("/").last }
+      return numbers.map { |number| commits_sha_from_pull_request(number: number) }.map(&:last)
```
taquitos: `return numbers.map { |number| commits_sha_from_pull_request(number: number) }.map(&:last)`
`return numbers.map { |number| commits_sha_from_pull_request(number: number).last }`

services/code_hosting/git_hub_service.rb
```diff
@@ -1,118 +1,352 @@
...
+      client.pull_commits(self.project.repo_config.full_name, number).map(&:sha)
+    end
+
+    # Returns the last sha for every pull request that targets a list of branches (or all if not given) and status (or open if not given)
+    # @param [Array<String>, nil] branches, Either an array of target branches names or nil.
+    # @param [String] state, Either open, closed, or all to filter by state. Default: open.
+    # @return [Array<String>] Array of the last commit SHA for the given repo and state.
+    def last_commit_sha_for_pull_requests(branches: nil, state: "open")
+      pull_requests_urls = self.pull_requests(branches: branches, state: state)
+      numbers = pull_requests_urls.map { |url| URI.parse(url) }.map(&:path).collect { |path| path.split("/").last }
```
taquitos: `numbers = pull_requests_urls.map { |url| URI.parse(url) }.map(&:path).collect { |path| path.split("/").last }`
`numbers = pull_requests_urls.map { |url| URI.parse(url).path.split("/").last }`

services/code_hosting/git_hub_service.rb
```diff
@@ -1,118 +1,352 @@
...
         end
       end
       return updated_commits
     end
 
-    # TODO: parse those here or in service layer?
-    def repos
-      client.repos({}, query: { sort: "asc" })
+    # @return [Array<String>]
+    def branches
```
taquitos: rename `branch_names`

services/code_hosting/git_hub_service.rb
```diff
@@ -140,5 +374,29 @@ def set_build_status!(repo: nil, sha: nil, state: nil, target_url: nil, descript
...
+    end
+
+    protected
+
+    def repo_from_url(repo_url)
+      return repo_url.sub("https://github.com/", "")
+    end
+
+    def url_from_repo(repo)
+      return "https://github.com/" + repo unless repo.include?("https://github.com/")
```
taquitos: I think this will return `nil` 
Maybe what you want is:
`return repo.include?("https://github.com/") ? repo : "https://github.com/" + repo`

services/config_service.rb
```diff
@@ -33,14 +33,14 @@ def active_code_hosting_service_key(provider_credential: nil)
...
-      return code_hosting_service unless code_hosting_service.nil?
+      return code_hosting_service if code_hosting_service.kind_of?(FastlaneCI::CodeHostingService)
 
       # TODO: not a big deal right now, but we should have a way of automatically generating the correct
       # CodeHostingService subclass based on the provider_credential type and maybe not have it right here.
       # A Java-style factory might be the right move here.
       case provider_credential.type
       when FastlaneCI::ProviderCredential::PROVIDER_CREDENTIAL_TYPES[:github]
-        code_hosting_service = GitHubService.new(provider_credential: provider_credential)
+        code_hosting_service = FastlaneCI::GitHubService
```
taquitos: this is weird, what's going on here? Do you mean to access the `Class` instead of an instance of the class?

minuscorp: So `GitHubService` class methods are intended to be called out of the `project`'s scope. So whenever we want to operate without a `Project` in the scope, it's all being done through its class methods. So this is done in purpose, I want to access to the `Class` itself.

minuscorp: My future review is to create a mixin that provides a unified interface for all `GitHubService`, `BitbucketService`, etc.

services/code_hosting/git_hub_service.rb
minuscorp: Yeah, I meant, as you have commited in #290, I should be able to pass that `id` into this method to cleanup the directory (as for now it doesn't know what he is actually deleting)

services/code_hosting/git_hub_service.rb
minuscorp: When you go too far with functional programming

services/code_hosting/git_hub_service.rb
minuscorp: I kinda like `branches` for simplicity, as there's no confusion with what other thing might be, do you prefer `branch_names` tho? I mean, a branch itself it's (basically) his name, so I thought it was a little verbose, but can refactor if you still want to 😄

services/code_hosting/git_hub_service.rb
taquitos: haha, I love the functional aspect of it, but I'm like 99.99% sure ruby doesn't know how to optimize what you're doing there so I'm guessing the VM ends up creating like 1,000,000 temporary objects.

services/code_hosting/git_hub_service.rb
taquitos: The reason why I like `branch_names` is because `branch` is an object with metadata, and you're mapping an array of `branch` objects to an array of `name` strings. So I think for clarity, to show that these are no longer branch objects but instead, they are the names of each branch, renaming to `branch_names` is a good idea. 

I often times think things can seem verbose, but if you've never seen the code, that verbosity can help with understanding. So I actually prefer verbosity over conciseness. It usually lends itself to more maintainable code 😄

services/config_service.rb
taquitos: Ah, ok, `code_hosting_service` implies to me that you're storing an object instance reference here, not a class reference. Maybe you can change the name to `code_hosting_service_class` and that will help clear up my confusion 😄

workers/github_worker_base.rb
```diff
@@ -58,11 +58,17 @@ def thread_id
     # This is a separate method, so it's lazy loaded
     # since it will be run on the "main" thread if it were
     # part of the #initialize method
-    def git_repo
-      @git_repo ||= GitRepo.new(
-        git_config: project.repo_config,
-        provider_credential: provider_credential
-      )
+    def git_repo_service
```
minuscorp: Gonna need some help to refactor this class, PRs into my branch are more than welcome 😄

minuscorp:
OMG, the conflicts are real, I'd need a rebase I think 🤣

workers/github_worker_base.rb
minuscorp: @taquitos

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,409 @@
...
+
+      def cache
+        @cache ||= {}
+        return @cache
+      end
+
+      # Client for GitHub-related operations, uses an in-memory cache
+      # to fast things.
+      # @param [String] api_token
+      def client(api_token)
```
taquitos: I think this might be premature optimization. I don't think the Octokit::Client is expensive to build. It might also be best recreate them anyway.

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,409 @@
...
+      #  }
+      #
+      # @param repo_url [String]
+      # @param branch [String]
+      # @param path [String]
+      # @param provider_credential [GithubProviderCredential]
+      # @return [Hash<String, Hash>] being the key the name of the platform (:ios, :android, :no_platform) and the Hash the underlying lane (name, actions).
+      def peek_fastfile_configuration(repo_url: nil, branch: nil, provider_credential: nil, path: self.temp_path, cache: true)
+        repo = repo_from_url(repo_url)
+        return self.cache[[repo, branch].join("/")] if cache && self.cache && !self.cache[[repo, branch].join("/")].nil? && self.cache[[repo, branch].join("/")].kind_of?(Hash)
```
taquitos: this is a really complicated line, let's see if there's a way we can make this more imperative so people can understand each check you're making.

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,409 @@
...
+          # we recover from the error making the clone and checkout
+          Git.open(git_path)
+          fastfile_path = self.fastfile_path(root_path: git_path)
+          fastfile = Fastlane::FastfileParser.new(path: fastfile_path)
+          fastfile_config = {}
+          # TODO: FastfileParser provides a nice and clean tree of the Fastfile given,
+          # but there's a specific case when there might be lanes or actions out of
+          # a platform scope, in those cases the key is nil. We sanitize that so the user
+          # still can select lanes without platform. But this should be done from the
+          # FastfileParser side.
```
taquitos: Suggest linking an issue here for the `FastfileParser` repo

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,409 @@
...
     include FastlaneCI::Logging
 
     class << self
-      attr_accessor :status_context_prefix
-    end
+      attr_writer :temporary_git_storage
+
+      attr_accessor :temporary_storage_path
+
+      attr_writer :cache
```
taquitos: I don't like the idea of using caches right now. I think it's too early to optimize and it adds a ton of potential for things to go wrong.

services/code_hosting/git_hub_service.rb
taquitos: yeah, I don't think we should use a cache for this right now

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,409 @@
...
+          self.cache[[repo, branch].join("/")] = fastfile, fastfile_config
+          return fastfile, fastfile_config
+        rescue ArgumentError
+          self.clone(
+            repo_url: repo_url,
+            branch: branch,
+            provider_credential: provider_credential,
+            path: path
+          )
+          fastfile, fastfile_config = self.peek_fastfile_configuration(
```
taquitos: this feels like we're just asking for infinite recursion. Maybe we should add an escape hatch somehow.

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,409 @@
...
+            repo_url: repo_url,
+            branch: branch,
+            provider_credential: provider_credential,
+            path: path,
+            cache: cache
+          )
+          return fastfile, fastfile_config
+        rescue RuntimeError
+          # This is because no Fastfile config was found, so we cannot go further.
```
taquitos: let's log this error then 🎉

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,409 @@
...
+            else
+              fastfile_config[key.to_sym] = fastfile.tree[key]
+            end
+          end
+          fastfile_json_path = File.join(path, "fastfile.json")
+          FileUtils.touch(fastfile_json_path) unless File.exist?(fastfile_json_path) && !File.zero?(fastfile_json_path)
+          File.write(fastfile_json_path, JSON.pretty_generate(fastfile_config))
+          self.cache[[repo, branch].join("/")] = fastfile, fastfile_config
+          return fastfile, fastfile_config
+        rescue ArgumentError
```
taquitos: should log this error too

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,409 @@
...
+      # @param [GithubProviderCredential] provider_credential
+      # @return [Git::Base]
+      def clone(repo_url: nil, name: nil, branch: nil, sha: nil, provider_credential: nil, path: nil)
+        repo = self.repo_from_url(repo_url)
+
+        folder_name = name || repo.split("/").last
+        clone_path = path || self.temp_path
 
-    GitHubService.status_context_prefix = "fastlane.ci: "
+        FileUtils.rm_rf(File.join(clone_path, folder_name)) if File.directory?(File.join(clone_path, folder_name))
```
taquitos: I suggest you ad intermediate temporary named variables so that we understand why `File.join(clone_path, folder_name)` is needed, so people can read the code and understand what is happening easier.

services/code_hosting/git_hub_service.rb
minuscorp: Just for clarify, this `cache` acts as a cache for the `Fastfile` configuration for a given repo, so is not for the `Octokit::Client` but for the follwing parsing that is being done (cloning, looking for the `Fastfile`, get the configuration) that is a lot of work that is done many times in the server lifecycle.

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,409 @@
...
+        return git
+      end
+
+      # Class method that finds the directory for the first Fastfile found given a root_path
+      # @param [String] root_path
+      # @return [String, nil] the path of the Fastfile or nil if no Fastfile found.
+      def fastfile_path(root_path: nil)
+        fastfiles = Dir[File.join(root_path, "fastlane/Fastfile")]
+        fastfiles = Dir[File.join(root_path, "**/fastlane/Fastfile")] if fastfiles.count == 0
+        fastfile_path = fastfiles&.first
```
taquitos: what is the `&` for here?

services/code_hosting/git_hub_service.rb
minuscorp: So the only event that triggers the `ArgumentError` is the `Git.open` method, when the git does not exists. So yeah, this method will try to clone the repo until is really being opened without issues, maybe I can just add a `retry_count` that raises whenever after `n_retries` the clone is not being made.

services/code_hosting/git_hub_service.rb
minuscorp: I didn't wanted to log the error itself because it was very confusing, as `Git` does not provide any relevant information when this happens, don't know if logging such a generic exception would be useful for the debugging. I can make a custom `.error` message log tho.

services/code_hosting/git_hub_service.rb
minuscorp: Yeah, doesn't makes sense, `fastfiles` will never be `nil` because of `Dir[]`, so `first` will be the `nil` returner, does not safe us from nothing. Will fix it.

services/code_hosting/git_hub_service.rb
minuscorp: Will open an issue on this, ideally, `FastfileParser` will generate a `to_json` nicely, so the processing below isn't needed

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,409 @@
...
-      all_open_pull_requests_on_branch = all_open_pull_requests.select { |pull_request| branch_set.include?(pull_request.head.ref) }
+      if branches&.empty?
+        return all_open_pull_requests.map(&:html_url)
+      end
 
+      pull_requests_on_branch = all_open_pull_requests.select { |pull_request| branches.include?(pull_request.base.ref) }
       # we want only the PRs whose latest commit was to one of the branches passed in
-      logger.debug("Returning all open prs from: #{repo_full_name}, branches: #{branches}, pr count: #{all_open_pull_requests.count}")
-      return all_open_pull_requests_on_branch
+      logger.debug("Returning all open prs from: #{self.project.repo_config.full_name}, branches: #{branches}, pr count: #{pull_requests_on_branch.count}")
```
taquitos: This debug comment should be `Returning all #{state} prs from...`

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,409 @@
...
-    # branches should be nil if you want all branches to be considered
-    def open_pull_requests(repo_full_name: nil, branches: nil)
-      all_open_pull_requests = client.pull_requests(repo_full_name, state: "open")
+    # Returns the urls of the pull requests for a given branch and state.
+    # If branches is not provided, all target branches are considered.
+    # @param [Array<String>, nil] branches, Either an array of target branches names or nil.
+    # @param [String] state, Either open, closed, or all to filter by state. Default: open.
+    # @return [String] HTML URL for the given pull request query.
+    def pull_requests(branches: nil, state: "open")
+      all_open_pull_requests = client.pull_requests(self.project.repo_config.full_name, state: state)
```
taquitos: variable name is wrong now, could open or closed

services/code_hosting/git_hub_service.rb
minuscorp: That's a good point

services/code_hosting/git_hub_service.rb
minuscorp: I'll split this in several statements so is more explicit for people to understand what is really doing.

services/code_hosting/git_hub_service.rb
minuscorp: True, renaming incoming.

services/configuration_repository/configuration_repository_decorator.rb
```diff
@@ -0,0 +1,31 @@
+require_relative "../services"
+
+module FastlaneCI
+  # This module acts as a decorator to be added to any method that potentially will make changes
+  # into the ci-config repository directory.
+  module ConfigurationRepositoryUpdater
+    # This decorator method is responsible of pulling changes from the ci-config repo.
+    # rubocop:disable Style/ClassVars
+    @@mutex = Mutex.new
```
taquitos: Let's not do `@@`, instead we can do:
```
    class << self
      attr_accessor :mutex
    end
    ConfigurationRepositoryUpdater.mutex = Mutex.new
```

taquitos: `mutex` for what? Let's make a better name 👍

minuscorp: Fell into the old-school Ruby syntax for clarity and less verbosity, but you're right.

services/configuration_repository/configuration_repository_decorator.rb
```diff
@@ -0,0 +1,31 @@
...
+    def self.mutex
+      return @@mutex
+    end
+    # rubocop:enable Style/ClassVars
+
+    def pull_before(func_name)
+      new_name_for_old_function = "#{func_name}_old".to_sym
+      alias_method(new_name_for_old_function, func_name)
+      define_method(func_name) do |*args|
+        if ConfigurationRepositoryUpdater.mutex.locked?
```
taquitos: Isn't this saying:
If the mutex is locked, go ahead and do that thing, otherwise, wait to acquire the mutex?

That seems like once the first thing holds the mutex, the next operation will be allowed to continue. If that's the case, that's exactly what a mutex is meant to prevent.

What is it you were trying to do here?

services/configuration_repository/configuration_repository_service.rb
```diff
@@ -0,0 +1,252 @@
...
+require "pathname"
+
+module FastlaneCI
+  # Provides operations to create and mutate the FastlaneCI configuration
+  # repository
+  module ConfigurationRepositoryService
+    include FastlaneCI::Logging
+
+    # rubocop:disable Style/ClassVars
+    @@mutex = Mutex.new
```
taquitos: same mutex comments

services/configuration_repository/configuration_repository_service.rb
```diff
@@ -0,0 +1,252 @@
...
+    # will stop redirecting the user to the `/configuration` page. In the
+    # `/dashboard` page, there's a requirement the current user has a
+    # provider credential , or it will raise an exception. By writing the
+    # provider credentials to the private configuration repo for the CI User,
+    # it prevents this issue
+    #
+    # @return [String]
+    def serialized_users
+      users = [
+        User.new(
```
taquitos: why do we need this user to be created each time? I feel like we only need that as long as we haven't written out the first users.json file with the ci_user

services/configuration_repository/configuration_repository_decorator.rb
minuscorp: I'm using the mutex as a throttle mechanism. So whenever some object has acquired the lock, the incoming pull requests will be discarded, as there's one being done ATM. This way I reduce processing overhead.

services/configuration_repository/configuration_repository_service.rb
minuscorp: This was part of @AndrewMcBurney code, I conserved it, but he might throw some insight about the purpose of it.

services/configuration_repository/configuration_repository_decorator.rb
minuscorp: I'm terrible at naming, any suggestion?

services/code_hosting/git_hub_service.rb
minuscorp: https://github.com/fastlane/fastfile-parser/issues/8

services/code_hosting/git_hub_service.rb
minuscorp: Replied below |
                        V

services/code_hosting/git_hub_service.rb
minuscorp: Justification below

shared/models/project.rb
```diff
@@ -52,29 +60,32 @@ def builds
...
         if relative
-          fastfile_path = Pathname.new(fastfile_path).relative_path_from(Pathname.new(project_path))
+          fastfile_path = Pathname.new(fastfile_path).relative_path_from(Pathname.new(local_repo_path(sha: sha)))
         end
       end
       return fastfile_path
     end
+
+    def local_repo_path(sha: nil)
+      return File.join([File.expand_path("~/.fastlane/ci/"), self.id, self.repo_config.full_name, sha, self.repo_config.name].reject { |i| i.nil? || i.empty? })
```
taquitos: let's break this down into a few steps.

services/configuration_repository/github_configuration_repository_service.rb
```diff
@@ -0,0 +1,127 @@
...
+    # Instantiates new `ConfigurationRepositoryService` class
+    #
+    # @param  [ProviderCredential] provider_credential
+    def initialize(provider_credential: nil)
+      raise "Invalid provider credential" unless provider_credential.type == FastlaneCI::ProviderCredential::PROVIDER_CREDENTIAL_TYPES[:github]
+      self.provider_credential = provider_credential
+      @code_hosting_service_class = FastlaneCI::GitHubService
+      @client = @code_hosting_service_class.client(provider_credential.api_token)
+
+      ObjectSpace.define_finalizer(self, self.class.finalizer)
```
taquitos: Let's add some comments around why we're doing this

services/configuration_repository/configuration_repository_service.rb
```diff
@@ -0,0 +1,252 @@
...
+          ConfigurationRepositoryService.mutex.synchronize do
+            handle_repo_changes(git, paths["files"])
+          end
+        end
+      end
+    end
+
+    def unwatch_changes
+      `watchman shutdown-server`
+      @socket&.close
```
taquitos: why the `&`?

services/configuration_repository/configuration_repository_service.rb
AndrewMcBurney: Yeah, we could use the `ci_user` too. I didn't think about that when I originally wrote it. 👍

shared/models/project.rb
minuscorp: Don't know if that's really possible, I mean, the main issue here is that the `sha` might nil and the other pieces might be empty strings (shouldn't, but I fall into safety first). Do you have any idea in mind?

services/configuration_repository/configuration_repository_service.rb
minuscorp: Safety purposes, `@socket` might be nil at some point, I don't want to crash because of the reference to the socket being lost.

taquitos:
Ok, so this is a huge PR, and I can't fully understand the scope of it all. So, here's what I'd like to do:

1. Check it out and see how it performs against my local setup
2. Check it out and see how it performs against the currently running fastlane.ci system
3. You should add some imperative-style tests to this
4. You should explain what testing you've done
5. Add more documentation around the logic of why you're doing what you're doing.



So this is a great refactor, I see what you're doing, but don't totally understand it whole package. I see a lot of benefit here, but I'm pretty scared because I know how CI works now, and there is a lot of stuff in this PR that isn't super clear what is going on.

services/configuration_repository/configuration_repository_decorator.rb
taquitos: Whatever the mutex is protecting + `_` ha.

minuscorp:
I refer to my first comment https://github.com/fastlane/ci/pull/283#issue-174481693 and the rationale of the event-driven architecture https://github.com/fastlane/ci/pull/311#issuecomment-373847822 maybe that'd help in understanding what I'm trying to do here. But if you think this is something we should analyze more deeply, we can go for a call or something.
About the tests, I think the PR itself is big enough and (as I commented in Slack) I'd be committed in adding tests in follow up work if this gets approved and merged. This took almost a week to design, implement and test, and I don't see that adding tests would take me less time, and right now I'm fighting over getting this PR up-to-date with the current development course (rebasing and merging) instead of focusing on improve this work.

services/code_hosting/git_hub_service.rb
minuscorp: My main concern is that there's a lot of operations that relies on the `Octokit::Client` inside this class, so memoize it for a given `api_token` seem kinda nice, and allows future code hosting services to adopt this approach easily (as it's implemented yet)

services/configuration_repository/configuration_repository_service.rb
minuscorp: @AndrewMcBurney as I don't want to mess up everything, could you please make a PR on this? 🙏

services/code_hosting/git_hub_service.rb
minuscorp: Added something more useful tho 😆

services/code_hosting/git_hub_service.rb
```diff
@@ -1,131 +1,421 @@
...
+      #
+      # @param repo_url [String]
+      # @param branch [String]
+      # @param path [String]
+      # @param provider_credential [GithubProviderCredential]
+      # @return [(Fastlane::FastfileParser, Hash<String, Hash>)] the FastlaneParser instance used for the parsing and a hash, being the key the name of the platform (:ios, :android, :no_platform) and the Hash the underlying lane (name, actions).
+      def peek_fastfile_configuration(repo_url: nil, branch: nil, provider_credential: nil, path: self.temp_path, cache: true)
+        repo = repo_from_url(repo_url)
+        # If cache is allowed for this call, and there's a Hash under the key, use the cache.
+        if cache && self.cache&.[]([repo, branch].join("/"))&.kind_of?(Hash)
```
minuscorp: I think this is safer, shorter and prettier 🎨

services/configuration_repository/configuration_repository_service.rb
taquitos: Ha, that's fine, I'm just not familiar with that Ruby language feature, any idea when it was added?

shared/models/project.rb
taquitos: I know it's not nearly as satisfying, but I think it's easier to understand at a glance like this:
```ruby
potentially_empty_path_components = [self.id, self.repo_config.full_name, sha, self.repo_config.name]
path_components = potentially_empty_path_componentsFile.reject { |i| i.nil? || i.empty? }
local_repo_path = File.join(File.expand_path("~/.fastlane/ci/"), path_components)
```

services/code_hosting/git_hub_service.rb
taquitos: A couple things: I'm not familiar enough with `Octokit::Client` to know if it has an age limit on its own authentication, or if it manages its own internal state perfectly. Often frameworks do some sort of handshake and token exchange, and if you let a client sit for too long, the current auth gets stale and then reauth is needed, which often isn't implemented correctly. Also, the longer an object lives, the more opportunities for state to change and break things. So I prefer to just create new objects until performance testing shows that we need caching.

By not using caches until its obvious that caches are needed means that you eliminate the possibility of running into many classes of bugs hard to track down problems in other people's code. Also, caching always seems deceptively easy until it's not. In my experience it's one of the largest sources of problems.

As a rule of thumb, it's best to avoid caches until you have proof that they are needed, and that they actually fix a problem.

services/code_hosting/git_hub_service.rb
taquitos: Oooh I like that

services/code_hosting/git_hub_service.rb
taquitos: 🎉

services/code_hosting/git_hub_service.rb
taquitos: It's pretty, butttt I have no idea what it's doing 🙃 
I'm sure you know this is coming:
I know it's clever, but it's really confusing and would be easier for new people (and me) to understand if you broke it into multiple lines with intermediate temporary variables describing the steps.

taquitos:
Ok, so you're right about rebasing a ton and writing tests might just cause more trouble. 

So here's my new proposed plan for merging:
1. You should do one more round of updates per comments
2. I need to test it locally and on our fastlane.ci setup
3. we have to double check with our legal folks about the facebook license (we're already doing that)

In the future, we should work together to figure out how to split this stuff into smaller PRs 👍 I'm happy to help with that ❤️

services/configuration_repository/configuration_repository_service.rb
minuscorp: 2.3 😄 http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

services/code_hosting/git_hub_service.rb
minuscorp: Gonna split it! 😄 But just in case you're not familiar with the `&.` is the safe operator chaining in Ruby (I don't know if you're familiar with Swift, but it's like the optional chaining `?.`) So long history short, calls the chain of methods until a `nil` is received. I don't know if this is a feature that I should be using systematically (I think this adds a ton of code safety).
After the explanation, do I split this? I have no problem, I adapt to your point of view 😄

minuscorp:
This was really my fault because I couldn't attend to the meeting we had to address this issue and I went by my way without any guidance.
This afternoon (before you even wake up) I'll try to get this working with the final comments of your review (prioritizing the client cache, and the cache resolving).

README.md
```diff
@@ -63,6 +63,11 @@ Requires Ruby 2.3.3 or higher. macOS and Xcode are required when building iOS pr
 bundle install
 git submodule init
 git submodule update
+brew update && brew upgrade
```
taquitos: `brew upgrade` upgrades all installed tools, we probably shouldn't suggest this since it might break something.

taquitos:
I just tried it out, and I couldn't get it to startup so I deleted my `ci` directory and re- onboarded. Same error though. Here's my pry session that shows what error is happening, I guess creating a `Git` object works fine, but as soon as `.status` is called on it, 💥 
Any thoughts on what I can do here?

screen shot 2018-03-20 at 6 33 10 pm

taquitos:
A few observations here:
One of the core tenants of fastlane.ci is that we really need to make it easy for people to install fastlane.ci and get up and running. By requiring people install `brew` and also install `watchman`, that adds a hurdle. It sounds trivial, but people are very resistant to installing other software packages as part of dependency management, and it's not a given that people will have `homebrew`.

It looks like you're using `watchman` to separate git pull/fetch/commit/push actions from reacting to file changes in the repo. That's very _reactive_ which I really like, but it make it more confusing to follow the logic if you're unfamiliar with this style, while also making the fastlane.ci install process more complicated (requiring homebrew, and watchman binary).

Architecturally speaking, I think leaving config repo management code style imperative makes it easier to understand what is going on. I think we can certainly clean up what we have. Ideally we'll be able to see clearly what we want to commit, when we want to commit it, and how we do it. This makes it easier to jump into the project and ramp up for any new contributors.

I know this is not what you wanted to hear, but we have to really prioritize making our developers' lives easier, and sometimes that means making our lives a little harder 😞

So, unfortunately, I don't think we can merge this PR in its current form. I see lots of great stuff here so if you can break it apart to submit individual PRs covering the other improvements, we should do that. Like the `artifacts` work, and the `new project` work. Also, the `pull_before(:my_custom_function)` decorator is :100: too. Such a good idea. So these are just a couple examples of things you should totally submit separate PRs for. 

I'm really sorry we didn't see this earlier. What you wrote isn't a bad idea, or bad implementation, it just doesn't fit with our vision for a seamless user experience in all phases of using fastlane.ci. Lots of great ideas in there, though.

minuscorp:
How do I proceed now with this? I mean, the watchman thing is just a method and I can clearly clean up. I was thinking on dropping it in favour of a global `ConfigurationRepositoryService.commit(file: )` and `ConfigurationRepositoryService.push` instead of doing this through Watchman. Let me know if this is something could be acceptable. I totally understand the complexity of the PR, I'm trying to get things clear to be able to merge it, if possible.

minuscorp:
I've dropped the control of the ci-config repository up to the user (except for the pulling) I do not make any commit or push into my proposal (to be added in future PR). About the issue you're seeing, it seems that `ruby-git` is giving many issues lately due to deprecations on git commands https://github.com/ruby-git/ruby-git/issues/345#issuecomment-369260500
Can I request a new review with the code as it is for now? If you consider this un-mergeable, I'll close it and try to split into pieces. The main issue is the **strong** dependency of all the code with `GitRepo`. If I deprecate it, it must be replaced with something else, so I decided to rewrite it all.
I leave the final decision to you, no pressure, I think many of the code will be used sooner or later.

minuscorp:
Done another rebase, I think @taquitos that your work in #343 is already done here, the only thing to do in a follow up would be the usage of the global `git_mutex` for all operations. I kinda tried it in several occasions, but it seemed to highly reduce the performance of the server. That being said, I do not perform any of the operations in this refactor in an asynchronous way, that should be done in a follow up PR.

minuscorp:
Here's the REPL for the `commit_after` decorator, it seems to work just fine, still doesn't make use of it on the codebase, that is for a follow up PR.

https://repl.it/repls/BadMonumentalCensorware

minuscorp:
Just final comment of my monologue, I think this is ready to merge, and everything that previously relied in `GitRepo` now works just fine with the new architecture. Better concern separation and a cleaner and clearer control by the developer of what is going on at every step they take. 🚀

minuscorp:
captura de pantalla 2018-03-22 a las 22 54 10
🎉 🎉 🎉

minuscorp:
#360 #361 #362

taquitos:
Closing in favor of breaking into smaller piece 🎈

PR #284, opened 2018-03-13
Make API clients injectable modules with memoized `client` methods.

AndrewMcBurney:
### Usage

```ruby
class ClassThatNeedsBotUserCredentialsForAPI
  include FastlaneCI::BotUserGitHubClient

  def use_client
    client.repository?("fake_repo")
  end
end
```

> CI Bot User Credentials

```ruby
class ClassThatNeedsCloneUserCredentialsForAPI
  include FastlaneCI::CloneUserGitHubClient

  def use_client
    client.repository?("fake_repo")
  end
end
```

> Clone User Credentials

### Rationale

These modules can be easily extended/modified for when we wish to extend `fastlane.ci` for other API clients (Gitlab, etc.).

Thoughts?

### DIFF
--- a/services/api_clients/api_client.rb
+++ b/services/api_clients/api_client.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module FastlaneCI
+  # Injectable ApiClient module exposing APIs for the CI user, and clone users
+  #
+  # @abstract
+  module APIClient
+    # Returns an API client object with the CI user credentials
+    #
+    # @abstract
+    def client
+      not_implemented(__method__)
+    end
+  end
+end
--- a/services/api_clients/github_clients/bot_user_github_client.rb
+++ b/services/api_clients/github_clients/bot_user_github_client.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require_relative "../api_client"
+
+module FastlaneCI
+  # An injectable module for easy access to the GitHub API
+  module BotUserGitHubClient
+    extend APIClient
+
+    # Returns an GitHub API client object with the CI user credentials
+    #
+    # @return [Octokit::Client]
+    def client
+      @client ||= Octokit::Client.new(
+        login: FastlaneCI.env.ci_user_email,
+        password: FastlaneCI.env.ci_user_password,
+        api_endpoint: "https://api.github.com/"
+      )
+    end
+  end
+end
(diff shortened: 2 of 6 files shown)


minuscorp:
I think this kinda overlaps with #283, what are you trying to achieve with this? Can we merge this in a same PR to avoid duplicating code? I'm open to help you with this! 😄

taquitos:
Ok ok, so I like this approach, but I can't help but ask: `what problem is this solving?` and `does this increase complexity more than correctness?`
While this is totally a valid and correct approach, I wonder if it will add some confusion for new comers to the project. What do you think?

AndrewMcBurney:
Here are my thoughts:

### Advantages

- It would allow us to extend for new APIs (i.e., Gitlab) relatively easily
- It abstracts away configuration details from classes that need to use the APIs
- The modules adhere to the Single Responsibility Principle, with the single responsibility of configuring an API client for use
- Lets us avoid injecting provider credentials or the `Octokit::Client` objects into the constructors of classes which need access to the GitHub API. Instead, we can use the GitHub API (and other APIs) by including the module corresponding to it

### Neutral Points:
- Should we have a need for including multiple clients in a single class (say `BotUserGitHubClient` and `CloneUserGitHubClient`), we can include both and refer to them by their module name (`BotUserGitHubClient.client`, `CloneUserGitHubClient.client`) to avoid namespace collisions
- Having an injectable module with a memoized `client` method works a little bit like a singleton instance. Usually I'm not a fan of singletons because of the [global state you're implicitly managing by mutating instance variables](http://williamdurand.fr/2013/07/30/from-stupid-to-solid-code/#singleton). However, the usage pattern for including the modules is stateless, since the `client` method you're providing is immutatable (with one caveat which might be the deal-breaker for me - _see disadvantages below_)

### Disadvantages:

- Should we mutate the environment configuration variables, we would need to reset the API clients that depend on said variables, [similar to what we currently do for the `Services` class](https://github.com/fastlane/ci/blob/master/services/services.rb#L24). This contradicts my point about the modules not having the same issues as singletons, since technically we would modify the state of the module (the credentials used for configuration)
- It does add extra complexity, and the advantages might not be worth the disadvantages aforementioned by @taquitos  and @minuscorp

I'm totally cool with closing this PR, and continuing with our current approach. My PR may be a bit of a premature abstraction. Other approaches (i.e., using classical inheritance for API clients) might be more suited for our future needs.

AndrewMcBurney:
Closing due to my previous comment's [disadvantages section](https://github.com/fastlane/ci/pull/284#issuecomment-372720809).

Issue #292, opened 2018-03-14
UserService.update_user!(user:) turns provider_credentials into Hashes


taquitos:
Instead of keeping the provider_credentials as the correct object type, this method will turn them into hashes. I suspect there is a json mapping issue somewhere.

taquitos:
🎉

PR #293, opened 2018-03-14
Workaround for update_user! bug

taquitos:
Workaround for issue #292


minuscorp:
Indeeed, this was a bug I experienced (and tried to fix) when creating a `User`, it seemed that then embedding the `ProviderCredential` inside the `User` initializer, made the `ProviderCredential` turn into a `Hash` (why is this happening?) but if you initialize the `User` and the `ProviderCredential` separately and then just append the `ProviderCredential`, it works 💥

Issue #213, opened 2018-03-05
Don't send HTML code from the server to the client, generate HTML using JS instead


KrauseFx:
Via https://github.com/fastlane/ci/pull/200
Fixes https://github.com/fastlane/ci/issues/147
More discussion at the bottom of https://github.com/fastlane/ci/issues/147
Remove the `FastlaneOutputToHtml` class

KrauseFx:
@nakhbari We're gonna close this card, don't be shocked when you see we're sending HTML code to the front-end. This is something we can easily change once we have your frontend architecture ready to go 👍

PR #295, opened 2018-03-14
Migration to have BuildRunnerOutputRow class to represent build output

KrauseFx:


KrauseFx:
Couldn't test this yet, as server doesn't start up for me

features/build/build_controller.rb
```diff
@@ -27,7 +27,7 @@ class BuildController < AuthenticatedControllerBase
         project: project,
         build: build,
         title: "Project #{project.project_name}, Build #{build.number}",
-        existing_rows: current_build_runner.all_build_output_log_lines.collect { |a| a[:html] }
+        existing_rows: current_build_runner.all_build_output_log_rows.collect(&:html)
```
taquitos: I kinda think we should standardize on `map` instead of `collect`. `map` is more widely known.

features/build_runner/build_runner_output_row.rb
```diff
@@ -0,0 +1,44 @@
...
+    attr_accessor :type
+
+    # The raw message to show
+    attr_accessor :message
+
+    # Timestamp (Time)
+    attr_accessor :time
+
+    # The plan is to remove this, as we want to render it on the front-end
+    attr_accessor :html
```
taquitos: Should we have a `TODO:` on here so that we check it before we release?

features/build_runner/fastlane_build_runner.rb
```diff
@@ -51,8 +51,14 @@ def run(completion_block: nil, &block)
           #
           #   {"type":"success","message":"Driving the lane 'ios beta'","html":"<p class=\"success\">Driving the lane 'ios beta'</p>","time"=>...}
           #
-          row[:html] = FastlaneOutputToHtml.convert_row(row)
-          block.call(row)
+          # Also we use our custom BuildRunnerOutputRow class to represent the current row
+          current_row = FastlaneCI::BuildRunnerOutputRow.new(
```
taquitos: might be good to extract this into its own method

features/build/build_controller.rb
KrauseFx: `collect` is English though

features/build_runner/build_runner_output_row.rb
KrauseFx: We have a GitHub issue for this https://github.com/fastlane/ci/issues/213 I'll update the comment to link to it. No need for a TODO if we have a GitHub issue I think, especially since this is a not a blocking thing, like if we ship the first beta with this, it's fine, no security risk, and no issue with migrating to something new

features/build_runner/build_runner_output_row.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def initialize(type: nil, message: nil, time: nil)
```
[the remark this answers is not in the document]
KrauseFx: Yep, still need to get used to that

features/build_runner/fastlane_build_runner.rb
KrauseFx: 👍

features/build/build_controller.rb
taquitos: It is english language, but in my experience, because `map` is more commonly found in *programming languages* vs `collect`, it might be better to go with since it seems more universal.

PR #298, opened 2018-03-15
Add Angular Web App

nakhbari:
This is the addition of an Angular web app framework to our solution.

This PR includes files that are automatically generated from the Angular CLI. I'd like to get this into our version history in case we need them in the future. After submitting this PR, I'll work on getting it to the barebones for what we need.

### MVP to start using Angular
- [x] Initialize an Angular Web app
- [x] Build the App programmatically at launch 
- [x] Sinatra hosts the Web App at "/"
- [x] Add method to switch between current UI and the Angular App for the time being.

### After we're able to switch between the web apps
- [ ] Remove any items here that is unnecessary for the time being https://github.com/fastlane/ci/pull/331
- [ ] Add routes (ex. /projects) to Angular router
- [ ] Migrate endpoints to serve JSON (This will later be broken down by "feature")
- [ ] Migrate view logic to Angular


package-lock.json
```diff
@@ -0,0 +1,8368 @@
...
+      "version": "0.3.2",
+      "resolved": "https://registry.npmjs.org/@schematics/package-update/-/package-update-0.3.2.tgz",
+      "integrity": "sha512-7aVP4994Hu8vRdTTohXkfGWEwLhrdNP3EZnWyBootm5zshWqlQojUGweZe5zwewsKcixeVOiy2YtW+aI4aGSLA==",
+      "requires": {
+        "rxjs": "5.5.7",
+        "semver": "5.5.0",
+        "semver-intersect": "1.3.1"
+      }
+    },
+    "abbrev": {
```
KrauseFx: Too many dependencies for GitHub to handle our scale :D

KrauseFx:
This is awesome, excited to have a front-end based application soon. I'm a little surprised by the number of configuration files we have to pull in to make things work. You know a lot more about JS than I do, I thought we could get a more basic initial version first, and then slowly grow it where necessary? Or should we proceed with this, and go from there?

taquitos:
Yeah, also excited here. One suggestion:
While you're working on this, how about using `/app/` as the root for the web app? 
That way you can merge more things quicker without removing all the current UI we have. At some point, we can then switch `/app/` over to be the root, and we'll flip the switch to use just this frontend. In the meantime, we want to be able to continue to write features and simple HTML pages to test that stuff.

I would also suggest prefixing all your routes with `/app/` creating separate controllers too for all the endpoints you'll need.

fastlane-web/src/polyfills.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Edge >= 13 on the desktop, and iOS 10 and Chrome on mobile.
+ *
+ * Learn more in https://angular.io/docs/ts/latest/guide/browser-support.html
+ */
+
+/***************************************************************************************************
+ * BROWSER POLYFILLS
+ */
+
+/** IE9, IE10 and IE11 requires all of the following polyfills. **/
```
taquitos: lol

package-lock.json
```diff
@@ -0,0 +1,8368 @@
+{
```
taquitos: Should this be checked in?

nakhbari:
@KrauseFx So most the configuration files here are for the angular CLI.

What does this CLI provide?
It has functions that ease the development of angular by 100x. 
ex. ng g component -> creates new component 
ng build -> compiles the TS and scss files, minimizes, and obfuscates the code.

I was thinking that this is a lot of dependencies too. One potential solution is that we don't require the angular CLI. Instead, we manually update the output files as we make changes to the app/

One thing i'm not a fan of here is that code search when working on the server is now including all the web code. 

In terms of moving fast with our current front-end my thought were to have the Angular app serve the http code returns from /dashboard. This will enter into our current web code and we can move around in there as we start migrating files over.

Something i am hoping we are OK putting on hold though is login. I'd like to do this at the end since it's a little more technically challenging.


@KrauseFx @taquitos WDYT?

KrauseFx:
Beautiful, thanks for clarifying on the # of files 👍

nakhbari:
So right now this is fully working with our current backend. I have it so that the "Home path": "/" redirects to /dashboard which is going to start returning the serverside HTML.

I'm going to look into whittling down these extra files for now. I don't know how committed we are to having tests for our component 😛

package-lock.json
nakhbari: yes, this is going to ensure everyone gets exactly the same dependency tree as me.

AndrewMcBurney:
screen shot 2018-03-20 at 6 51 48 pm
LGTM!! :shipit: 😂 💯

taquitos:
🍬:goat:

PR #299, opened 2018-03-15
Add monkey patch to fix ruby-git bug

nakhbari:
Add monkey patch to fix ruby-git bug until https://github.com/ruby-git/ruby-git/pull/320 is resolved

Not sure whether or not we should do this, but this is what allows me to spin up the CI server.

ruby mixins are pretty cool 😄


shared/git_monkey_patches.rb
```diff
@@ -0,0 +1,16 @@
...
+    def ls_files(location=nil)
+      location ||= '.'
+      hsh = {}
+      command_lines('ls-files', ['--stage', location]).each do |line|
+        (info, file) = line.split("\t")
+        (mode, sha, stage) = info.split
+        file = eval(file) if file =~ /^\".*\"$/ # This takes care of quoted strings returned from git
+        hsh[file] = {:path => file, :mode_index => mode, :sha_index => sha, :stage => stage}
+      end
+      hsh
```
minuscorp: explicit return statement 😄

taquitos: Also let's comment where the code came from 👍

shared/models/git_repo.rb
```diff
@@ -162,7 +163,6 @@ def setup_repo
       else
         logger.debug("Cloning #{self.git_config.git_url} into #{self.git_config.local_repo_path}")
         self.clone
-
```
taquitos: 👾

PR #300, opened 2018-03-15
bug fix: git repo config has incorrect local repo path

nakhbari:
The JSON converter would start by creating a param-less version of the model. The constructor would use a random UUID for the `repo_config ID`. This would be used for the `local_repo_path`. Later as the JSON converter begins to fill in the model's properties the ID is changed to it's proper value, but the `local_repo_path` remains the same.

Git Clone would clone to the containing folder (`~/.fastlane/ci`) with the child folder of `git_config.id`. And we use the `git_config.local_repo_path` to reference this repo but it's in a different castle. 

Had to debug quite a bit to find this issue 😩 BUT NOW IT'S FIXED! 😸


taquitos:
🎉 wow great find/fix 🐐:goat::goat::goat:!

Issue #303, opened 2018-03-15
Enable fastlane build runner to use a specific task queue


taquitos:
fastlane build runner does a lot of work, and it might do `Dir.chdir`. To prevent collisions and warnings, we need to share a single queue to do all the operations that might cause a `Dir.chdir` so we don't break things.

Also related: https://github.com/fastlane/ci/issues/215
Should be fixed when this issue is fixed

taquitos:
Fixed!

Issue #302, opened 2018-03-15
Only care about pending commits if the SHA hasn't been rebuilt in lately


taquitos:
When CI starts up, we scan for pending statuses. If we find one, we submit a new build. We don't modify the old status, which means every startup we'll detect a pending status and kick off a build. We need to only care about pending if this commit sha hasn't be rebuilt with a status other than `pending`

taquitos:
Fixed!

Issue #215, opened 2018-03-05
Investigate git chdir warnings when connecting fastlane.ci with fastlane


KrauseFx:
Introduced with https://github.com/fastlane/ci/pull/200

```
lib/git/lib.rb:460: warning: conflicting chdir during another chdir block
```

_fastlane_ makes use of `Dir.chdir` as part of executing the `Fastfile`. This is necessary as we want the `Fastfile` to run in its directory and not in the project root directory. Unfortunately, this seems to cause weird conflicts with the git library we use. From what I see, it's not possible to stop using `Dir.chdir` as part of _fastlane_, as we do want to run in a different directory. However, I don't yet understand why this causes a conflict, as we don't run `fastlane` as part of a `git` command. This will require some further investigation.

taquitos:
Fixed!

KrauseFx:
yesss 🚀

PR #304, opened 2018-03-15
fastlane build runner to use same queue as git

taquitos:
Address #303, #302, and #215

### DIFF
--- a/services/build_service.rb
+++ b/services/build_service.rb
@@ -15,11 +15,39 @@ module FastlaneCI
     end
 
     def list_builds(project: nil)
-      self.build_data_source.list_builds(project: project)
+      return self.build_data_source.list_builds(project: project)
     end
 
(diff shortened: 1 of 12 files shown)


services/build_service.rb
```diff
@@ -15,11 +15,39 @@ def initialize(build_data_source: nil)
...
     end
 
     def pending_builds(project: nil)
-      self.build_data_source.pending_builds(project: project)
+      return self.build_data_source.pending_builds(project: project)
+    end
+
+    # returns a list of commit shas where the status is pending
+    # and it hasn't been superceeded by a newer build
+    def pending_build_shas_needing_rebuilds(project:)
```
taquitos: @minuscorp is this functional enough? I can't tell yet 😆

minuscorp: **MORE** function chaining

PR #307, opened 2018-03-16
Fix code style to make rubocop happy for git monkey patch

KrauseFx:


shared/git_monkey_patches.rb
```diff
@@ -14,16 +14,20 @@ module Git
...
-      location ||= '.'
+    def ls_files(location = nil)
+      location ||= "."
       hsh = {}
-      command_lines('ls-files', ['--stage', location]).each do |line|
+      command_lines("ls-files", ["--stage", location]).each do |line|
         (info, file) = line.split("\t")
         (mode, sha, stage) = info.split
+
+        # rubocop:disable Security/Eval
```
taquitos: 🙈🙉🙊

KrauseFx: I didn't check if the actual `git` implementation has this? If if does, it should be fine I think?

taquitos: I didn't check either, but it's *probably* fine

PR #309, opened 2018-03-16
Add useful exception if build information JSON file can't be parsed

KrauseFx:


services/data_sources/json_build_data_source.rb
```diff
@@ -94,8 +94,13 @@ def list_builds(project: nil)
       end
 
       most_recent_builds = build_files.map do |build_path|
-        build_object_hash = JSON.parse(File.read(build_path))
-        build = Build.from_json!(build_object_hash)
+        begin
+          build_object_hash = JSON.parse(File.read(build_path))
+          build = Build.from_json!(build_object_hash)
+        rescue => ex
+          logger.debug(ex.to_s)
```
taquitos: don't need `.to_s` our logger does stuff with it. If it doesn't do what you think it should, let's update the logger to handle printing exceptions 😺

KrauseFx: I know it doesn't, I assume though that we do want to be consistent when it comes to data types for parameters as much as possible.

services/data_sources/json_build_data_source.rb
```diff
@@ -94,8 +94,13 @@ def list_builds(project: nil)
...
+          raise "Error parsing build information on path '#{File.expand_path(build_path)}'"
```
snatchev: It would be nicer to re-raise the same exception type that was raised originally. Or at least use a appropriate exception type here instead of `RuntimeError`

minuscorp: I think I should wrap the `from_json!` and `to_object_dictionary!` in a safe block, catch all exceptions, and raise `JSONConvertibleException` with the original underlying exception, so the exception is grouped by the tool that raised the exception in first place.

KrauseFx:
I'll merge this for now, feel free to submit a follow-up PR 👍 Thanks

Issue #310, opened 2018-03-16
Define source of truth of fastlane.ci


KrauseFx:
Where is the source of truth?

1. `ci-config` git remote
1. The locally cloned `ci-config` repo
1. The GitHub PR build status (only applies to build status)

Meaning, that if a user changes e.g. a project setting, we update the JSON file locally, commit the change, and try to push to the git remote. If the push fails, a user has manually updated the configuration file, meaning it might be enough to `git pull` and `git push` again. However if the `git pull` causes merge conflicts, we'll have to show an error message to the user, telling that to resolve the merge conflict, or we could just discard the local changes, and take whatever the remote tells us. In the future, we could provide a UI for this to ask the user what to do.

We now define a few assumptions for fastlane.ci for the `ci-config` repo:
- The `ci-config` repo is only updated by a single node, being the master `fastlane.ci` machine. Even if the developer has a fleet of 100 Macs, they all communicate with the master, this ain't a decentralized blockchain
- The only reason why the `ci-config` repo might be changed outside of the `fastlane.ci` process on the Mac, is when the user manually edits a file, which is a valid use-case. By storing all configurations in version control, giving the user the power of those files is one of the reasons why we decided to use git and JSON.
- It's enough to `git pull` every few minutes on the `fastlane.ci` main machine on the background worker (something we already do AFAIK), `git pull` takes less than a second, so it's fine to do that often in the background. Users won't expect a JSON file change to **instantly** show up on their local server, they'd either try to restart the server (which would work as we pull on startup) or wait for a few minutes

taquitos:
Yeah, `ci-config` should be totally automated interactions, like your `match` repo. If there is a conflict, we can provide UI that helps the developer to understand what their options are for resolving it.

KrauseFx:
Moved over to https://github.com/fastlane/ci/pull/311

PR #311, opened 2018-03-16
Add information about source of truth

KrauseFx:
Moved over from https://github.com/fastlane/ci/issues/310

### DIFF
--- a/docs/CiConfigRepo.md
+++ b/docs/CiConfigRepo.md
@@ -1,0 +1,14 @@
+# `ci-config` repo
+
+## Where is the source of truth when it comes to configuration?
+
+1. `ci-config` git remote
+1. The locally cloned `ci-config` repo
+1. The GitHub PR build status (only applies to build status)
+
+Meaning, that if a user changes e.g. a project setting, we update the JSON file locally, commit the change, and try to push to the git remote. If the push fails, a user has manually updated the configuration file, meaning it might be enough to `git pull` and `git push` again. However if the `git pull` causes merge conflicts, we'll have to show an error message to the user, telling that to resolve the merge conflict, or we could just discard the local changes, and take whatever the remote tells us. In the future, we could provide a UI for this to ask the user what to do.
+
+## We now define a few assumptions for fastlane.ci for the `ci-config` repo:
+- The `ci-config` repo is only updated by a single node, being the master `fastlane.ci` machine. Even if the developer has a fleet of 100 Macs, they all communicate with the master, this ain't a decentralized blockchain
+- The only reason why the `ci-config` repo might be changed outside of the `fastlane.ci` process on the Mac, is when the user manually edits a file, which is a valid use-case. By storing all configurations in version control, giving the user the power of those files is one of the reasons why we decided to use git and JSON.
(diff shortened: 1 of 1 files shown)


minuscorp:
I'll try to elaborate a little bit why I think that polling the repo for pulls is not the desired behavior to implement in the CI:
### Cons:
- First of all, pulling might be done when is not necessary at all, not that it is a computational cost problem, but unnecessary bloat.
- Second, race conditions. That might be rare cases when we're pushing changes when we think that we're up-to-date with remote (because a pull has been done a few minutes earlier) but we end up with conflicts. Those conflicts will be hard to resolve from a programmatically point of view, and I don't see asking the user to use the CLI or the web interface to resolve conflicts (I think we're trying to build a plug&play CI).

### Proposed solution:
- Use the `Python`'s inherited decorator pattern. Just we have to hook the methods (they're just a few ones really) that makes changes to the `ci-config` repo **before** they're done. This can be achieved in a programmatically way, for example:
```ruby
module Documenter
  def document(func_name)   # This is the function that will DO the decoration: given a function, it'll extend it to have 'documentation' functionality.
    new_name_for_old_function = "#{func_name}_old".to_sym   # We extend the old function by 'replacing' it - but to do that, we need to preserve the old one so we can still call it from the snazzy new function.
    alias_method(new_name_for_old_function, func_name)  # This function, alias_method(), does what it says on the tin - allows us to call either function name to do the same thing.  So now we have TWO references to the OLD crappy function.  Note that alias_method is NOT a built-in function, but is a method of Class - that's one reason we're doing this from a module.
    define_method(func_name) do |*args|   # Here we're writing a new method with the name func_name.  Yes, that means we're REPLACING the old method.
      puts "about to call #{func_name}(#{args.join(', ')})"  # ... do whatever extended functionality you want here ...
      send(new_name_for_old_function, *args)  # This is the same as `self.send`.  `self` here is an instance of your extended class.  As we had TWO references to the original method, we still have one left over, so we can call it here.
      end
    end
  end

class Squarer   # Drop any idea of doing things outside of classes.  Your method to decorate has to be in a class/instance rather than floating globally, because the afore-used functions alias_method and define_method are not global.
  extend Documenter   # We have to give our class the ability to document its functions.  Note we EXTEND, not INCLUDE - this gives Squarer, which is an INSTANCE of Class, the class method document() - we would use `include` if we wanted to give INSTANCES of Squarer the method `document`.  <http://blog.jayfields.com/2006/05/ruby-extend-and-include.html>
  def square(x) # Define our crappy undocumented function.
    puts x**2
    end
  document(:square)  # this is the same as `self.document`.  `self` here is the CLASS.  Because we EXTENDED it, we have access to `document` from the class rather than an instance.  `square()` is now jazzed up for every instance of Squarer.
```
The proposed decorator should `pull` changes for every method that potentially will create a change in the `ci-config` repo (`create_user!`, `create_user!`, etc.) **before** the method is actually being called. With this approach we ensure that we **always** have the latest version of the repo, and thus, we cannot have any conflict (because those `writing` or `read` methods, make changes after the pull is being done) and when the change is actually being done, we have already implemented `Watchmen` (see #283 for more info) which is responsible of committing every changed file on-demand and pushing it to remote.

This is something I'm totally in favor of implement in my current PR (which is close of being finished) but want to have other opinions on this. 

I really think that this adds a verbosity desired and will prevent future problems with the `ci-config` repo operations. That's why I have rebuilt it from the ground up.

Sorry for the long post, here is a 🥔tho.

PR #318, opened 2018-03-17
Disable broken user service test that made assumption about user object

taquitos:
- User object returned from data source is a new instance and therefor
comparing their identities fails.


AndrewMcBurney:
For the future, instead of commenting out the test, we can use rspec [`xit`](https://apidock.com/rspec/Spec/Example/ExampleGroupMethods/xit) instead to mark the test as pending.

PR #321, opened 2018-03-17
Don't call `split` on the `FastlaneCI.env.repo_url` if it doesn't exist.

AndrewMcBurney:
This is an onboarding edge-case where the user's local configuration repo doesn't exist, and the server tries to see if the remote configuration repository is valid without a valid `ENV["FASTLANE_CI_REPO_URL"]`.

https://github.com/fastlane/ci/blob/afd7dacbbcc47f800d3cd57bac2abf9c87f15f56/launch.rb#L22

https://github.com/fastlane/ci/blob/afd7dacbbcc47f800d3cd57bac2abf9c87f15f56/launch.rb#L68-L69

> When the first expression in the conditional is true (i.e., no local repository exists), and it checks the second expression in the conditional

https://github.com/fastlane/ci/blob/afd7dacbbcc47f800d3cd57bac2abf9c87f15f56/services/configuration_repository_service.rb#L160

> This `split` will fail if the `ENV["FASTLANE_CI_REPO_URL"]` has not been set

### DIFF
--- a/services/configuration_repository_service.rb
+++ b/services/configuration_repository_service.rb
@@ -157,6 +157,7 @@ module FastlaneCI
     #
     # @return [String]
     def repo_name
+      return "" unless FastlaneCI.env.repo_url
       FastlaneCI.env.repo_url.split("/").last
     end
 
@@ -164,6 +165,7 @@ module FastlaneCI
     #
     # @return [String]
     def repo_shortform
+      return "" unless FastlaneCI.env.repo_url
       FastlaneCI.env.repo_url.split("/").last(2).join("/")
     end
   end


services/configuration_repository_service.rb
```diff
@@ -157,13 +157,15 @@ def remote_file_a_json_array?(file_path)
...
+      return "" unless FastlaneCI.env.repo_url
       FastlaneCI.env.repo_url.split("/").last
     end
 
     # The short-form of the configuration repository URL `ueser/repo`
     #
     # @return [String]
     def repo_shortform
+      return "" unless FastlaneCI.env.repo_url
       FastlaneCI.env.repo_url.split("/").last(2).join("/")
```
taquitos: Coming in late here, but 2 things:
1) missing a return statement
2) Maybe it could be rewritten to be `return FastlaneCI.env.repo_url ? FastlaneCI.env.repo_url.split("/").last(2).join("/") : ""` 

but this is totally stylistic ^

snatchev: style wise, I prefer the "guard" statement with an early return over the ternary operator.

taquitos: Great point @snatchev, definitely prefer what is idiomatic, I'm still learning those bits about Ruby 👍

minuscorp: That's so Swifty's `if let` 🤣

AndrewMcBurney: We have this pattern [here](https://github.com/fastlane/ci/blob/master/shared/models/github_provider_credential.rb#L31-L32), so I thought it was cool to use it [here](https://github.com/fastlane/ci/blob/master/services/configuration_repository_service.rb#L160-L161) as well. I agree with @snatchev that the guard statement seems more like idiomatic ruby, but I'm open to changing it to the normal `if/else` statement if it's more readable.

AndrewMcBurney: Also, I included the extra return in my [explicit return PR](https://github.com/fastlane/ci/pull/325/files#diff-51bddc1ca8b9376a96ac62cf51d8814fR161). I kept them separate but worked on them both at the same time.

PR #324, opened 2018-03-17
Reduce global static method code smells and reduce responsibilities of `Launch` class.

AndrewMcBurney:
This introduces fairly substantial changes, but I believe they're necessary in order to reduce code smells that have accumulated over time (largely due to me).

### Changes

The two main goals of this PR are to:

1. Decouple business logic from the static methods in the `Launch` class, and move them to appropriate services
2. Remove references to the static helper functions defined in the `Services` class, and make them `private_class_method`s. Instead of using these helpers as global static methods, inject them into the appropriate services in their constructors.

### DIFF
--- a/services/config_service.rb
+++ b/services/config_service.rb
@@ -11,8 +11,10 @@ module FastlaneCI
     attr_accessor :project_service
     attr_accessor :ci_user
     attr_accessor :active_code_hosting_services # dictionary of active_code_hosting_service_key to CodeHosting
+    attr_accessor :clone_user_provider_credential
 
-    def initialize(project_service: FastlaneCI::Services.project_service, ci_user: nil)
+    def initialize(project_service: FastlaneCI::Services.project_service, ci_user: nil, clone_user_provider_credential:)
+      @clone_user_provider_credential = clone_user_provider_credential
       self.project_service = project_service
       self.ci_user = ci_user
       self.active_code_hosting_services = {}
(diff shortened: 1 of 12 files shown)


services/config_service.rb
```diff
@@ -11,8 +11,10 @@ class ConfigService
     attr_accessor :project_service
     attr_accessor :ci_user
     attr_accessor :active_code_hosting_services # dictionary of active_code_hosting_service_key to CodeHosting
+    attr_accessor :clone_user_provider_credential
 
-    def initialize(project_service: FastlaneCI::Services.project_service, ci_user: nil)
+    def initialize(project_service: FastlaneCI::Services.project_service, ci_user: nil, clone_user_provider_credential:)
+      @clone_user_provider_credential = clone_user_provider_credential
```
taquitos: we can use `self.<variable>` = for our initialization now. It's best to standardize on something, and it seems to be the most popular (CC @KrauseFx who mentioned this a while back to me, I think)

services/config_service.rb
```diff
@@ -28,6 +30,89 @@ def active_code_hosting_service_key(provider_credential: nil)
...
+        pending_build_shas_needing_rebuilds.each do |sha|
+          logger.debug("Found sha #{sha} that needs a rebuild for #{project.project_name}")
+          build_runner = FastlaneBuildRunner.new(
+            project: project,
+            sha: sha,
+            github_service: github_service,
+            work_queue: FastlaneCI::GitRepo.git_action_queue # using the git repo queue because of https://github.com/ruby-git/ruby-git/issues/355
+          )
+          build_runner.setup(parameters: nil)
+          Services.build_runner_service.add_build_runner(build_runner: build_runner)
```
taquitos: So, in general, services shouldn't really know about each other. Ideally we'd have an instance of something that uses these two two services: `config_service` and `build_runner_service`
It might need something like a `SystemStartupCoordinator` to act like a controller without a view (coordinators are controllers without views? I just made that up.)

services/config_service.rb
```diff
@@ -28,6 +30,89 @@ def active_code_hosting_service_key(provider_credential: nil)
...
+
+        # Collect all the branches from the triggers on this project that are commit-based
+        branches_to_check = project.job_triggers
+                                   .select { |trigger| trigger.type == FastlaneCI::JobTrigger::TRIGGER_TYPE[:commit] }
+                                   .map(&:branch)
+                                   .uniq
+
+        repo_full_name = project.repo_config.full_name
+        # let's get all commit shas that need a build (no status yet)
+        commit_shas = github_service.last_commit_sha_for_all_open_pull_requests(repo_full_name: repo_full_name, branches: branches_to_check)
```
taquitos: Same here, really, this logic should be broken down between the services and a coordinator. I'm the worst, I know.

services/onboarding_service.rb
```diff
@@ -7,18 +7,32 @@ module FastlaneCI
...
     include FastlaneCI::Logging
 
+    # @return [GitRepoConfig]
+    attr_reader :ci_config_repo
+
+    # @return [GitHubProviderCredential]
+    attr_reader :clone_user_provider_credential
+
+    def initialize(ci_config_repo:, clone_user_provider_credential:)
+      @ci_config_repo = ci_config_repo
```
taquitos: `self.<variable> = <variable>`

services/onboarding_service.rb
```diff
@@ -7,18 +7,32 @@ module FastlaneCI
...
+      @clone_user_provider_credential = clone_user_provider_credential
+    end
+
     # Verify that fastlane.ci is already set up on this machine.
     # If that's not the case, we have to make sure to trigger the initial clone
     def trigger_initial_ci_setup
       logger.info("No config repo cloned yet, doing that now")
 
       # Trigger the initial clone
+      # TODO: remote this in favour of a different approach
```
taquitos: `remote` -> remove
Could you also comment what you think would be a different/better approach?

services/project_service.rb
```diff
@@ -12,12 +12,15 @@ module FastlaneCI
...
+    attr_accessor :clone_user_provider_credential
+    attr_accessor :configuration_git_repo
 
-    def initialize(project_data_source: nil)
+    def initialize(project_data_source: nil, clone_user_provider_credential:, configuration_git_repo:)
       unless project_data_source.nil?
         raise "project_data_source must be descendant of #{ProjectDataSource.name}" unless project_data_source.class <= ProjectDataSource
       end
 
+      @clone_user_provider_credential = clone_user_provider_credential
```
taquitos: `self. ___ = `

services/services.rb
```diff
@@ -79,46 +79,49 @@ def self.ci_user
...
+      @_clone_user_provider_credential ||= GitHubProviderCredential.new(
         email: FastlaneCI.env.initial_clone_email,
         api_token: FastlaneCI.env.clone_user_api_token
       )
     end
 
+    # These service helper methods should not be exposed, since they become
+    # global static methods, which when referenced in arbitrary classes, becomes
+    # a code smell
+    private_class_method :ci_config_git_repo_path, :ci_config_repo,
```
taquitos: I like that this is a thing I didn't know about, thank you

services/worker_service.rb
```diff
@@ -10,8 +10,12 @@ class WorkerService
     include FastlaneCI::Logging
 
     attr_accessor :project_to_workers_dictionary
+    attr_accessor :ci_user
+    attr_accessor :provider_credential
 
-    def initialize
+    def initialize(ci_user:, provider_credential:)
+      @ci_user = ci_user
+      @provider_credential = provider_credential
```
taquitos: `self.___ = `

services/worker_service.rb
```diff
@@ -47,6 +51,32 @@ def start_workers_for_project_and_credential(project: nil, provider_credential:
...
+    def start_github_workers
+      return unless Services.onboarding_service.correct_setup?
+      launch_workers unless ENV["FASTLANE_CI_SKIP_WORKER_LAUNCH"]
+    end
+
+    def launch_workers
+      logger.debug("Starting workers to monitor projects")
+      # Iterate through all provider credentials and their projects and start a worker for each project
+      ci_user.provider_credentials.each do |provider_credential|
+        projects = Services.config_service.projects(provider_credential: provider_credential)
```
taquitos: More services calling into a global service.
I think we need to find a way to coordinate this in a less global way. 
Either through coordinator usage, or dependency injection, or something else.
The `launch.rb` was doing all this stuff because it was a one-time setup, in one place, and IMO that's ok for this code smell stuff, as long as all the smell comes from 1 file, that's ok.

taquitos:
Ok, overall I love the idea behind this PR, and I think it mostly accomplishes what you set out todo 🎉 I'd love to have your thoughts on the idea of limiting services from calling into global services.

services/config_service.rb
AndrewMcBurney: Yeah, that's cool with me - I'll make a follow-up PR to remove other instances in the codebase where we use `@variable =` too.

services/config_service.rb
AndrewMcBurney: Yeah, I agree with you 💯%. We have a lot of services which already have other knowledge of each other - do you think this issue is okay to tackle in another PR where we introduce the `SystemStartupCoordinator` you mentioned - or would you like that done in this PR?

services/config_service.rb
AndrewMcBurney: No, you're not the worst - I'd prefer to not have this coupled so it doesn't cause design headaches in the future.

services/onboarding_service.rb
AndrewMcBurney: Oops, yeah - autocomplete ftl (_for the loss_).

I don't like how the clone is done implicitly upon invoking the constructor. It's not apparent creating the object will do a remote call to the GitHub API. I'd prefer to have the constructor not have that responsibility, and make a `clone_project!` method which should be invoked after `new` is called.

This would make the clone a bit more apparent. The downside to this approach is those two methods become tightly coupled with object instantiation - but I think the tradeoff of being more explicit is worth it.

Thoughts?

minuscorp: #328 🙄

minuscorp: #283 sorry

taquitos: I'm just now getting a chance to test #283 @minuscorp, been a busy day, sorry!

minuscorp: Hey! No need to rush, just pointing out to avoid wasting people's time 😁

taquitos:
Hey @AndrewMcBurney if you have time to address the feedback and update the PR, that would be awesome, but I totally understand if you don't. I'm going to close this for now though 🌮

PR #326, opened 2018-03-17
Removed dead code in `notifications_controller` and refactored view helpers.

AndrewMcBurney:
I initially thought it would be cool to expose an API for the notifications controller so that you could create notifications externally as detailed in https://github.com/fastlane/ci/pull/142.

I don't really know if it makes sense to expose an API, and think it would be better to just create/update notifications through the [`Services.notification_service`](https://github.com/fastlane/ci/blob/master/services/notification_service.rb), and just use the notification controller to display them.

Thoughts?


taquitos:
For now, I think that's a good idea. It's a more advanced feature that we don't plan on using just yet 👍

Issue #242, opened 2018-03-08
Management UI shouldn't generate/show the UDID ahead of time


KrauseFx:
<img width="1091" alt="screen shot 2018-03-08 at 1 49 44 pm" src="https://user-images.githubusercontent.com/869950/37170001-9df79abe-22d7-11e8-9193-bb00a3ad82fb.png">

This seems like it will potentially cause security issues, let's generate the UDID only when receiving the `POST` request, so that the user can't edit the UDID on the client-side.

PR #327, opened 2018-03-17
[WIP] Don't pre-generate the object IDs

AndrewMcBurney:
https://github.com/fastlane/ci/issues/242

Don't display a generated UUID for user ahead of time in creation form for `users_controller`.


taquitos:
This has gotten too out of sync with master, but this change is definitely something we want to have. I'm going to close this for now, and keep the issue (#242) open so that we can address this in the near future.

PR #331, opened 2018-03-20
Thinned out Angular Web App Files

nakhbari:
Removed end to end tests and other files that are not needed for the web app


minuscorp:
The thinning is real 💯

taquitos:
@nakhbari I think your github name/email might not match what your local .git has, which is why CLA bot doesn't like your commit. It happens a lot when you edit stuff from the github ui instead of locally and then pushing to your branch. No big deal.

taquitos:
🌮🐐

Issue #340, opened 2018-03-21
Build always report success even on failure


taquitos:
Builds were reporting success back to github, even on failure

PR #341, opened 2018-03-21
Improvements to build reporting

taquitos:
- Build failures form UI.error now reporting :error are caught by the build status (#340)
- Clearer reporting of status setting (project, status, and commit sha)
- Reporting of server errors vs fastlane errors in status now supported with new flag :ci_problem
- rubocop updates

### DIFF
--- a/features/build_runner/build_runner_output_row.rb
+++ b/features/build_runner/build_runner_output_row.rb
@@ -3,7 +3,7 @@ module FastlaneCI
   # meaning it's a single message that we want to show to the user
   # and store as a build artifact as part of the build log
   class BuildRunnerOutputRow
-    BUILD_FAIL_TYPES = [:user_error, :build_error, :crash, :shell_error, :build_failure, :test_failure, :abort].to_set
+    BUILD_FAIL_TYPES = [:user_error, :build_error, :crash, :error, :shell_error, :build_failure, :test_failure, :abort].to_set
 
     # The type of message (e.g. `:message`, `:error`, `:important`)
     attr_accessor :type
--- a/features/build_runner/fastlane_build_runner.rb
+++ b/features/build_runner/fastlane_build_runner.rb
@@ -31,6 +31,7 @@ module FastlaneCI
       @platform = lane_pieces.count > 1 ? lane_pieces.first : nil
       @lane = lane_pieces.last
       @parameters = parameters
+      @encountered_failure_output = false # Did we encounter a row that signaled failure?
     end
 
     # completion_block is called with an array of artifacts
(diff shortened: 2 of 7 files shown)


features/build_runner/build_runner_output_row.rb
```diff
@@ -3,7 +3,7 @@ module FastlaneCI
   # meaning it's a single message that we want to show to the user
   # and store as a build artifact as part of the build log
   class BuildRunnerOutputRow
-    BUILD_FAIL_TYPES = [:user_error, :build_error, :crash, :shell_error, :build_failure, :test_failure, :abort].to_set
+    BUILD_FAIL_TYPES = [:user_error, :build_error, :crash, :error, :shell_error, :build_failure, :test_failure, :abort].to_set
```
taquitos: @KrauseFx I noticed that `UI.error` in a Fastfile always triggers `:error` for a fail type, so I added that here. Is this something we just missed, or do we expect it to be something else?

features/build_runner/fastlane_build_runner.rb
```diff
@@ -73,10 +74,9 @@ def run(completion_block: nil, &block)
         # Execute the Fastfile here
         logger.info("starting fastlane run lane: #{self.lane} platform: #{self.platform}, params: #{self.parameters} from #{fast_file_path}")
 
-        build_success = true
         # Attach a listener to the output to see if we have a failure. If so, this build failed
         self.add_listener(proc do |row|
-          build_success = false if row.did_fail_build?
```
taquitos: I'd rather use an instance variable here

features/build_runner/fastlane_build_runner.rb
```diff
@@ -102,7 +102,8 @@ def run(completion_block: nil, &block)
         completion_block.call(artifacts)
       rescue StandardError => ex
         logger.debug("Setting build status to failure due to exception")
-        self.current_build.status = :failure
+        self.current_build.status = :ci_problem
```
taquitos: now we can tell when the runner crashes vs when the test fails 🎉

launch.rb
```diff
@@ -42,7 +42,7 @@ def self.require_fastlane_ci
 
     def self.verify_app_built
       if ENV["WEB_APP"]
-        app_exists = File.file?(File.join('public', '.dist', 'index.html'))
+        app_exists = File.file?(File.join("public", ".dist", "index.html"))
```
taquitos: @nakhbari before submitting, do a quick `bundle exec rubocop -a` 😄

services/code_hosting/git_hub_service.rb
```diff
@@ -121,13 +121,13 @@ def set_build_status!(repo: nil, sha: nil, state: nil, target_url: nil, descript
         description = "Still running" if state == "pending"
 
         # TODO: what's the difference?
-        description = "Something went wrong" if state == "failure"
-        description = "Something went wrong" if state == "error"
+        description = "Build encountered a failure" if state == "failure"
```
taquitos: Just in case we start using different states, I'd like to see it show a slightly different message

launch.rb
nakhbari: Shoot sorry about that. You're supposed to use single tick in JS/TS got used to it

PR #342, opened 2018-03-21
Added JSON endpoint for projects and data server in the app

nakhbari:
- Added Project JSON controller in new features-json folder. This is where the controllers return json exist. It's separate from the other features for now for separation of concerns.
- Added Project Model in web app which will be used to wrap the Project data we receive from the backend. 
- Added DataService in Angular which makes an http request to the backend to get the project data.
- Added unit tests for both the Project model and the DataService

Tested Locally

Next step is to make an Overview Page with all the projects.

### DIFF
--- a/app/src/app/services/data.service.ts
+++ b/app/src/app/services/data.service.ts
@@ -1,0 +1,20 @@
+import {HttpClient} from '@angular/common/http';
+import {Injectable} from '@angular/core';
+import {Observable} from 'rxjs/Observable';
+import {map} from 'rxjs/operators';
+
+import {Project, ProjectResponse} from '../models/project';
+
+// Data server is currently locally hosted.
+const HOSTNAME = '';
+
+@Injectable()
+export class DataService {
+  constructor(private http: HttpClient) {}
+
+  getProjects(): Observable<Project[]> {
+    const url = `${HOSTNAME}/projects`;
+    return this.http.get<ProjectResponse[]>(url).pipe(
+        map((projects) => projects.map((project) => new Project(project))));
+  }
+}
--- a/features-json/project_json_controller.rb
+++ b/features-json/project_json_controller.rb
@@ -1,0 +1,16 @@
+require_relative "../shared/authenticated_controller_base"
+
+module FastlaneCI
+  # Controller for providing all data relating to projects
+  class ProjectJsonController < AuthenticatedControllerBase
+    HOME = "/projects"
+
+    get "#{HOME}" do
+      current_provider_credential = self.check_and_get_provider_credential
+      current_user_config_service = self.current_user_config_service
+      all_projects = current_user_config_service.projects(provider_credential: current_provider_credential)
+
+      return all_projects.to_json
+    end
+  end
+end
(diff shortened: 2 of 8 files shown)


features-json/project_json_controller.rb
```diff
@@ -0,0 +1,16 @@
+require_relative "../shared/authenticated_controller_base"
+
+module FastlaneCI
+  # Controller for providing all data relating to projects
+  class ProjectJsonController < AuthenticatedControllerBase
```
taquitos: I think the general consensus is to use JSON in all caps for json when it comes to class naming `ProjectJsonController` -> `ProjectJSONController`

launch.rb
```diff
@@ -110,6 +110,11 @@ def self.register_available_controllers
       FastlaneCI::FastlaneApp.use(FastlaneCI::ProviderCredentialsController)
       FastlaneCI::FastlaneApp.use(FastlaneCI::UsersController)
       FastlaneCI::FastlaneApp.use(FastlaneCI::BuildController)
+
+      # Load JSON controllers
+      require_relative "features-json/project_json_controller"
```
taquitos: I like have you have this space down here, it should help with avoiding merge conflicts 😄

app/src/app/services/data.service.ts
```diff
@@ -0,0 +1,20 @@
+import {HttpClient} from '@angular/common/http';
+import {Injectable} from '@angular/core';
+import {Observable} from 'rxjs/Observable';
```
taquitos: Ohhh boy, reactive js?

taquitos: Ok, so let's chat about this one. Last project I worked on with a reactive core couldn't find people who knew it, ramp up time was really long, and overall, it was very difficult to make large changes without rewriting all the things. In my experience, if you plan on making changes and experimenting reactive slows you down as you have to remodel how your data flows over and over again.

nakhbari: yeah Angular uses RxJS. Looking at Observables they're pretty cool.

taquitos: That's not to say we can't use it, I just want to make sure we're making the right decision here and that this addition will meet all our goals

features-json/project_json_controller.rb
nakhbari: done

app/src/app/services/data.service.ts
nakhbari: @taquitos so i believe we have to use it. I just asked around, and it seems like the only way not to use it is to immediately convert to promises. 

I think the bigger question is should we be using ngRX. That is something that deeply uses React. Observables are really only conceptually minor changes to promises.

nakhbari: I think we can punt the ngRX discussion down the line.

features-json/project_json_controller.rb
taquitos: ngRX might be overkill for now. 
Observables are really cool, in theory, but great care must be taken when you start operate with actual data. I would suggest checking out: http://rxmarbles.com, it has all the operators with interactive graphical representations of what they do.
There's also a really great iOS app: https://itunes.apple.com/us/app/rxmarbles/id1087272442

PR #345, opened 2018-03-22
moving submodule projects into vendored gems

snatchev:
This PR makes the external submodule dependencies as vendored gems. The benefit is that we can take updates easier and have only 1 process for keeping dependencies up to date. Also, this opens the ability to not use submodules by using bundler's `:git` source. http://bundler.io/v1.12/git.html#gems-from-git-repositories

https://github.com/fastlane/TaskQueue/pull/7
https://github.com/fastlane/fastfile-parser/pull/9

These gems are not intended to be publicly available yet.

PR #350, opened 2018-03-22
Move application code into `app/` directory

snatchev:
This PR builds on top of #345 which puts external dependencies in `vendor/`

Motivation
----
The root directory of the project contains a lot of meta-data, static assets, docs, and development configs. I believe we should leverage pr #298 which creates a root-level `app/` directory to contain source files of the application. I have added a document in `docs/ProjectLayout.md` as a place (WIP) to describe how to find and organize different source files.

The PR also makes the `require` statements in specs more idiomatic. A spec should always start with `require "spec_helper"` to setup the test (if a test is to be run in isolation) and then requires the file to be tested.


taquitos:
I just tested this out, and it's great. I'm going to merge it because I'm about to make some changes that might conflict unless I do.

Issue #265, opened 2018-03-09
Dashboard status display not implemented yet


taquitos:
<img width="504" alt="screen shot 2018-03-09 at 9 49 10 am" src="https://user-images.githubusercontent.com/1304821/37221535-3b2fcbf6-237f-11e8-8e2a-3ed5610792b6.png">

<img width="420" alt="screen shot 2018-03-09 at 9 49 14 am" src="https://user-images.githubusercontent.com/1304821/37221539-3e583444-237f-11e8-80c5-dc595cc0083a.png">

KrauseFx:
I love how we use the 🌔 emoji 👍

KrauseFx:
I rephrased the title, as currently we just have a static emoji there 👍

taquitos:
🎉

PR #358, opened 2018-03-23
Dashboard last build status per project

viktorasl:
Tiny contribution. Solves #265. Questionable: I've added build_status.erb but it basically contains mapping from status to status emoji.


taquitos:
I think that's perfect, love that you extracted out that logic. We might update it in the future, so that makes it easier once we have designs 🎉

PR #360, opened 2018-03-23
Added FastfilePeeker

minuscorp:
This utility class adds a `peek` method using a `GitRepo` instance and a `branch` or a `sha` to peek its Fastfile configuration.
Uses a unique cache identifier key to store in-memory previously peeked repos (enabled by default, disabled by parameter).

### DIFF
--- a/services/fastfile_peeker/fastfile_peeker.rb
+++ b/services/fastfile_peeker/fastfile_peeker.rb
@@ -32,12 +32,11 @@ module FastlaneCI
         hash = Digest::SHA2.hexdigest(cache_key)
         return self.cache[hash] if self.cache[hash].kind_of?(Fastlane::FastfileParser)
         git_repo.fetch
-        git = git_repo.git
         if !branch.nil? || !branch.empty?
           # This perform the checkout of the latest commit in the branch.
           git_repo.checkout_branch(branch)
         elsif !sha.nil? || !sha.empty?
-          git.chekout_commit(sha)
+          git_repo.chekout_commit(sha)
         else
           raise "Invalid branch or sha where provided"
         end
--- a/services/fastfile_peeker/fastfile_peeker.rb
+++ b/services/fastfile_peeker/fastfile_peeker.rb
@@ -1,5 +1,4 @@
-require_relative "../../fastfile-parser/fastfile_parser"
-
+require "fastfile_parser"
 require "digest"
 
 module FastlaneCI
--- a/services/fastfile_peeker/fastfile_peeker.rb
+++ b/services/fastfile_peeker/fastfile_peeker.rb
@@ -1,0 +1,54 @@
+require_relative "../../fastfile-parser/fastfile_parser"
+
+require "digest"
+
+module FastlaneCI
+  # Utility class designed for parsing Fastfiles in a repo.
+  class FastfilePeeker
+    class << self
+      def cache
+        @cache ||= {}
+        return @cache
+      end
+
+      # Class method that finds the directory for the first Fastfile found given a root_path
+      # @param [String] root_path
+      # @return [String, nil] the path of the Fastfile or nil if no Fastfile found.
+      def fastfile_path(root_path: nil)
+        fastfiles = Dir[File.join(root_path, "fastlane/Fastfile")]
+        fastfiles = Dir[File.join(root_path, "**/fastlane/Fastfile")] if fastfiles.count == 0
+        fastfile_path = fastfiles.first
+        return fastfile_path
+      end
+
+      protected :cache, :fastfile_path
+
+      # @param [GitRepo] git_repo
+      # @param [String, nil] branch
+      # @param [String, nil] sha
+      # @return [Fastlane::FastfileParser]
+      def peek(git_repo: nil, branch: nil, sha: nil)
+        cache_key = git_repo.git_config.git_url + (branch || sha)
+        hash = Digest::SHA2.hexdigest(cache_key)
+        return self.cache[hash] if self.cache[hash].kind_of?(Fastlane::FastfileParser)
+        git_repo.fetch
+        git = git_repo.git
+        if !branch.nil? || !branch.empty?
+          # This perform the checkout of the latest commit in the branch.
+          git_repo.checkout_branch(branch)
+        elsif !sha.nil? || !sha.empty?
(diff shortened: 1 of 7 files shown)


minuscorp:
Gonna rebase this for tomorrow and get ready for review!

services/fastfile_peeker/fastfile_peeker.rb
```diff
@@ -0,0 +1,55 @@
+require_relative "../../fastfile-parser/fastfile_parser"
+
+require "digest"
+
+module FastlaneCI
+  # Utility class designed for parsing Fastfiles in a repo.
+  class FastfilePeeker
```
KrauseFx: I love the name 😍

snatchev:
@minuscorp this looks good. Any chance we could add a test for at least the `#peek` method? I would be happy to work with you on adding an rspec if you're not familiar with it. Let me know.

minuscorp:
I'm familiar with `rspec`, but I'm afraid I'm not having time for adding the tests by today. Maybe for Wednesday I can have it, out of you want to help me out with them I'm happy too😁

minuscorp:
Trying to get this covered by tomorrow🙌

services/fastfile_peeker/fastfile_peeker.rb
```diff
@@ -0,0 +1,55 @@
...
+    class << self
+      def cache
+        @cache ||= {}
```
KrauseFx: This could just be 
```ruby
@_cache ||= {}
```

services/fastfile_peeker/fastfile_peeker.rb
```diff
@@ -0,0 +1,55 @@
...
+          return self.cache[hash] if self.cache[hash].kind_of?(Fastlane::FastfileParser)
+        end
+        git_repo.fetch
+        if !branch.nil? || !branch.empty?
+          # This perform the checkout of the latest commit in the branch.
+          git_repo.checkout_branch(branch)
+        elsif !sha.nil? || !sha.empty?
+          git_repo.chekout_commit(sha)
+        else
+          raise "Invalid branch or sha where provided"
```
KrauseFx: "were"

services/fastfile_peeker/fastfile_peeker.rb
```diff
@@ -0,0 +1,55 @@
...
+      # @param [String, nil] sha
+      # @return [Fastlane::FastfileParser]
+      def peek(git_repo: nil, branch: nil, sha: nil, cache: true)
+        cache_key = git_repo.git_config.git_url + (branch || sha)
+        hash = Digest::SHA2.hexdigest(cache_key)
+        if cache
+          return self.cache[hash] if self.cache[hash].kind_of?(Fastlane::FastfileParser)
+        end
+        git_repo.fetch
+        if !branch.nil? || !branch.empty?
```
KrauseFx: Lots of negatives and `||` etc, could we refactor to have the `if else` block more naturally?

services/fastfile_peeker/fastfile_peeker.rb
```diff
@@ -0,0 +1,55 @@
...
+          # This perform the checkout of the latest commit in the branch.
+          git_repo.checkout_branch(branch)
+        elsif !sha.nil? || !sha.empty?
+          git_repo.chekout_commit(sha)
+        else
+          raise "Invalid branch or sha where provided"
+        end
+        if (fastfile_path = self.fastfile_path(root_path: git_repo.local_folder))
```
KrauseFx: I'm not 100% sold on this pattern, it makes it slightly more confusing to read, can we split this into more lines maybe to make it easier for new contributors?

services/fastfile_peeker/fastfile_peeker.rb
minuscorp: Just for convention, I thought somewhere we discussed about making explicit `return` statements to improve code readability

minuscorp:
Added tests, ready for review + merge 😄

services/fastfile_peeker/fastfile_peeker.rb
```diff
@@ -0,0 +1,55 @@
...
+    class << self
+      def cache
+        @cache ||= {}
+        return @cache
+      end
+
+      # Class method that finds the directory for the first Fastfile found given a root_path
+      # @param [String] root_path
+      # @return [String, nil] the path of the Fastfile or nil if no Fastfile found.
+      def fastfile_path(root_path: nil)
```
taquitos: We have very similar code inside `FastfileFinder` : https://github.com/fastlane/ci/blob/master/app/shared/fastfile_finder.rb#L16

Should we reference the `FastfileFinder` here or do something else?

services/fastfile_peeker/fastfile_peeker.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+
+      protected :cache, :fastfile_path
+
+      # @param [GitRepo] git_repo
+      # @param [String, nil] branch
+      # @param [String, nil] sha
+      # @return [Fastlane::FastfileParser]
+      def peek(git_repo: nil, branch: nil, sha: nil, cache: true)
+        cache_key = git_repo.git_config.git_url + (branch || sha)
```
taquitos: I think we should just remove the cache for now. It adds complexity and I don't think the performance gains are big enough to worry about right now.

services/fastfile_peeker/fastfile_peeker.rb
minuscorp: Didn't knew about that, gonna use it instead of this method

services/fastfile_peeker/fastfile_peeker.rb
minuscorp: `peek` is being done each time the user selects a new `branch` in the `New Project` workflow. I can't imagine the user waiting to the whole repo being cloned each time he changes the branch from the selector in the interface. As I think that cache might be complex in certain scenarios, I think this must be keeped for this one

minuscorp: I correct myself, as it has a direct dependency on `GitRepo` and it will be cloned whether the cache is used or not, I'll delete it until the `GitRepo` API allows to clone the repo out of the initializer.

PR #361, opened 2018-03-23
Fix onboarding without existing configuration.

minuscorp:
Fixes: 
```
NoMethodError: undefined method `git_url' for nil:NilClass
  /Users/GabrielVillarrubia/Documents/ci/shared/models/git_repo.rb:72:in `initialize'
  /Users/GabrielVillarrubia/Documents/ci/services/config_data_sources/json_project_data_source.rb:97:in `new'
  /Users/GabrielVillarrubia/Documents/ci/services/config_data_sources/json_project_data_source.rb:97:in `after_creation'
  /Users/GabrielVillarrubia/Documents/ci/services/data_sources/json_data_source.rb:22:in `create'
  /Users/GabrielVillarrubia/Documents/ci/services/onboarding_service.rb:19:in `clone_remote_repository_locally'
  /Users/GabrielVillarrubia/Documents/ci/launch.rb:80:in `clone_repo_if_no_local_repo_and_remote_repo_exists'
  /Users/GabrielVillarrubia/Documents/ci/launch.rb:23:in `take_off'
```

### DIFF
--- a/app/services/onboarding_service.rb
+++ b/app/services/onboarding_service.rb
@@ -1,5 +1,6 @@
 require "json"
 require_relative "../shared/logging_module"
+require_relative "../services/services"
 
 module FastlaneCI
   # Provides operations to create and mutate the FastlaneCI configuration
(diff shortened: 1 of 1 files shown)


services/onboarding_service.rb
```diff
@@ -80,7 +75,11 @@ def local_configuration_repo_exists?
 
     # @return [Boolean]
     def remote_configuration_repository_valid?
-      return Services.configuration_repository_service.configuration_repository_valid?
+      if Services.configuration_repository_service.respond_to?(:configuration_repository_valid?)
```
taquitos: Weird, why do we need to check this?

minuscorp: I have encountered cases when `configuration_repository_service` was nil, but couldn't track the source of this issue, so I made this extra safe call just in case. Don't know if this is the best way to proceed tho.

taquitos: Ohhh weird, I think we should leave this check off then. That way when we encounter it we can fix it. If we can't find it for some reason, we can always follow up and add it back

minuscorp: Done! Rebasing tomorrow😁

taquitos: 🎉

PR #362, opened 2018-03-23
Avoid duplicate runners

minuscorp:
Filter `build_runner`s that have been already queued. It happens a lot with `CheckForNewCommitsWorker`.

### DIFF
--- a/app/services/build_runner_service.rb
+++ b/app/services/build_runner_service.rb
@@ -33,7 +33,7 @@ module FastlaneCI
     end
 
     # Fetch all the active runners, and see if there is one WIP
-    def find_build_runner(project_id: nil, sha: nil, build_number: nil)
+    def find_build_runner(project_id:, sha: nil, build_number: nil)
       return self.build_runners.find do |build_runner|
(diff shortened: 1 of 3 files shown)


services/build_runner_service.rb
```diff
@@ -33,9 +33,9 @@ def build_runner_task_queue
     end
 
     # Fetch all the active runners, and see if there is one WIP
-    def find_build_runner(project_id:, build_number:)
+    def find_build_runner(project_id: nil, sha: nil, build_number: nil)
```
taquitos: I think you might want to leave off `nil` for these params: `project_id: nil, sha: nil, build_number: nil`
that way all of them  are required: `(project_id:, sha:, build_number:)`

minuscorp: The idea is that you can either filter by `sha` or `build_number`, being the `project_id` mandatory in all cases (that way I don't break things)

taquitos: Ahh ok, maybe just remove the `nil` for `project_id: nil`?

minuscorp:
I rebase tomorrow and merge it, pretty late here🌙

taquitos:
Sleep well 💤

PR #363, opened 2018-03-23
Make use of the FastfilePeeker

minuscorp:
- Depends on #360

### DIFF
--- a/app/features/project/project_controller.rb
+++ b/app/features/project/project_controller.rb
@@ -1,5 +1,7 @@
 require_relative "../../shared/authenticated_controller_base"
 require_relative "../../shared/models/git_repo"
+require_relative "../../services/fastfile_peeker/fastfile_peeker"
+require_relative "../../shared/fastfile_finder"
 
 require "pathname"
 require "securerandom"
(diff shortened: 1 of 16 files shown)


features/project/project_controller.rb
```diff
@@ -70,11 +70,32 @@ class ProjectController < AuthenticatedControllerBase
       erb(:new_project, locals: locals, layout: FastlaneCI.default_layout)
     end
 
+    get "#{HOME}/lanes/*/*/*" do |org, repo_name, branch|
```
KrauseFx: What is this URL used for?

KrauseFx: Maybe we can add a comment here with a little more details

features/project/project_controller.rb
```diff
@@ -70,11 +70,32 @@ class ProjectController < AuthenticatedControllerBase
...
+      content_type :json
+
+      provider_credential = check_and_get_provider_credential(type: FastlaneCI::ProviderCredential::PROVIDER_CREDENTIAL_TYPES[:github])
+
+      selected_repo = github_service.repos.detect { |repo| repo_name == repo.name }
```
KrauseFx: This will also make sure you can only access files you actually have access to, right?

features/project/views/new_project_form.erb
```diff
@@ -4,27 +4,78 @@
       <span class="mdl-layout__title"><%= title %></span>
     </div>
   </header>
-
+  <script>
+  document.addEventListener("DOMContentLoaded", function() {
```
KrauseFx: Why do we all of this on the front-end?

minuscorp: So the main idea was to first display the form of `New Project` as we have today, but now provide a list of `branches` to select. Whenever the user selects a new `branch`, we request to the backend the `lanes` and `platforms` existing for a given repo and a given branch.

features/project/project_controller.rb
minuscorp: `GitHubService` uses the `api_token` from the `provider_credential` so it'll only show up repos which the user has access to

features/project/project_controller.rb
minuscorp: That's to be done once #360 is merged and I can actually test that everything works fine 😄

taquitos:
@minuscorp what more do you think needs to be done before this PR is in a state to be merged? Is it just the dependency on https://github.com/fastlane/ci/pull/360 ?

minuscorp:
I have to say, that this is slow as 🐢. I don't know why cloning repos are taking so long, I think one of the major issues is that we clone the whole repo instead of just shallow clone it, which would speed up things, I have opened a discussion and proposal in https://github.com/fastlane/ci/issues/371 to tackle this sooner than later, I think this is something we should discuss. Anyway, this works just as intended and designed, but it is slow, sorry :(

taquitos:
Slowness is fine. We're not really concerned with that at this time, we're more focused on making everything work the way we expect, and then we'll spend time tuning the slow parts 😄
We want to make sure we understand what is actually slow so that we can make targeted improvements. We might find out cloning isn't the slowest part.

minuscorp:
Tried to do my best anyway, if you think this and its parent are ready, hit the merge! If not, I'll tackle comments tomorrow morning 😁

app/features/project/project_controller.rb
```diff
@@ -70,77 +72,87 @@ class ProjectController < AuthenticatedControllerBase
...
       }
 
-      # Delete the project
-      FileUtils.rm_rf(repo_path) if File.directory?(File.join(repo_path, ".git"))
-
       erb(:new_project_form, locals: locals, layout: FastlaneCI.default_layout)
     end
 
-    post "#{HOME}/add/*" do |repo_name|
+    post "#{HOME}/add/*/*" do |org, repo_name|
```
KrauseFx: This doesn't consider the host, in our case currently always being `github`, right?

app/features/project/project_controller.rb
```diff
@@ -70,77 +72,87 @@ class ProjectController < AuthenticatedControllerBase
...
       provider_credential = check_and_get_provider_credential(type: FastlaneCI::ProviderCredential::PROVIDER_CREDENTIAL_TYPES[:github])
 
       github_service = FastlaneCI::GitHubService.new(provider_credential: provider_credential)
-      selected_repo = github_service.repos.detect { |repo| repo_name == repo.name }
+      selected_repo = github_service.repos.detect { |repo| org + "/" + repo_name == repo.full_name }
```
KrauseFx: Instead of combining the string and then comparing the result, can we compare the repo name, and the org name directly?

app/features/project/project_controller.rb
```diff
@@ -70,77 +72,87 @@ class ProjectController < AuthenticatedControllerBase
...
-      fastfiles = Dir[File.join(repo_path, "fastlane/Fastfile")]
-      # If not, it might be in a subfolder
-      fastfiles = Dir[File.join(repo_path, "**/fastlane/Fastfile")] if fastfiles.count == 0
-
-      if fastfiles.count > 1
-        logger.error("Ugh, multiple Fastfiles found, we're gonna have to build a selection in the future")
-        # for now, just take the first one
+      fastfile_config = {}
+
+      fastfile.tree.each_key do |key|
```
KrauseFx: I'm not sure on why we need to do this, maybe add a few lines of comment explaining it?

minuscorp: Yeah just added everything super fast in this and didn't add comments, I'll do it in a follow up

app/features/project/project_controller.rb
minuscorp: It considers the current provider credential of the user for getting the code hosting service

app/features/project/project_controller.rb
minuscorp: Have to check if it has that information but yeah😁

PR #366, opened 2018-03-23
Added Overview page with initial Grid stylings

nakhbari:
Added similar Grid stylings to what is done in the Firebase console. This grid styling will great help us be mobile responsive with only now adding tags like `fci-grid` and `fci-cell`.

`fci` is used as the namespace for all stylings in order not to conflict with any other styles. 

I added a shared folder with shared styles. Eventually i'd like to have a library of colors, fonts and paddings in there so that any sub-style sheet can reference the same constants.
#### GIF 
overview title


app/src/app/app.module.ts
```diff
@@ -1,15 +1,19 @@
+import {HttpClientModule} from '@angular/common/http';
 import {NgModule} from '@angular/core';
+import {MatGridListModule} from '@angular/material';
```
snatchev: is this used anywhere?

nakhbari: Ah no that's for another CL. Can remove for this one
Good catch! 🔥

nakhbari: done.

PR #367, opened 2018-03-23
Allow web app to reload inside Angular

nakhbari:
Angular has it's own router which manages the address navigation so if you go to `/` it can reroute you to `/overview` without having to reload all the assets.

However, if you refresh the page Sinatra will not have a route called `/overview` and it will error out. As a workaround to this, I set up a all the data routes under `/data/...` and anything that isn't a data route should return the Angular page.

This is the pattern Angular describes: https://angular.io/guide/deployment#routed-apps-must-fallback-to-indexhtml

TODO: Need to add fallback route for Angular should the route not exist in their router either.

### DIFF
--- a/app/features-json/project_json_controller.rb
+++ b/app/features-json/project_json_controller.rb
@@ -3,7 +3,7 @@ require_relative "../shared/authenticated_controller_base"
 module FastlaneCI
   # Controller for providing all data relating to projects
   class ProjectJSONController < AuthenticatedControllerBase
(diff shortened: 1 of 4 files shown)


nakhbari:
@snatchev I thought about this too. I was thinking we shouldn't load the other controllers in WEB_APP mode. 

I'm trying not to put too much work in the html returning endpoints since i'm expecting to remove them at some point.

PR #370, opened 2018-03-23
Move ng app into web

snatchev:
This PR moves the angular front-end app from `/app/src` to `/web`


taquitos:
Seems reasonable to me but will defer to @nakhbari

PR #372, opened 2018-03-23
Use `detect` over `select ... first`

minuscorp:


app/services/services.rb
```diff
@@ -77,6 +77,8 @@ def self.configuration_git_repo
     end
 
     def self.ci_user
+      require "pry"
```
taquitos: whoopsies 😆

PR #378, opened 2018-03-23
Enable Octokit cache

minuscorp:


app/shared/models/git_repo.rb
```diff
@@ -71,6 +84,7 @@ class << self
     # @param sync_setup_timeout_seconds [Integer] When in sync setup mode, how many seconds to wait until raise an exception. (Defaults to 300)
     # @param callback [proc(GitRepo)] When in async setup mode, the proc to be called with the final GitRepo setup.
     def initialize(git_config: nil, local_folder: nil, provider_credential: nil, async_start: false, sync_setup_timeout_seconds: 300, callback: nil)
+      GitRepo.load_octokit_cache_stack
```
taquitos: I think loading this with every `GitRepo` initialization is probably not needed. It seems like you can load it just 1 time and then every client will use it:
`Octokit.middleware = @stack`

app/shared/models/git_repo.rb
```diff
@@ -60,6 +61,18 @@ class GitRepo
...
+
+      # Loads the octokit cache stack for speed-up calls to github service.
+      # As explained in: https://github.com/octokit/octokit.rb#caching
+      def load_octokit_cache_stack
+        @stack ||= Faraday::RackBuilder.new do |builder|
+          builder.use(Faraday::HttpCache, serializer: Marshal, shared_cache: false)
+          builder.use(Octokit::Response::RaiseError)
+          builder.adapter(Faraday.default_adapter)
+        end
+        return if Octokit.middleware.handlers.include?(Faraday::HttpCache)
```
minuscorp: That's how this is done, memoized cache 😄 @taquitos

taquitos: Duh, yup. Ignore me. 🤦‍♂️

PR #386, opened 2018-03-26
Adds `platform` as a Project attribute

minuscorp:


app/features/project/project_controller.rb
```diff
@@ -149,7 +149,8 @@ class ProjectController < AuthenticatedControllerBase
         name: project_name,
         repo_config: repo_config,
         enabled: true,
-        lane: lane
+        platform: lane.split(" ").last,
+        lane: lane.split(" ").first
```
snatchev: Everything in this PR looks good, but the assumption that a `selected_lane` request parameter has a exactly one space separating the lane name and platform name doesn't seem very strong. What happens if the `selected_lane` does not have space? I am assuming that the lane name is provided by the user. Do we have a validation that it does not allow spaces?

Perhaps we should have a `selected_platform` request parameter?

minuscorp: This is the first step for the implementation. On a follow up PR we don't make that assumption and we actually receive both `platform` and `lane` as separate parameters from the frontend. This is just to not break anything.

snatchev: Got it. Thank you for the explanation.

PR #387, opened 2018-03-26
Favor `yield` over `block.call` for performance improvements.

minuscorp:


snatchev:
I prefer `yield` as well, but for style reasons. I was not aware of any performance differences. Could you point me to some docs regarding this?

minuscorp:
Yup, of course, my bad I didn't add the reference: [here](https://stackoverflow.com/a/1410176)

taquitos:
I prefer explicitly calling out block.call since we used named parameters a lot and it helps with confusion around what block is being executed. So my preference is to not use yield.

taquitos:
I do understand yield is more beautiful and ruby-esq, though, so I don't think y'all are carzy 😄 I just like how understandable code is (if you're not as familiar with ruby) when you use `.call`

minuscorp:
Closing this for now then?

PR #388, opened 2018-03-26
require spec helper and absolute path

snatchev:
`require` with relative path is a bit brittle. Also we should make sure our specs start with `require "spec_helper"` so they can be run in isolation.


viktorasl:
That totally makes sense. Don't know where I've seen that brittle relative path import. Sorry ☮️

PR #394, opened 2018-03-27
Added `typesafe_accessor` util

minuscorp:


snatchev:
Thanks for raising this up for discussion. While I see how this might seem helpful in certain specific cases, e.g. during a serialization/deserialization operation for instance, I think this is should not be used to provide any kind of general type safety.

It doesn't provide real guarantees (this behavior can be bypassed intentionally or un-intentionally) and Ruby is just too dynamic to attempt this holistically. Plus the benefit of strict types generally diminish if they are only checked at runtime.

Thats just my initial reaction. Open to wider discussion :)

minuscorp:
If this doesn't bring any kind of safety at _development_ time, I think, as you said, could be very useful in certain cases. When I was first thinking in why would I implement this into a language like Ruby, I was clearly targeting to models. We have a pretty big code base reliant on persisted models and, one error on setting it, can propagate multiple issues throughout the whole application. I think that models like `Build`, `Project`, `User`, etc. has many attributes that really need to have a certain type in order to properly work well. As we want contributors to bring their ideas and enhancements to the platform, I think we need some kind of _type safety_ as this is done in _fastlane_ through (_placeholder because I cannot find which was the method that did that_).

taquitos:
@minuscorp I think you're remembering ConfigItems: https://github.com/fastlane/fastlane/blob/ff0d98d640b9ba76773ae633a90674cb676d2df9/deliver/lib/deliver/options.rb#L310

(verify block checking type)

I think there are a few places where we really need to be careful about types, but ultimately, type checking should be done very rarely and knowing it could be bypassed on accident. If we do add generic type checking code, people might get a little too comfortable and trust it too much.

minuscorp:
I might not be seeing downsides is people getting used to type safety (talking as a Swift developer), true that maybe getting this tool in a global scope might be too much due to its concrete use cases, maybe it can be encapsulated into a module that make more visible to developers and contributors the idea of: hey, you're using **this** tool right here, you **have** to include it. Making it easily found when used out of its designed scope. Suggestions? Modifications? Closing?
Leave here your comments🙌

taquitos:
I see 💯 why this feels like a great solution. But ultimately, it comes down to the fact that Ruby isn't a statically typed language and that grafting type checkers and including that sort of thing is just very non-rubyesq and ends up causing more subtle issues. Believe me, I'm a C#, Java, Swift, engineer, I totally agree with you about the benefits of type safety, it's just that there are some real strengths that come with dynamic languages and mixing these things tend to cause more problems than solve. It creates a little bit of overhead for people to learn, and it ultimately, the sorts of issues that we're solving by implementing a type checker could potentially be solved by more robust tests. So I would prefer that we add more tests that guarantee good behavior when things are done correctly, and fail fast when they don't, but really leveraging the dynamic nature of ruby instead of relying on types.

minuscorp:
Closing this?

taquitos:
Yeah, let's close this out, thanks for your understanding ❤️

PR #396, opened 2018-03-27
[WIP] Add new proto file to control what we expose via json api

nakhbari:
This should most likely be a placeholder for whatever protos we use in the future. But the idea here is that we separate the models we expose from the ones we pass around in the backend.


taquitos:
As mentioned in slack:

> Heya, I think it would really help out if we put together a design doc around the proto work that is going on so that we all can have the same shared context around what problems we’re solving with protos and how we’re thinking about architecting it. That will also help to share knowledge for people who aren’t familiar with this stuff in the future :thumbsup:

app/features-json/proto/project_summary_proto.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module FastlaneCI
+  # Proto to expose the basic info about a project.
+  class ProjectSummaryProto
+    class << self
+      def object_class
+        FastlaneCI::Project
+      end
+
+      def proto_from(object:)
```
taquitos: I don't know that `proto` is the right word for this. I understand you're thinking about protobufs, but maybe there is some other word we can use that is more familiar to devs who don't know protobuf stuff.

nakhbari: I 100% agree. I don't know what the correct term is. But this is something much simpler and might eventually get replaced by something more like a proto.

taquitos: If this is used solely for the view, I might learn towards `ViewData`

minuscorp:
I'll try to work on the idea that created this PR from tomorrow and see if that can make the trick for you!

app/features-json/proto/project_summary_proto.rb
minuscorp: `ViewModel`? 😇

minuscorp:
I'll try to get this thing in a PR tomorrow, I was a little busy today 😅

PR #397, opened 2018-03-27
Added card for each project

nakhbari:
Later i'm planning on providing the status of the latest build to the UI in order to display it in place of that icon.

I'm also thinking if we can add the lane that is running there as a sub-text item.

#### Screenshot
image


taquitos:
Hey @chemichelle, do you have any thoughts on this design? I think as a proof of concept this is great, but long-term I think cards *might* be an inefficient use of space, but I'm definitely not an expert on this stuff.

KrauseFx:
I was thinking a bit more about the CI setups I worked with in the past - I wonder if a card design would scale well with a higher number of projects?

chemichelle:
+1 to @KrauseFx 's concern about scaling.
@taquitos are we planning to display other info about each project at this level (short description, iOS vs Android, permissions, etc)? If we are, it would make sense to show the projects in grid view - so users can quickly get some extra info about the project and to differentiate them at a glance. If the common use case is that users have lots of projects, it might make more sense to have smaller cards or list view, also considering allowing them to filter/search for projects. 
We could also have a tab which allows users to switch between grid and list views, like what gDrive has.

taquitos:
The most common use case for the dashboard is seeing a list of projects, and then doing one of the following actions: trigger it, edit it, or view the last output (view the last output is most common, then trigger, then edit). 

Generally speaking, people name their projects things like "App Release - iOS" and having meta about the project isn't super useful and just ends up cluttering the list.

If people have a bunch of projects, having a simple search bar that does real-time filtering of projects is even more useful so it trims the list down as you type.

I think having a tab would be really cool, like if people have just a few projects we can show cards with more detailed information. That being said, I think we should focus on the list view for MLP 😄 

My mantra: "Is this enough to be lovable (viable)?" and if the answer is yes, I move on to the next must-have feature for launch 😃

nakhbari:
@taquitos @KrauseFx @chemichelle Awesome feedback everyone. I will go ahead and work on a list view.

I think for MVP we should push forward with all the views (project, builds, settings) and then take a step back and revise how it all fits together. I think this will give us a basis to compare with and a better idea of the most important journeys in the UI.

Let me know what you think.

taquitos:
I think it would be really helpful to create the user stories before moving forward with building ui components. You can see where we're at with our current set here: https://github.com/fastlane/ci/milestone/9
It's really valuable to create the user stories, talk through them, and then move forward with building something. That way, if we don't have designs done, we at least know what the goal of each view is, and if we're accomplishing it. Since you're designing things to be really modular, it means once we do have designs, it should be super easy to update our UI!

minuscorp:
BuddyBuild example dashboard

taquitos:
@minuscorp yeah, we've seen that too. For the average use case, you'll only have 1 or 2 cards there, and it creates an extra step to click into your repo. Instead of being focused on a repo, we're suggesting you think about your projects as jobs you would like to have done. People have just a few jobs for just 1 or two projects. Bundling them all together in one view cuts down on needing to traverse a hierarchy. 

We're expecting a setup that is actually a little more similar to jenkins, where you have a project as the top-most part, which has a self-describing name like `My App Integration Tests` like this:

jenkins-dashboard

nakhbari:
@chemichelle and I spoke offline. While she works on the UX exploration we'll be getting the basic stuff in here so that we can ensure that we have the right data models and communication with backend. 

So let's not get too attached to what it looks like for now, just that there will be a page for projects, project details and build details.

nakhbari:
gonna redo from scratch since the design is different

PR #403, opened 2018-03-28
Fix first server launch trying to clone repos

KrauseFx:


launch.rb
```diff
@@ -113,6 +112,8 @@ def self.register_available_controllers
     end
 
     def self.clone_project_repos
+      return unless Services.onboarding_service.correct_setup?
```
taquitos: Oh man, this looks familiar, I think I broke this. Sorry! 🤦‍♂️

PR #432, opened 2018-03-29
Simplify repo creation inside of build runner

taquitos:


app/workers/check_for_new_commits_on_github_worker.rb
```diff
@@ -27,38 +28,35 @@ def initialize(provider_credential: nil, project: nil)
...
       end
 
-      logger.debug("Creating a build task for commits: #{shas_to_build} from #{self.project.project_name} (#{repo_full_name})")
+      pr_details = new_commit_prs.map { |pr| "#{pr.repo_full_name}:#{pr.branch}:#{pr.current_sha}" }
+      logger.debug("Creating build task(s) for #{self.project.project_name} (#{repo_full_name}): #{pr_details}")
 
-      shas_to_build.each do |sha|
-        self.create_and_queue_build_task(sha: sha)
+      new_commit_prs.each do |pr|
+        git_fork_config = GitForkConfig.new(current_sha: pr.current_sha,
```
taquitos: always create a git fork config because we can use it on non-forks too. Maybe rename this.

app/workers/github_worker_base.rb
```diff
@@ -54,25 +54,18 @@ def thread_id
...
     end
 
-    def create_and_queue_build_task(sha:)
+    def create_and_queue_build_task(sha:, git_fork_config: nil)
       credential = self.provider_credential
       current_project = self.project
       current_sha = sha
       return unless Services.build_runner_service.find_build_runner(project_id: current_project.id, sha: current_sha).nil?
 
-      repo = GitRepo.new(
```
taquitos: don't do a big heavy thing before dispatching to a queue, instead, dispatch the initialization and then do the heavy lifting!

launch.rb
```diff
@@ -198,19 +198,11 @@ def self.run_pending_github_builds(projects: nil, github_service: nil)
                                                   clone_url: matching_open_pr.clone_url)
           end
 
-          repo = GitRepo.new(
```
taquitos: so much deletion 😄

PR #359, opened 2018-03-23
Added GitRepo utility decorators

minuscorp:
This PR is a piece of my older PR, deprecated by now I think.

Adds:

- `pull_before` decorator.
- `commit_after` decorator.

PR #459, opened 2018-03-30
Make use of GitRepoDecorator in ProjectService

minuscorp:
Relates to #359

### DIFF
--- a/app/services/code_hosting/decorators/git_repo_decorator.rb
+++ b/app/services/code_hosting/decorators/git_repo_decorator.rb
@@ -41,7 +41,11 @@ module FastlaneCI
           send(new_name_for_old_function, *args)
         else
           GitRepoDecorator.pull_operation_mutex.synchronize do
-            git_repo&.pull
+            begin
+              git_repo&.pull
+            rescue StandardError => ex
+              logger.error(ex)
+            end
             send(new_name_for_old_function, *args)
           end
         end
--- a/app/services/code_hosting/decorators/git_repo_decorator.rb
+++ b/app/services/code_hosting/decorators/git_repo_decorator.rb
@@ -30,7 +30,7 @@ module FastlaneCI
       end
     end
 
-    def pull_before(func_name, git_repo: Services.configuration_git_repo)
+    def pull_before(func_name, git_repo: FastlaneCI::Services.configuration_git_repo)
       new_name_for_old_function = "#{func_name}_old_pull".to_sym
       alias_method(new_name_for_old_function, func_name)
       define_method(func_name) do |*args|
@@ -54,7 +54,7 @@ module FastlaneCI
       end
     end
 
-    def commit_after(func_name, git_repo: Services.configuration_git_repo)
+    def commit_after(func_name, git_repo: FastlaneCI::Services.configuration_git_repo)
       new_name_for_old_function = "#{func_name}_old_commit".to_sym
       alias_method(new_name_for_old_function, func_name)
       define_method(func_name) do |*args|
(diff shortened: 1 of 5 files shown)


minuscorp:
Don't know if there's interest on this, should I resolve conflicts and move forward with the merge?

minuscorp:
Do I rebase to ease your work @KrauseFx?

app/services/code_hosting/decorators/git_repo_decorator.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      def included(base)
+        base.send(:extend, InstanceMethods)
+      end
+    end
+
+    # Add this as Instance Methods
+    module InstanceMethods
+      def pull_before(func_name, git_repo: GitRepoDecorator.configuration_repo)
+        new_name_for_old_function = "#{func_name}_old_pull_#{SecureRandom.uuid}".to_sym
```
KrauseFx: Sorry for the noob question: isn't there a better way in Ruby to achieve this? Seems kind of hacky and in-elegant

app/services/code_hosting/decorators/git_repo_decorator.rb
```diff
@@ -0,0 +1,100 @@
+require_relative "../../../shared/logging_module"
+require "securerandom"
+
+module FastlaneCI
+  # This module acts as a decorator to be added to any method that potentially will make changes
```
KrauseFx: I love this

app/services/code_hosting/decorators/git_repo_decorator.rb
snatchev: This technique is similar to Rails' `alias_method_chain`. I _think_ we could achieve similar results using `Module#prepend` as mentioned here: https://www.justinweiss.com/articles/rails-5-module-number-prepend-and-the-end-of-alias-method-chain/

snatchev: My personal preference would be to try to use `Module#prepend`. If for whatever reason that is not possible here, I would prefer not to create methods with uuids in them. Something like `pull_before_#{func_name}` is totally reasonable. If that method already exists, perhaps we throw an exception, so we don't overwrite it.

app/services/project_service.rb
```diff
@@ -6,18 +6,21 @@
...
+    include FastlaneCI::GitRepoDecorator
+
     attr_accessor :project_data_source
 
     def initialize(project_data_source: nil)
       unless project_data_source.nil?
         raise "project_data_source must be descendant of #{ProjectDataSource.name}" unless project_data_source.class <= ProjectDataSource
       end
-
+      GitRepoDecorator.configuration_repository(Services.configuration_git_repo)
```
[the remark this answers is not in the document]
minuscorp: So the issue here is that the decorators at loaded at `require` level, so there are no instances of any repo at that point, that's why I had to fallback to provide the default parameter at initialization time, so the parameter actually exists when this call is being made.

app/services/code_hosting/decorators/git_repo_decorator.rb
minuscorp: AFAIK, prepend just changes the call method chain, but for me to do the `prepend`, the module would need to have implemented the method that is being prepended like:
```ruby
module Decorator
   def method_prepended
      puts "I'm prepended"
   end
end
class Service
   prepend Decorator

   def method_prepended
      puts "I'm the last of the chain"
   end
end

# So this will end in:
service = Service.new
service.method_prepended # => "I'm prepended" "I'm the last of the chain"
```

So `Decorator` *needs* to know the name of the method that will be prepended before it actually is defined by the class it will prepend. I don't know if I'm explaining myself right 😅

snatchev:
I am going to close this PR at this time. I believe that the solution to the problem can be addressed with a more naive and procedural approach. The proposed solutions here are certainly interesting, useful and probably "correct" but ultimately I believe there is a maintenance cost that is a bit too high.

We might have to chalk this up to the "Worse is Better" philosophy.

PR #460, opened 2018-03-30
First approach to ViewModels

minuscorp:
Relates to https://github.com/fastlane/ci/pull/396 and adds the base architecture design.


app/features-json/view_models/view_model.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      # Creates the ViewModel (in shape of a Hash) from a given object of a certain class type.
+      # Defined by the base_model method.
+      # @params [Any] object
+      # @return [Hash] object dictionary representation for the ViewModel.
+      def viewmodel_from!(object:)
+        raise "Override base_model(model) in order to use the ViewModel mixin." if @base_model.nil?
+        raise "Incorrect object type. Expected #{@base_model}, got #{object.class}" unless object.kind_of?(@base_model)
+        if object.respond_to?(:to_object_dictionary)
+          return object.to_object_dictionary(ignore_instance_variables: @base_model.instance_variables - included_attributes)
```
nakhbari: Noob ruby question. Some of our models have methods with important data instead of just the attributes. For example Project Model has a method to get the builds associated with that project. How do I get that data in this view data object?

minuscorp: hmmmmm in this case we should split into two `ViewModel`s, as we try to maintain the relationship within its `Model`. For a given `Build` we make a `BuildViewModel` for example.

minuscorp: The key element of this `ViewModel`s is that being attached to a single model (that could be a combination of other models, we're not tied to that) so it's easy to take a look and not see nested models. I do not like a lot having too deep JSON. If Angular allows it, it's preferable to have several JSON (`ViewModel`s) to prevent future issues. But if this is an Angular requirement, we can maybe make this so instead to tie the `ViewModel` to a single `Model` can tie it to several ones and instead of nest it, provide them at the same JSON level.

nakhbari: I think you're right we can allow for multiple smaller requests instead of bundling them up.

PR #461, opened 2018-03-31
Allow the user to add at least one trigger when creating a Project

### LINKED ISSUE #455: Add new project which only allows a release style trigger (no manual, only commit)

minuscorp:
- [x] Add a minimal interface to add a single `JobTrigger` when creating a new `Project`.
- [x] Disable `Trigger Job` button when no `ManualJobTrigger` is not found in the `Project`

Fixes #455

### DIFF
--- a/app/features/project/project_controller.rb
+++ b/app/features/project/project_controller.rb
@@ -159,19 +159,6 @@ module FastlaneCI
         trigger = FastlaneCI::NightlyJobTrigger.new(branch: branch, hour: hour.to_i, minute: minute.to_i)
       end
 
-      dir = Dir.mktmpdir
-      # Do this so we trigger the clone of the repo.
-      # TODO: Do this wherever it should be done, as we must redirect
-      # to the project details only when this task is finished.
-      repo = GitRepo.new(
-        git_config: repo_config,
-        provider_credential: provider_credential,
-        local_folder: dir,
-        async_start: false
-      )
-
-      repo.checkout_branch(branch: branch)
-
       # We now have enough information to create the new project.
       # TODO: add job_triggers here
       # We shouldn't be blocking manual trigger builds
@@ -187,6 +174,18 @@ module FastlaneCI
       )
 
       if !project.nil?
+        # Do this so we trigger the clone of the repo.
+        # TODO: Do this wherever it should be done, as we must redirect
+        # to the project details only when this task is finished.
+        repo = GitRepo.new(
+          git_config: repo_config,
+          provider_credential: provider_credential,
+          local_folder: project.local_repo_path,
+          async_start: false
+        )
+
+        repo.checkout_branch(branch: branch, use_global_git_mutex: false)
+
         redirect("#{HOME}/#{project.id}")
       else
         raise "Project couldn't be created"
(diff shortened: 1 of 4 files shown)


app/features/project/project_controller.rb
```diff
@@ -146,19 +146,18 @@ class ProjectController < AuthenticatedControllerBase
...
-        async_start: false
-      )
-
-      repo.checkout_branch(branch: branch)
+      trigger_type = params["selected_trigger"]
+      hour = params["hour"]
+      minute = params["minute"]
+
+      case trigger_type
+      when "commit"
```
taquitos: Let's use the enum:
```
case trigger_type
when FastlaneCI::JobTrigger::TRIGGER_TYPE[:commit]
  ...
when FastlaneCI::JobTrigger::TRIGGER_TYPE[:manual]
  ...
when FastlaneCI::JobTrigger::TRIGGER_TYPE[:nightly]
  ...
end
```

app/features/project/project_controller.rb
```diff
@@ -171,10 +170,22 @@ class ProjectController < AuthenticatedControllerBase
         platform: lane.split(" ").last,
         lane: lane.split(" ").first,
         # TODO: Until we make a proper interface to attach JobTriggers to a Project, let's add a manual one for the selected branch.
-        job_triggers: [FastlaneCI::ManualJobTrigger.new(branch: branch)]
+        job_triggers: (trigger.nil? ? [] : [trigger])
```
taquitos: I think trigger should never be nil, so maybe we should `raise` an exception?

app/features/project/project_controller.rb
```diff
@@ -171,10 +170,22 @@ class ProjectController < AuthenticatedControllerBase
...
+        # TODO: Do this wherever it should be done, as we must redirect
+        # to the project details only when this task is finished.
+        repo = GitRepo.new(
+          git_config: repo_config,
+          provider_credential: provider_credential,
+          local_folder: project.local_repo_path,
+          async_start: false
+        )
+
+        repo.checkout_branch(branch: branch, use_global_git_mutex: false)
```
taquitos: fyi you almost always want to use `use_global_git_mutex: true`
The only time you don't want to use it is when you're already using the `GitRepo.git_action_queue`, since the serial queue isn't reentrant 👍

app/features/project/views/new_project_form.erb
```diff
@@ -100,6 +141,32 @@
       <select name="selected_lane", id="selected_lane" onchange="check_lane()">
       </select>
       <br />
+      <h5 id="trigger">Select trigger</h5>
+      <div id="trigger_selector">
+        <select name="selected_trigger", id="selected_trigger", onchange="check_trigger(this)">
+          <option disabled selected value>Select a trigger</option>
+          <option value="commit">Commit</option>
```
taquitos: We could populate these values on the server-side instead of hard coding, but I don't think you should prioritize that. Just in the future, know that's an option.

minuscorp:
Ready to go!

taquitos:
This is really great. cc @KrauseFx

KrauseFx:
Resolved merge conflicts

Issue #400, opened 2018-03-28
Favor `self.var = _` over `@var = _` and attr_reader


taquitos:
Using proper scopes for our attributes would be 💯. We have a better idea of what things we need access to now that CI has been moving along.
We had a PR open for this, but it got really out of sync with master so we closed it: https://github.com/fastlane/ci/pull/344

AndrewMcBurney:
@KrauseFx, why did you unassign me from this issue?

AndrewMcBurney:
Closed with #462

PR #462, opened 2018-04-01
Unify attributes

AndrewMcBurney:
#400

This PR will bring about substantial changes to large parts of the codebase. I believe it is necessary to reduce code smells and prevent technical debt from accumulating.

### Rationale

The changes follow Ruby idioms and best practices, described in Sandi Metz's [Practical Object-Oriented Design in Ruby](http://www.poodr.com/), [[1]](https://github.com/skmetz/poodr), Shopify's [Ruby style guide](http://shopify.github.io/ruby-style-guide/), AirBnb's [Ruby style guide](https://github.com/airbnb/ruby#redundant-self), GitHub's [Ruby style Guide](https://github.com/github/rubocop-github/blob/master/STYLEGUIDE.md), and the community [Ruby style guide](https://github.com/bbatsov/ruby-style-guide#no-self-unless-required).

### Propositions

**Proposition 1: Favour `@var_name = var_name` over `self.var_name = var_name`:**

I'm not sure we want a unified `self.var_name = var_name` over `@var_name = var_name`, since the `self.var_name = _` requires we have an `attr_accessor` defined.

For example, take the following class which illustrates what we're currently doing:

```ruby
# temp.rb

class Person
  def initialize(greeting)
    self.greeting = greeting
  end

  def speak
    puts self.greeting
  end
end

Person.new("Hello").speak
```
> Doesn't compile since we need to define an `attr_accessor`.

Running `ruby temp.rb` will yield the following error:

```
temp.rb:3:in `initialize': undefined method `greeting=' for #<Person:0x007fdbb40137d0> (NoMethodError)
        from temp.rb:11:in `new'
        from temp.rb:11:in `<main>'
```

This code will not run, since we must define an `attr_accessor` on `greeting` to use `self.greeting = greeting`:

```ruby
class Person
  # NOTE: this will expose both a setter AND getter for `greeting`, which may be
  # used by any class which holds a reference to it. Hence, any class which has
  # a person may mutate their `greeting`.
  #
  # @return [String]
  attr_accessor :greeting

  def initialize(greeting)
    self.greeting = greeting
  end

  def speak
    puts greeting
  end
end

Person.new("Hello").speak
```
> Compiles, but exposes a public setter for anything which "has a" `Person` to mutate

As stated in the comment above, any class which holds a reference to a `Person` instance may now mutate the `Person`'s `greeting` instance variable.

Here's an example of another class with reference to `Person` changing `Person`'s `greeting`:

```ruby
class Person
  # NOTE: this will expose both a setter AND getter for `greeting`, which may be
  # used by any class which holds a reference to it. Hence, any class which has
  # a person may mutate their `greeting`.
  #
  # @return [String]
  attr_accessor :greeting

  def initialize(greeting)
    self.greeting = greeting
  end

  def speak
    puts greeting
  end
end

class Translator
  # @return [Arary[Person]]
  attr_accessor :people

  # @param [Arary[Person]] people
  def initialize(people)
    self.people = people
  end

  def change_greetings
    people.each { |person| person.greeting = "Bonjour!" }
  end
end

person1 = Person.new("Hello!")
person2 = Person.new("Hi!")
person3 = Person.new("Aloha!")

person1.speak # => Hello!
person2.speak # => Hi!
person3.speak # => Aloha!

# Mutate all of the greetings
Translator.new([person1, person2, person3]).change_greetings

person1.speak # => Bonjour!
person2.speak # => Bonjour!
person3.speak # => Bonjour!
```

This makes debugging difficult because any arbitrary class which has an instance of the `Person` class may now mutate the data in the instance.

While this could be desired behaviour in some situations, I don't believe it should be what we default to in our class constructors.

**Proposition 2: Apply the default Rubocop `RedundantSelf` settings**

My opinion is to get rid of redundant `self` where possible ([the rubocop default](http://www.rubydoc.info/gems/rubocop/0.16.0/Rubocop/Cop/Style/RedundantSelf)). While in other languages it is common to use `self.var_name` when referring to an instance variable, in Ruby this is not the convention.

My proposition is that we only use `self` on an instance variable when:

1. Defining a class method: `def self.some_method`.
2. The *left-hand side* when calling an assignment method.
3. Referencing the current instance's class: `self.class`.

This way, it's explicit when mutating an instance variable since you only use `self` for instance variables when you're explicitly mutating something, which has a corresponding `attr_accesor`.


AndrewMcBurney:
I realize this is a pretty big PR, and a lot to take in at once. I can break this up into multiple PRs if it helps to expedite the review process or make it a little easier.

app/features/build_runner/fastlane_build_runner.rb
```diff
@@ -58,11 +60,11 @@ def run(completion_block: nil, &block)
...
       Fastlane.load_actions
 
-      fast_file_path = FastlaneCI::FastfileFinder.find_fastfile_in_repo(repo: self.repo)
+      fast_file_path = FastlaneCI::FastfileFinder.find_fastfile_in_repo(repo: repo)
       if fast_file_path.nil? || !File.exist?(fast_file_path)
-        logger.info("unable to start fastlane run lane: #{self.lane} platform: #{self.platform}, params: #{self.parameters}, no Fastfile for commit")
+        logger.info("unable to start fastlane run lane: #{lane} platform: #{platform}, params: #{parameters}, no Fastfile for commit")
         self.current_build.status = :missing_fastfile
-        self.current_build.description = "We're unable to start fastlane run lane: #{self.lane} platform: #{self.platform}, params: #{self.parameters}, because no Fastfile existed at the time the commit was made"
+        self.current_build.description = "We're unable to start fastlane run lane: #{lane} platform: #{platform}, params: #{parameters}, because no Fastfile existed at the time the commit was made"
```
taquitos: .description should be `attr_accessor` too

taquitos:
Ok, so next time you resolve the conflicts, I would recommend you test the following:
- startup (make sure server starts successfully)
- Configure CI for a project and make a PR with commits while CI is running to sure the ci system picks it up and can build it successfully
- kill the system, then create a PR with some changes, then start the system up and make sure they are picked up and build (testing that you can kill CI, and on restart it picks up any new work it might have missed while it was down)

This should be enough to flex all the code changes you made (which is what I was doing when you and I were going back and forth this morning) 👍

app/shared/models/artifact.rb
```diff
@@ -6,27 +6,27 @@ class Artifact
...
+    attr_reader :id
 
     # type is just a String that refers to the kind of data that the Artifact represents (i.e., ipa, apk, log, etc.)
     # @return [String] The type or kind of artifact.
-    attr_accessor :type
+    attr_reader :type
 
     # @return [String] The reference to the artifact to a certain ArtifactProvider
-    attr_accessor :reference
+    attr_reader :reference
```
KrauseFx: Oh so this silently broke artifact support, see https://github.com/fastlane/ci/pull/508, as it's being accessed in sub-classes. Will this change also affect other parts of the code base, where we didn't consider sub classes accessing the attribute? In this case it's https://github.com/fastlane/ci/blob/master/app/shared/models/gcp_artifact_provider.rb#L76

cc @AndrewMcBurney @taquitos

taquitos: Yeah, we went through many rounds of testing and launching, but that doesn't mean we caught it all. I think the best course of action is to flex the system as much as we can.

AndrewMcBurney: Yeah, it affects the provider one too 😳. I'll write a test suite for the artifact support so we won't run into problems like this is the future.

Issue #285, opened 2018-03-13
Enable code coverage reports once we have ci for fastlane.ci


KrauseFx:
Once we have a working fastlane.ci for fastlane.ci, we should enable coveralls.io or similar to see code coverage trends and ideally automatically post a comment if the coverage declined to encourage contributors to add new tests 👍

taquitos:
This is the best idea ever 🍬🐐!!

Issue #457, opened 2018-03-30
Support for distributed builds


KrauseFx:
- Have a way of having a master node and multiple worker nodes
- The worker nodes ask for tasks from the master node
- This allows teams to scale up `fastlane.ci`

This task includes transferring of all required metadata for a build, meaning the git repo details, branch to run, lane to run, and the environment variables set for the build (https://github.com/fastlane/ci/pull/465), and the environment variables set by the user

KrauseFx:
Blocked for now, please don't grab this task

minuscorp:
BTW, just leaving this here: https://github.com/mperham/sidekiq 
I have kind of strong experience in distributed task queues and so on, so I want you to know that you can rely on me 😊

taquitos:
For distributed builds, we're probably going to be working with a solution that doesn't require installing/maintaining another piece of server infrastructure like separate queues. The goal of distributed builds is to have multiple macs with `fastlane.ci` installed that find each other, and then configure themselves to be headless nodes in a cluster with 1 mac being the server and delegating out tasks.
Goals:
-  Keep setup and maintenance dead-simple
-  No extra software packages needed
-  Automatic configuration
-  One server node to manage ci-config repo prevents races
-  One server node to manage artifacts prevents races
-  One server node to rule them all

AndrewMcBurney:
> - One server node to rule them all

![image](https://user-images.githubusercontent.com/8942499/38755798-1914e89a-3f35-11e8-9eb2-f040287e23d0.png)

minuscorp:
I can't figure out a messaging system that does not rely on third-party libraries (unless we actually make one ofc 🤣 ), I meant as for operating system level. Maybe a TCP socket-based locking system could be a nice feature to make use of.
Just throwing out ideas that comes into my mind 🙃

taquitos:
I'm ok with third-party libs, we just don't want to install a separate, stand-alone piece of server infra, which is what you generally have to do in traditional queues.

Really though, I think we can go very simple. My first thoughts are: [bonjour](https://developer.apple.com/bonjour/) + [grpc](https://grpc.io/docs/tutorials/basic/ruby.html).

minuscorp:
Seems legit too. Maybe easier that what I've proposed 🤣

Ashraf-Ali-aa:
[...]

PR #465, opened 2018-04-02
Set needed CI specific ENV variables per build

KrauseFx:
Needed for https://github.com/fastlane/ci/issues/285

### DIFF
--- a/app/features/build_runner/build_runner.rb
+++ b/app/features/build_runner/build_runner.rb
@@ -44,6 +44,9 @@ module FastlaneCI
     # Work queue where builds should be run
     attr_accessor :work_queue
 
+    # Array of env variables that were set, that we need to unset after the run
+    attr_accessor :environment_variables_set
+
     def initialize(project:, sha:, github_service:, work_queue:, git_fork_config: nil)
       # Setting the variables directly (only having `attr_reader`) as they're immutable
       # Once you define a FastlaneBuildRunner, you shouldn't be able to modify them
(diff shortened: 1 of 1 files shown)


app/features/build_runner/build_runner.rb
```diff
@@ -129,6 +132,52 @@ def checkout_sha
...
+        JOB_NAME: self.project.project_name,
+        WORKSPACE: self.project.local_repo_path,
+        GIT_URL: self.repo.git_config.git_url,
+        GIT_SHA: self.current_build.sha,
+        BUILD_URL: "https://fastlane.ci", # TODO: actually build the URL, we don't know our own host, right?
+        CI_NAME: "fastlane.ci"
+      }
+
+      if self.git_fork_config && self.git_fork_config.branch.to_s.length > 0
+        env_mapping[:GIT_BRANCH] = self.git_fork_config.branch # TODO: does this work?
```
taquitos: it's tricky because the git branch might be a branch on a fork, soooo I dunno.

app/features/build_runner/build_runner.rb
```diff
@@ -129,6 +132,52 @@ def checkout_sha
...
+      env_mapping.each do |key, value|
+        set_build_specific_env_variable(key: key, value: value)
+      end
+
+      # TODO: to add potentially
+      # - BUILD_ID
+      # - BUILD_TAG
+    end
+
+    def set_build_specific_env_variable(key:, value:)
```
taquitos: this is dangerous because we're setting the environment for all builds that could be running. Currently it's not an issue, but if we ever do concurrent builds, this will cause race conditions.

app/features/build_runner/build_runner.rb
```diff
@@ -139,6 +188,15 @@ def reset_repo_state
     def post_run_action
       logger.debug("Finished running #{self.project.project_name} for #{self.sha}")
       self.reset_repo_state
+
+      self.unset_build_specific_environment_variables
```
taquitos: let's make this the first action for `post_run_action`, in case we have an exception. Also we might want to: #todo: make `post_run_action` run no matter what (like an ensure block). Or provide `post_run_ensure`... thoughts?

app/features/build_runner/build_runner.rb
KrauseFx: This is how Jenkins handles it

```
GIT_BRANCH - Name of the remote repository (defaults to origin), followed by name of the branch currently being used, e.g. "origin/master" or "origin/foo"
```

I think that makes sense for now to use `origin/[branch_name]`, and then we can think more about how this works for forks

taquitos: just saying `here is the branch` is not enough to know how to check it out if it's on a fork. I'm sure Jenkins handles this by updating the origin or something.

app/features/build_runner/build_runner.rb
KrauseFx: Cool, I see a few potential solutions:
1) If we use a child process, we can safely set the ENV variables just for the child process, with no way of the ENV variable surfacing up to the fastlane.ci main thread. This works very easily when using shell commands (which we don't)
2) The current implementation should work fine since our assumption is that one macOS hosts runs only one fastlane instance at the same time from what I understand. As soon as we implement distributed building (https://github.com/fastlane/ci/issues/457) we need a way to transfer the build metadata (git repo, branch, lane and build environment variables) anyway, so it's something we're gonna have to think about & implement once we take on #457 

Does that sound correct?

app/features/build_runner/build_runner.rb
taquitos: Ahh yeah, so maybe you should include the actual `clone_url` or something

app/features/build_runner/build_runner.rb
KrauseFx: Oh yes, I didn't know we don't guarantee running it yet, since I thought we always want to run `reset_repo_state`. I'll create an issue for it 👍

PR #471, opened 2018-04-02
Cache successful repo state checks

taquitos:


KrauseFx:
so punny, just like @jasdev

PR #485, opened 2018-04-02
Don't list details about the repos without access

KrauseFx:
This could potentially cause "privacy" issues, as a developer you might not have access to all repos. Instead let the user know how to be added to more projects, currently by being added to a given repo on GitHub

### DIFF
--- a/app/features/dashboard/views/dashboard.erb
+++ b/app/features/dashboard/views/dashboard.erb
@@ -34,7 +34,7 @@
 
     <% if projects_without_access.count > 0 %>
       <h3>There are some projects you don't have access to</h3>
-      <p>To gain access to all projects, reach out to your GitHub org admin. They'll have to give you write access to the repo</p>
+      <p>to gain access to any repos you are missing from your organization, reach out to your GitHub org admin. They'll have to give you write access to the repo</p>
       <br />
     <% end %>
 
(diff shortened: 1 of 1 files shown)


app/features/dashboard/views/dashboard.erb
```diff
@@ -32,26 +32,11 @@
...
-          </tr>
-        <% end %>
-      </tbody>
-    </table>
-    <p>To get access to those projects, make sure your GitHub account has access to them</p>
-
-    <br />
+    <% if projects_without_access.count > 0 %>
+      <h3>There are some projects you don't have access to</h3>
+      <p>To gain access to all projects, reach out to your GitHub org admin. They'll have to give you write access to the repo</p>
```
taquitos: You need "write" access to see repos? TIL

taquitos: Maybe change the message to: "to gain access to any repos you are missing from your organization, reach out..."

KrauseFx:
Good point, fixed

app/features/dashboard/views/dashboard.erb
KrauseFx: Mh, I thought that's the case, we should find out at some point 👍

PR #490, opened 2018-04-02
Speed up boot

taquitos:
- Dispatch run_pending_github_builds and enqueue_builds_for_open_github_prs_with_no_status to queue on startup

### DIFF
--- a/launch.rb
+++ b/launch.rb
@@ -16,6 +16,8 @@ module FastlaneCI
     end
 
     def self.take_off
+      @launch_queue = TaskQueue::TaskQueue.new(name: "fastlane.ci launch queue")
+
       require_fastlane_ci
       verify_app_built
       verify_dependencies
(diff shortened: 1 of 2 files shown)


launch.rb
```diff
@@ -133,8 +135,15 @@ def self.restart_any_pending_work
       github_projects = Services.config_service.projects(provider_credential: Services.provider_credential)
       github_service = FastlaneCI::GitHubService.new(provider_credential: Services.provider_credential)
 
-      run_pending_github_builds(projects: github_projects, github_service: github_service)
-      enqueue_builds_for_open_github_prs_with_no_status(projects: github_projects, github_service: github_service)
+      restart_pending_builds_task = TaskQueue::Task.new(work_block: proc {
+        run_pending_github_builds(projects: github_projects, github_service: github_service)
+      })
+      @launch_queue.add_task_async(task: restart_pending_builds_task)
```
KrauseFx: Can we access using `self.something`?

taquitos: Ahh, I'd rather not, since I don't want to expose this queue externally. Is there a way to do that so it isn't exposed?

KrauseFx: Got it

Issue #491, opened 2018-04-02
Render basic artifacts as part of build details


KrauseFx:
This task requires loading of historic builds from the JSON file + build artifacts

KrauseFx:
Working on this now

Issue #312, opened 2018-03-16
Implement loading of previous log outputs


KrauseFx:
Related https://github.com/fastlane/ci/pull/313

Right now we're just loading in-progress build output, but never the build output from previous runs. This task requires us to access the build artifacts (in particular the `.log` file) and show it to the user

minuscorp:
This is something I could be able to do _easily_, as I implemented the whole Artifact thingy

Issue #497, opened 2018-04-02
Reload build page once the build is complete to render all the build artifacts


KrauseFx:

KrauseFx:
Fixed with https://github.com/fastlane/ci/pull/498

Issue #495, opened 2018-04-02
Loading the build output from a local path doesn't properly render colored output


KrauseFx:
![image](https://user-images.githubusercontent.com/869950/38192694-b0b2ebf4-36a0-11e8-9fb0-38de3eaa7f99.png)

But also, this made me realize, we're not persisting the `FastlaneCI::BuildRunnerOutputRow` objects, which we use to generate HTML code. We'll have to think if we can justify doing that for longer builds, or if we just store and load the raw output for the time being, and close the bridge when we come to it.

Either way, adding support for colors might be worthwhile

PR #498, opened 2018-04-02
Add rendering of existing build output & rendering of all artifacts

KrauseFx:
screenshot 2018-04-02 06 13 31

---

After server restart

screenshot 2018-04-02 06 15 00

Created https://github.com/fastlane/ci/issues/495 as a result

---

Added more build information

screen shot 2018-04-03 at 2 24 52 pm

---

Fixes https://github.com/fastlane/ci/issues/491
Fixes https://github.com/fastlane/ci/issues/312


app/features/build/build_controller.rb
```diff
@@ -21,17 +21,28 @@ class BuildController < AuthenticatedControllerBase
...
+      else
+        # `current_build_runner` is only defined if the build was just run a while back
+        # if the server was restarted, we're gonna end here in this code block
+        build_log_artifact = build.artifacts.find do |current_artifact|
+          # We can improve the detection in the future, to actually mark an artifact as "default output"
+          current_artifact.type == "log" && current_artifact.reference.end_with?("fastlane.log")
+        end
+
+        if build_log_artifact
+          existing_rows = File.read(build_log_artifact.reference).gsub("\n", "<br />")
```
minuscorp: Please use https://github.com/fastlane/ci/blob/61e42317bc8e3c3104406368b517e8691f4ed7b6/app/shared/models/artifact_provider.rb#L26 instead of `reference`

app/features/build/views/build.erb
```diff
@@ -34,11 +34,34 @@
...
+          <th>Artifact type</th>
+          <th>Path</th>
+        </tr>
+      </tr>
+    </thead>
+    <tbody>
+      <% build.artifacts.each do |current_artifact| %>
+        <tr>
+          <td><%= current_artifact.type %></td>
+          <td><%= current_artifact.reference %></td>
```
minuscorp: Same, use `retrieve!` method if possible, maybe I should limit the readability scope of the `reference` attribute

app/features/build/build_controller.rb
KrauseFx: Ah nice, will update tomorrow 👍

app/features/build/views/build.erb
KrauseFx: Yep

app/features/build/views/build.erb
```diff
@@ -34,11 +34,34 @@
...
+  </table>
+
+  <hr />
+<% end %>
 
 
 <code id="log">
-  <% if existing_rows.count > 0 %>
-    <%= existing_rows.join("\n") %>
+  <% if existing_rows.to_s.length > 0 %>
```
AndrewMcBurney: Could we use [`positive?`](https://ruby-doc.org/core-2.3.0/Numeric.html#method-i-positive-3F) instead, please? I feel it's a bit more idiomatic ruby.

```ruby
if existing_rows.to_s.length.positive?
  # ...
end
```

app/features/build/views/build.erb
```diff
@@ -56,7 +79,7 @@
     uri += "/?&project=<%= project.id %>&build=<%= build.number %>"; // we add that extra & to make the Ruby URI parser work
     var ws = new WebSocket(uri);
     console.log("Listening to web socket connection now for URL " + uri + "...")
-    var firstMessageReceived = <%= existing_rows.count == 0 ? false : true %>;
+    var firstMessageReceived = <%= existing_rows.to_s.length == 0 ? false : true %>;
```
AndrewMcBurney: Could we use [`.zero?`](https://ruby-doc.org/core-1.9.3/Numeric.html#method-i-zero-3F) instead, negate and remove the ternary, please?

```ruby
!existing_rows.to_s.length.zero?
```

app/features/build/views/build.erb
KrauseFx: In this case, `existing_rows` isn't an array, but a string 👍 So we can't use `positive?`, which AFAIK would only work in combination with `.count`

app/features/build/views/build.erb
KrauseFx: Same as https://github.com/fastlane/ci/pull/498#discussion_r178723810

taquitos: @KrauseFx:
```
1.zero?
=> false
```

taquitos: it works on `Fixnums`

app/features/build/views/build.erb
taquitos: @KrauseFx 
```
irb(main):004:0> 0.positive?
=> false
irb(main):005:0> 1.positive?
=> true
irb(main):
```
I think we can use it 👍

app/features/build/views/build.erb
KrauseFx: Ah I misread the comment. I'm ok with keeping `length == 0`

app/features/build/views/build.erb
```diff
@@ -81,6 +81,8 @@
   Back to Project
 </a>
 
+<% unless build_complete %>
```
KrauseFx: For builds that are already complete, we don't build up a socket connection any more 👍

KrauseFx:
Aaaand tests are green

KrauseFx:
Fixes https://github.com/fastlane/ci/issues/497

app/shared/models/artifact_provider.rb
```diff
@@ -13,7 +13,7 @@ class ArtifactProvider
     # @param [Build] build, the build which generated the artifact.
     # @param [Project] project, the project which generated the build.
     # @return [Artifact] the resulting Artifact after being stored by the ArtifactProvider.
-    def store!(artifact: nil, build: nil, project: nil)
+    def store!(artifact:, build:, project:)
```
taquitos: yessss

app/features/build_runner/build_runner.rb
```diff
@@ -117,11 +117,13 @@ def complete_run(start_time:, artifact_paths: [])
       save_build_status!
     ensure
       # Make sure to notify the listeners that the build is over
-      new_row(FastlaneCI::BuildRunnerOutputRow.new(
-                type: :last_message,
-                message: nil,
-                time: Time.now
-      ))
+      new_row(
+        FastlaneCI::BuildRunnerOutputRow.new(
```
taquitos: yeah that looks better to me

KrauseFx:
Wohooo 🎉

PR #502, opened 2018-04-02
Check the uniqueness of the Project's name before creating it.

### LINKED ISSUE #501: Validate the Project's name uniqueness before creating the project.
So this prevents lots of ugly errors being raised at the end of the `New Project` flow.

Just adding the `is-invalid` class to the `<input>` tag is enough!

minuscorp:
Fixes #501


app/features/project/project_controller.rb
```diff
@@ -74,6 +74,26 @@ class ProjectController < AuthenticatedControllerBase
       erb(:new_project, locals: locals, layout: FastlaneCI.default_layout)
     end
 
+    get "#{HOME}/*/valid" do
+      content_type :json
+
+      project_name = params[:splat].first
+
+      if !project_name.nil?
+        if Services.project_service.project(name: project_name).nil?
```
taquitos: this is case-sensitive, maybe we should make it case-insensitive?

KrauseFx:
Nice! Let's make sure to not spend too much of our time on front-end related code, as @nakhbari is working on an Angular based front-end, meaning we're gonna have to re-write those lines 👍

minuscorp:
This was just to prevent crash you got when you added a project that had the same name 😄

Issue #243, opened 2018-03-08
[Proposal] Don't ask for the user's email address, it should be available via the GitHub API using the personal access token


KrauseFx:
^ title says it all

taquitos:
People might want to use a different address than their github address. 
**Example:** if I were using CI at google, I'd want to use my google email, but still use my gmail address for the token.

**Counter-proposal:** pre-populate the email field with the github email and allow the user to edit it.

KrauseFx:
Yes, good point 👍

PR #503, opened 2018-04-02
Automatically fill the email on the login with the API token provided.

minuscorp:
Fixes https://github.com/fastlane/ci/issues/243


app/features/login/views/login_fastlane_ci.erb
```diff
@@ -15,7 +15,7 @@
           <hr />
 
           <div class="mdl-textfield mdl-js-textfield mdl-textfield--floating-label">
-            <input class="mdl-textfield__input" type="text" name="email" id="email">
```
KrauseFx: Is that the only field we have to auto-fill? I think we ask for 2 API tokens, one for the bot, one for the user, right? I guess we could apply this for both

app/features/login/login_controller.rb
```diff
@@ -33,8 +36,15 @@ class LoginController < ControllerBase
 
     # Login with fastlane.ci credentials
     get "#{HOME}/ci_login" do
+
```
KrauseFx: This space shouldn't be here

app/features/login/views/login_fastlane_ci.erb
minuscorp: I'm not aware of more than the environment variable with the initial clone token to be used after the onboarding

app/features/login/login_controller.rb
```diff
@@ -33,8 +36,14 @@ class LoginController < ControllerBase
 
     # Login with fastlane.ci credentials
     get "#{HOME}/ci_login" do
+      client = Octokit::Client.new(access_token: FastlaneCI.env.clone_user_api_token)
```
KrauseFx: I just realized we're accessing Octokit in a controller, is @taquitos ok with this?

Issue #329, opened 2018-03-19
If git repo is unreachable, exception is thrown and kills the server


taquitos:
Whoops.

```
[2018-03-18 22:03:29] DEBUG CheckForNewCommitsOnGithubWorker:  Checking for new commits in All rubocops pass
Thread terminated with exception: git '--git-dir=/Users/jliebowitz/.fastlane/ci/86286442-360f-4b9f-8e2b-d60cd0f257e7/.git' '--work-tree=/Users/jliebowitz/.fastlane/ci/86286442-360f-4b9f-8e2b-d60cd0f257e7' fetch 'origin'  2>&1:fatal: unable to access 'https://github.com/fastlane/ci/': Could not resolve host: github.com
  /Users/jliebowitz/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/git-1.3.0/lib/git/lib.rb:937:in `command'
  /Users/jliebowitz/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/git-1.3.0/lib/git/lib.rb:731:in `fetch'
  /Users/jliebowitz/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/git-1.3.0/lib/git/base.rb:317:in `fetch'
  /Users/jliebowitz/dev/ci/shared/models/git_repo.rb:360:in `block in fetch'
```

minuscorp:
I think this does not happen anymore on #283 but should be tested😁

taquitos:
I guess it's time to dive into #283 
BRB, packing a lunch and a gallon of water to take with me.

minuscorp:
Take some survival materials too. You might need to camp in the office 🤣

taquitos:
😎

PR #504, opened 2018-04-03
Utilize NotificationService when we encounter git problems

taquitos:
- fix for #329 (If git repo is unreachable, exception is thrown and kills the server)
- Common exception handling for all git things
- Handle connection issues
- Raise all errors we don't explicitly handle
- Update notification service for extra non-user-fiendly notification details (like exception message)


app/shared/models/git_repo.rb
```diff
@@ -106,32 +118,55 @@ def initialize(git_config: nil, local_folder: nil, provider_credential: nil, asy
...
         sleep(2)
         now = Time.now.utc
       end
 
       raise "Unable to start git repo #{git_config.git_url} in #{sync_setup_timeout_seconds} seconds" if now > sleep_timeout
-      logger.debug("Done starting up repo: #{self.git_config.git_url}")
+      logger.debug("Done starting up repo: #{git_config.git_url}")
+    end
+
+    def handle_exception(ex, message: nil)
```
KrauseFx: Would be good to have documentation here on why we add `message` as a separate parameter

app/shared/models/git_repo.rb
```diff
@@ -106,32 +118,55 @@ def initialize(git_config: nil, local_folder: nil, provider_credential: nil, asy
...
+      raise ex unless notification_service
+
+      message = ex.message.to_s
+      if message.contains("unable to access")
+        priority = Notification::PRIORITIES[:urgent]
+        notification_service.create_notification!(
+          priority: priority,
+          name: "Repo access error",
+          message: "Unable to acccess #{git_config.git_url}",
+          details: ex.message
```
KrauseFx: We don't show `message` anywhere, except for in the CI terminal output?

app/shared/models/git_repo.rb
```diff
@@ -106,32 +118,55 @@ def initialize(git_config: nil, local_folder: nil, provider_credential: nil, asy
...
       end
 
-      logger.debug("Adding task to setup repo #{self.git_config.git_url} at: #{local_folder}")
+      logger.debug("Adding task to setup repo #{git_config.git_url} at: #{local_folder}")
 
       setup_task = git_action_with_queue(ensure_block: proc { callback_block(async_start) }) do
-        logger.debug("Starting setup_repo #{self.git_config.git_url}".freeze)
+        logger.debug("Starting setup_repo #{git_config.git_url}".freeze)
         setup_repo
-        logger.debug("Done setup_repo #{self.git_config.git_url}".freeze)
```
KrauseFx: why change from `self.` to non `self.`

AndrewMcBurney: We removed the `RedundantSelf` cop in the #462 PR.

app/shared/models/git_repo.rb
```diff
@@ -466,17 +545,22 @@ def clone_synchronously(repo_auth: self.repo_auth)
...
 
       self.temporary_storage_path = setup_auth(repo_auth: repo_auth)
       logger.debug("[#{git_config.id}]: Cloning git repo #{git_config.git_url} to #{@local_folder}")
-      Git.clone(git_config.git_url,
-                "", # checkout into the local_folder
-                path: local_folder,
-                recursive: true)
+
+      begin
+        Git.clone(git_config.git_url,
```
AndrewMcBurney: This is a nit, but could we potentially style it as this instead:

```ruby
Git.clone(
  git_config.git_url,
  "", # checkout into the local_folder
  path: local_folder,
  recursive: true
)  
```

Thoughts?

app/shared/models/notification.rb
```diff
@@ -69,13 +74,15 @@ class Notification
     # @param  [String] user_id
     # @param  [String] name
     # @param  [String] message
-    def initialize(id: nil, priority: nil, type: nil, user_id: nil, name: nil, message: nil, created_at: nil, updated_at: nil)
+    # @param  [String] details
+    def initialize(id: nil, priority: nil, type: nil, user_id: nil, name: nil, message: nil, created_at: nil, updated_at: nil, details: nil)
```
AndrewMcBurney: Do you think we could change the rubocop max line length from [370](https://github.com/fastlane/ci/blob/master/.rubocop.yml#L174) to something like 120?

app/shared/models/git_repo.rb
taquitos: that's right, I renamed it

app/shared/models/git_repo.rb
taquitos: I agree 🌮

app/shared/models/notification.rb
taquitos: Yeah, totally. I think 120 is perfect 👍

taquitos: (feel free to open an issue for that!)

AndrewMcBurney: Sounds good! Will do!

AndrewMcBurney: Updated in #506

app/shared/models/git_repo.rb
KrauseFx: Nice, sorry I missed that :)

PR #511, opened 2018-04-03
We don't want to have <<~ style multi-line everywhere since it breaks up logging

taquitos:
# rubocop:disable Metrics/LineLength some of the lines, and rewrite others to work with our new line length requirements.


taquitos:
cc @AndrewMcBurney

KrauseFx:
so bold

Issue #496, opened 2018-04-02
Clear out inactive BuildRunners?


KrauseFx:
I was just thinking: now that we load existing artifacts, there is no reason to keep the BuildRunner objects around, right? Can we trash them?

KrauseFx:
Boom https://github.com/fastlane/ci/pull/516

PR #516, opened 2018-04-03
Automatically clean completed build runners

KrauseFx:
- Make build_runner a required parameter for `add_build_runner`
- Automatically clean completed build runners
- Refactor to have named blocks for all BuildRunners & required parameters
- Don't call the completion block twice

Fixes https://github.com/fastlane/ci/issues/496

I haven't done any memory debugging in Ruby yet. With this PR we for sure remove the reference we were holding in the array, but how we can verify the BuildRunner object actually gets deallocated?

### DIFF
--- a/app/services/build_runner_service.rb
+++ b/app/services/build_runner_service.rb
@@ -28,6 +28,10 @@ module FastlaneCI
       return task
     end
 
+    def remove_build_runner(build_runner:)
+      build_runners.delete(build_runner)
+    end
+
     def build_runner_task_queue
       @_build_runner_task_queue ||= TaskQueue::TaskQueue.new(name: "build runner service")
     end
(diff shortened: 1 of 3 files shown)


KrauseFx:
Still pushing up, slow Mac is slow...

app/services/build_runner_service.rb
```diff
@@ -28,6 +28,10 @@ def add_build_runner(build_runner: nil)
       return task
     end
 
+    def remove_build_runner(build_runner:)
+      build_runners.delete(build_runner)
```
taquitos: This might not be thread-safe 😨

taquitos: Just checked, Arrays aren't thread-safe, but many interpreters are broken-ish (depending on who you ask 🙃) and don't do threading properly, so arrays accidentally become thread-safe. We shouldn't rely on that though and we should use proper locking for this.

Issue #517, opened 2018-04-03
Improve debugging by disabling pushes with env var


taquitos:
`FASTLANE_CI_DISABLE_PUSHES` so that we don't push to the remote repo.
- Useful when debugging your ci config using the same config as your production setup

taquitos:
🎷:goat:

PR #518, opened 2018-04-03
Disable pushes with env var

taquitos:
solves #517

### DIFF
--- a/app/shared/models/git_repo.rb
+++ b/app/shared/models/git_repo.rb
@@ -40,7 +40,9 @@ module FastlaneCI
   # It is **important** that from the outside you don't access `GitRepoObject.git.something` directly
   # as the auth won't be setup. This system is designed to authenticate the user per action, meaning
   # that each pull, push, fetch etc. is performed using a specific user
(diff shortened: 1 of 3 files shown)


app/shared/models/git_repo.rb
```diff
@@ -61,6 +63,16 @@ class GitRepo
     attr_accessor :callback
 
     class << self
+      def pushes_disabled?
```
KrauseFx: That seems like common ENV variable flag handling - don't we already have other boolean flag ENV variables where we can re-use the code? Maybe we should move this as a helper method to `FastlaneCI::EnvironmentVariables`?

taquitos: I didn't want to put it there because it's not a general property, it's very specific to this one class, so it didn't make sense to me to put them there. This is more of a fastlane.ci developer-specific switch. I wouldn't want our normal ci users to think they should muck around with it.

taquitos: I just updated the class comment for `EnvironmentVariables` to note this.

PR #523, opened 2018-04-03
Remove static ruby version

KrauseFx:
This makes the production use of the alpha version a lot more complicated, as the user will have to have this exact version of Ruby. I propose we remove it for now to make it easier for alpha testers to use `fastlane.ci` 👍


taquitos:
Can we specify 2.3.0 up to but not including 2.4?

KrauseFx:
Doesn't seem to be possible, also not possible as part of the Gemfile 😢 😢

PR #528, opened 2018-04-03
Sleep while RateLimitExceeded instead of crash the whole server

### LINKED ISSUE #527: [Octokit] CI crashes when RateLimitExceeded is reached
```
[2018-04-03 18:56:11] INFO Services:  Reseting all services
[2018-04-03 18:56:11] DEBUG OnboardingService:  local configuration repo doesn't exist
bundler: failed to load command: rackup (/Users/GabrielVillarrubia/.rbenv/versions/2.5.0/bin/rackup)
Octokit::TooManyRequests: GET https://api.github.com/repos/minuscorp/ci-config-1: 403 - API rate limit exceeded for 2.154.114.246. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) // See: https://developer.github.com/v3/#rate-limiting
```

minuscorp:
Fixes #527

### DIFF
--- a/app/services/configuration_repository_service.rb
+++ b/app/services/configuration_repository_service.rb
@@ -9,20 +9,6 @@ module FastlaneCI
     include FastlaneCI::Logging
     include FastlaneCI::GitHubHandler
 
-    class << self
-      # Loads the octokit cache stack for speed-up calls to github service.
-      # As explained in: https://github.com/octokit/octokit.rb#caching
-      def load_octokit_cache_stack
-        @stack ||= Faraday::RackBuilder.new do |builder|
-          builder.use(Faraday::HttpCache, serializer: Marshal, shared_cache: false)
-          builder.use(Octokit::Response::RaiseError)
-          builder.adapter(Faraday.default_adapter)
-        end
-        return if Octokit.middleware.handlers.include?(Faraday::HttpCache)
-        Octokit.middleware = @stack
-      end
-    end
-
     # @return [Octokit::Client]
     attr_reader :client
 
@@ -30,11 +16,7 @@ module FastlaneCI
     #
     # @param  [ProviderCredential] provider_credential
     def initialize(provider_credential: nil)
-      ConfigurationRepositoryService.load_octokit_cache_stack
       @client = Octokit::Client.new(access_token: provider_credential.api_token)
-      if client.access_token.nil?
-        client.access_token = provider_credential.api_token
-      end
       begin
         if client.rate_limit!.remaining.zero?
           sleep_time = client.rate_limit!.resets_in
@@ -44,6 +26,8 @@ module FastlaneCI
       rescue Octokit::TooManyRequests => ex
         logger.error(ex)
         raise ex
+      rescue Octokit::Unauthorized => ex # Maybe the token does not give access to rate limits.
+        logger.error(ex)
       end
     end
 
--- a/app/services/configuration_repository_service.rb
+++ b/app/services/configuration_repository_service.rb
@@ -17,18 +17,6 @@ module FastlaneCI
     # @param  [ProviderCredential] provider_credential
     def initialize(provider_credential: nil)
       @client = Octokit::Client.new(access_token: provider_credential.api_token)
-      begin
-        if client.rate_limit!.remaining.zero?
-          sleep_time = client.rate_limit!.resets_in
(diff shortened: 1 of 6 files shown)


app/services/configuration_repository_service.rb
```diff
@@ -14,6 +14,18 @@ class ConfigurationRepositoryService
     # @param  [ProviderCredential] provider_credential
     def initialize(provider_credential: nil)
       @client = Octokit::Client.new(access_token: provider_credential.api_token)
+      begin
```
taquitos: This is now in two places, do you think there is a way to generalize this logic so we only have it in one place?

minuscorp:
I was thinking the same, but because I have no knowledge of any class that is shared for this, I did the copy paste (not desired) my plans were to make an `GitHubProviderUtility` class that managed the cache Staci and the creation of clients (with the timeouts and scopes being checked) before serving them to the class. Again, I'm pretty busy until next Monday so I think I won't have time to do the refactor in time for the alpha😓

taquitos:
No worries! It's not a big deal as it mostly impacts us developers because we're restarting the server so often.

minuscorp:
I'll try to update it but cannot promise that will arrive for alpha tho 😊

KrauseFx:
This is now handled thanks to https://github.com/fastlane/ci/pull/584, right?

taquitos:
Negative ghost rider.

#548 only handles 5xx errors. Rate limit is *probably* a 429, but I don't know. Also, it looks like the octokit client has API to figure out if you hit a rate limit before actually making a request. #548 only deals with exceptions raised after the request.

minuscorp:
In #527 you have more information about the traceback that raises this issue.

taquitos:
@minuscorp yeah, that would be cool with me 👍

minuscorp:
I'm wondering if there's some way that we can inject the `Octokit::Client` onto the `github_action` or create the `Octokit::Client` in the `github_action` in order to retrieve `rate_limit`, `scope`. It'd be easy to do:
```ruby
github_action(access_token:) do |client|
    client.whatever action
end
``` 
or even to make it less confusing:
```ruby
github_client(*params) do |client|
   client.whatever_action
end
```

taquitos:
@minuscorp that's a good idea. What if we just require passing in `client` as a param instead of access token?

minuscorp:
Fair enough

app/shared/fastfile_peeker.rb
```diff
@@ -53,7 +53,7 @@ def fastfile_from_repo(repo_config: nil, branch: nil, sha: nil)
     end
 
     def fastfile_from_contents_map(contents_map)
-      github_action do
+      github_action(@client) do
```
AndrewMcBurney: Could we add an `attr_reader` above the constructor for `@client` and refer to that here, please?

app/shared/github_handler.rb
```diff
@@ -34,8 +34,22 @@ module FastlaneCI
   # TODO: handle if the user doesn't have write permission and they are trying to do some writing
   module GitHubHandler
     include FastlaneCI::Logging
-    def github_action(&block)
+    def github_action(client, &block)
       retry_count ||= 0
```
AndrewMcBurney: Could we use `retry_count = 0` please?

minuscorp: Afaik when Rey retry is called the variable context is preserved, so it would be zero over and over.

app/shared/fastfile_peeker.rb
minuscorp: It seems that this was on purpose to not expose the client out of the class scope I think 🤔

taquitos: That's right, I think it's best to not expose this outside of the class as a way of messaging to people "hey, don't access this ivar out of this class".

minuscorp: I'll resolve the conflict this afternoon and merge it if that's ok for you!

taquitos: 🎷:goat:

minuscorp: D'oh

AndrewMcBurney: Ahh, no worries - I didn't know about it until I read [poodr](http://www.poodr.com/). Highly recommend reading it! It's probably my favourite ruby book! 📖

PR #529, opened 2018-04-03
Should return the plain @api_token if it has not been encrypted yet

### LINKED ISSUE #530: We return `nil` from ProviderCredential `api_token`.
We should return at least the `@api_token`, didn't it?

minuscorp:
Fixes #530


taquitos:
Just merged a different patch for this 👍

PR #533, opened 2018-04-03
Add view model data for the projects list endpoint

nakhbari:


app/features-json/project_json_controller.rb
```diff
@@ -9,8 +10,11 @@ class ProjectJSONController < AuthenticatedControllerBase
       current_provider_credential = check_and_get_provider_credential
       current_user_config_service = self.current_user_config_service
       all_projects = current_user_config_service.projects(provider_credential: current_provider_credential)
+      all_projects_views_models = all_projects.map do |project|
+        ProjectSummaryViewModel.viewmodel_from!(object: project)
```
minuscorp: I think we can do this in a single line:

```ruby
.map(&ProjectSummaryViewModel.method(: viewmodel_from!))
```

app/features-json/view_models/view_model.rb
```diff
@@ -67,7 +67,7 @@ def viewmodel_from!(object:)
         raise "Override base_model(model) in order to use the ViewModel mixin." if @base_model.nil?
```
minuscorp: I think we should remove the named parameter `object:` and just keep it with `object`, that will clean things up and does not add any useful information. What do you think?

app/features-json/view_models/view_model.rb
```diff
@@ -67,7 +67,7 @@ def viewmodel_from!(object:)
...
         raise "Incorrect object type. Expected #{@base_model}, got #{object.class}" unless object.kind_of?(@base_model)
         if object.respond_to?(:to_object_dictionary)
-          return object.to_object_dictionary(ignore_instance_variables: @base_model.instance_variables - included_attributes)
+          return object.to_object_dictionary(ignore_instance_variables: object.instance_variables - included_attributes)
```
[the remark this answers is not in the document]
nakhbari: The @base_model had an empty list of instance variables. Is this to be expected?

app/features-json/project_json_controller.rb
nakhbari: nice!

app/features-json/view_models/view_model.rb
nakhbari: Not sure i'm following what you mean, can you propose something in a follow up PR?

app/features-json/view_models/view_model.rb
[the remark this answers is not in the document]
minuscorp: You're right, totally expected, `@base_model` is a class, not an instance.

nakhbari:
@minuscorp just read a little of the ruby style guide. It seems that we should only use the `!` at the end of a method if there is a safe version of that dangerous method.

https://github.com/bbatsov/ruby-style-guide#dangerous-method-bang

So i'm gonna remove than bang for now.

minuscorp:
Well, in the backend we are using the bang for methods that potentially raise exceptions, I think we should keep the bang just for clarity that the method is not safe if used wrongly

nakhbari:
re-added the bang

app/features-json/view_models/view_model.rb
taquitos: @minuscorp it looks like that's already done 🎉 is there anything else that's missing from this PR?

minuscorp: I think this: https://github.com/fastlane/ci/pull/566 supersedes this, but @nakhbari should decide 😄

PR #552, opened 2018-04-04
Raise exception if current user doesn't have access to repo

KrauseFx:
Better to raise an exception, than to fail with a `nil` error later on


KrauseFx:
Yes, I think we'll have to do proper error handling here anyway, and show an actual dialog, etc. Happy to add a card for this, for now, I ran into this by manually modifying the URL and running into a `nil` method error kind of thing

PR #561, opened 2018-04-04
Encrypt api token and store in correct ivar

### LINKED ISSUE #544: Unable to check repository status on launch if repo is private
During startup we check to make sure the ci-config repo is in a good state. If it's a private repo, this can fail.

taquitos:
fixes #544 
cc @AndrewMcBurney just a heads up, this was broken due to the unify pr. we worked on. Just wanted to ping you since this is a great thing to see that broke. No hard feelings 😄

### DIFF
--- a/app/shared/models/github_provider_credential.rb
+++ b/app/shared/models/github_provider_credential.rb
@@ -30,19 +30,23 @@ module FastlaneCI
       @id = id || SecureRandom.uuid
       @email = email
       @full_name = full_name
-      @api_token = api_token
       @provider_name = "GitHub"
       @type = PROVIDER_CREDENTIAL_TYPES[:github]
       @remote_host = "github.com"
+      @encrypted_api_token = encrypt_api_token(api_token)
+    end
+
+    def encrypt_api_token(token)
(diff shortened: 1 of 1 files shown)


app/shared/models/github_provider_credential.rb
```diff
@@ -30,21 +30,25 @@ def initialize(id: nil, email: nil, full_name: nil, api_token: nil)
       @id = id || SecureRandom.uuid
       @email = email
       @full_name = full_name
-      @api_token = api_token
```
taquitos: this used to be `self.api_token=` which called the setter we wrote, so when `@api_token` was swapped, it no longer provided access to the token since we overrode the getter and setter to decrypt and encrypt `@encrypted_api_token` on the fly.

AndrewMcBurney: Ahh, my apologies, that makes sense - was an overlook on my part. Thank you for finding the issue and fixing it!

taquitos: No worries! It's a tricky one, but really good to learn from!

PR #566, opened 2018-04-04
Updated ViewModel to support multiple embedded models.

minuscorp:
There's an working example at: https://repl.it/repls/StaleGentleEditors

### DIFF
--- a/app/features-json/view_models/view_model.rb
+++ b/app/features-json/view_models/view_model.rb
@@ -91,18 +91,21 @@ module FastlaneCI
 
     # Add this as Class Methods.
     module ClassMethods
+      attr_reader :_base_models
+
       # Required method to be called overriden by the created ViewModel
       # in order to indicate which class is the Model from which we are creating the ViewModel form.
       # @param [Class] models
       def base_models(*models)
-        @base_models = models
+        @_base_models = models
       end
 
       # An array of the included attributes of the base_model to be included in the ViewModel.
       # @return [Array<Symbol>] the included attributes for the given ViewModel. Defaults to all attributes.
+      # Defaults to all attr_accessor, attr_writer and attr_reader properties.
       def included_attributes
         attributes_per_model = {}
-        @base_models.each do |model|
+        _base_models.each do |model|
           attributes_per_model[model.to_s] = model::ATTRS
                                              .reject { |method| method.to_s.end_with?("=") }
                                              .map { |method| "@#{method}".to_sym }
@@ -126,9 +129,10 @@ module FastlaneCI
       private
 
       def _viewmodel_from(object)
-        raise "Override base_model(model) in order to use the ViewModel mixin." if @base_models.nil?
-        raise "Incorrect object type. Expected #{@base_models}, got #{object.class}" \
-          unless @base_models.any? { |base_model| object.kind_of?(base_model) }
+        raise "Override base_model(model) in order to use the ViewModel mixin." if _base_models.nil?
+        unless _base_models.any? { |base_model| object.kind_of?(base_model) }
+          raise "Incorrect object type. Expected #{_base_models}, got #{object.class}"
+        end
         if object.respond_to?(:to_object_dictionary)
           return object.to_object_dictionary(included_attributes[object.class.to_s])
         else
--- a/app/features-json/view_models/view_model.rb
+++ b/app/features-json/view_models/view_model.rb
@@ -1,6 +1,27 @@
 require "../../shared/json_convertible"
 
 module FastlaneCI
+  # Module used to get the attr_accessor, attr_writer and attr_reader associated methods for
+  # a given class, without needing to instanciate it.
+  # Credit: https://stackoverflow.com/a/34440466/4161167
+  # To use it, just call: YourClass::ATTRS.
+  # @return [Array<Symbol>]
+  module MethodTracer
+    TracePoint.trace(:c_call) do |t|
+      if t.method_id.to_s.start_with?("attr_")
+        t.self.extend(MethodTracer)
+
+        methods = t.self::ATTRS ||= []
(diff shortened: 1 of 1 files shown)


minuscorp:
Don't merge this yet as I have to add some fixes before😊

Gemfile.lock
```diff
@@ -1,14 +1,12 @@
 GIT
```
taquitos: let's not commit the Gemfile.lock

minuscorp: Ugh, do I make the commits again or can let it stay for this one (I'll remember not to commit it again)

taquitos: It's tricky right now because we're trying to keep the gemfile.lock on the most current master, instead of a version of fastlane, so I've noticed that when we commit the .lock file, sometimes I lose a critical change in fastlane/master. For right now, let's just back out that file 😞 It should get better in the future. If we commit this, as is, then fastlane.ci will break due to a missing commit on fastlane 🙃

minuscorp: Alright, I'll restart the branch and commit again no problem 😄

minuscorp:
I want to make sure that conceptually this is clear: view models are the mean to connect both models and views. Everything out of that scope (I.e., the example of serving Artifact links) can be abstracted with this architecture too. The only counterpart is that, as you say, there will be parts of the application that cannot be served through models and that's ok. This is just a tool to, whenever is possible, encourage the developer to adopt an architecture pattern. Anyway, this can be extended and allow users to include methods from models to be evaluated with needs as you have told. I think this is mergeable and we can make follow up PRs with all the needs your might encounter.

app/features-json/view_models/view_model.rb
```diff
@@ -48,29 +93,46 @@ def included(base)
...
+        viewmodel = {}
+        objects.each do |object|
+          viewmodel[object.class.to_s] = _viewmodel_from(object)
+        end
+        return viewmodel
+      end
+
+      private
+
+      def _viewmodel_from(object)
```
AndrewMcBurney: This is a bit of a nitpick / preference thing, but it's not really convention to prefix private methods with underscore `_` in Ruby (like you would in Python). Would it be possible to rename this method to `viewmodel_form`?

app/features-json/view_models/view_model.rb
```diff
@@ -48,29 +93,46 @@ def included(base)
     module ClassMethods
       # Required method to be called overriden by the created ViewModel
       # in order to indicate which class is the Model from which we are creating the ViewModel form.
-      # @param [Class]
-      def base_model(model)
-        @base_model = model
+      # @param [Class] models
+      def base_models(*models)
+        @base_models = models
```
AndrewMcBurney: Could you add an `attr_reader :base_models` if it's not mutated, please?

app/features-json/view_models/view_model.rb
minuscorp: It has the same name as the public method, and I want to make sure it's just the same method but I'm private way to allow better code management😕

taquitos:
I love the ideas here, but I think this is getting into the realm of too much meta-programming. I can't really grok what is going on, and I think for people who will be maintaining/bug fixing in the future, this requires a lot of thought and understanding before changes can be made.

Also, I think that for what we're trying to accomplish at this layer, it's not the right way to move forward. At this layer we will be picking and choosing which attributes we use, so it's better that we define them in each view model, explicitly, instead of relying on meta-programming magic and lists of attributes and mappings.

I know that means we can't generalize this solution like we did with `JSONConvertible`, but that's ok because it's clearer what is going on.

We're ok with duplication in the ViewModel layer and keeping coupling very light, as @nakhbari had suggested.
I especially like this idea:
>  individual view classes converting the models to json.

But doing it in a was that isn't as magical and is much clearer to novice eyes.

minuscorp:
Just for finishing my rationale, in an architectural way, I think that not tieing developers to some _architectural standards_ will end up in contributions of random ViewModels where the information where its information comes from a questionable resource. I this way, coupling the Model (a set of attributes and methods, whatever they are) and adding to them a `ViewModel` that automatically generates the needed JSON is a good thing, in terms that if they don't end up with the desired result, is that they're not using in a correct manner the ViewModel layer. Saying that `base_models` couple models with the ViewModel is a vague description, because `base_models` can be any object really, any information container, ViewModel's responsibilities is just to pipe that information sources to a JSON representation. I can do the upside-down implementation, where `ViewModel` does not include every method and every attribute by default, but provide methods to obtain them given a `Model`, just as a helper. The rest of the API I think is very well documented and easy to use, in a one-liner. Reconsider this before getting close (at least for the many hours of work spent on this 🤣 )

taquitos:
I know 😞, it's really cool too. There are a few reasons why this isn't the right thing _right now_ for the project. It's not that there might be a bug (there's always a bug anyway 🙃), but more, if we need to add 1 more feature, it will be very difficult because somebody would need to spend a lot of time understanding it. We also are not quite ready to use an abstraction around this just yet. Mainly, we aren't done really solidifying how we want to pass data to and from the client, what parts of the data we want to push to the client, and so for the time being we want to see how that process evolves and when we are really comfortable with how we use the view models then we should revisit generalizing it more. We need more flexibility right now, but at some point in the future we'll be able to understand our needs better.

You've done a really great job generalizing view models, but we aren't ready to commit to a general solution until we have more data around how we're going to use things in actual practice. 

I don't want you think this work was all done in vain, it helped us understand the direction we don't want to go in just yet. Let's shelf this for now and see how things develop. We might end up wanting to use it in the future.

taquitos:
Really, not a waste of time 😄, it helped us think more about what we wanted. Sorry you spent so much time on it. I'm going to close this for now, but we'll revisit it in the future. Thanks for your understanding 👍

PR #572, opened 2018-04-05
Make use of LaneManager to consider the default_platform

KrauseFx:
Needs https://github.com/fastlane/fastlane/pull/12239 to be merged
Fixes https://github.com/fastlane/ci/issues/571

### DIFF
--- a/app/features/build_runner/fastlane_build_runner.rb
+++ b/app/features/build_runner/fastlane_build_runner.rb
@@ -76,8 +76,8 @@ module FastlaneCI
         return
       end
 
-      fast_file = Fastlane::FastFile.new(fast_file_path)
       FastlaneCore::Globals.verbose = true
+      ENV["FASTLANE_SKIP_DOCS"] = true.to_s
 
       begin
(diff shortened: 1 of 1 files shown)


app/features/build_runner/fastlane_build_runner.rb
```diff
@@ -76,8 +76,8 @@ def run(new_line_block:, completion_block:)
         return
       end
 
-      fast_file = Fastlane::FastFile.new(fast_file_path)
       FastlaneCore::Globals.verbose = true
+      ENV["FASTLANE_SKIP_DOCS"] = true.to_s
```
KrauseFx: This would otherwise generate the new `README.md` in the `./fastlane` folder, we don't need that as part of CI, ever 👍

taquitos: true

KrauseFx:
I'll look into that TODO item

KrauseFx:
Let's leave that TODO for now, it's a little more complex to resolve, and isn't critical for the alpha

PR #577, opened 2018-04-05
Rescue exception in Sinatra middleware

KrauseFx:
This previously caused the whole server to crash, this should be fine now. Also I'm not catching just StandardError, as we probably want to catch ANY kind of error, as not getting real-time output is not as bad as the whole server crashing

### DIFF
--- a/app/features/build/build_websocket_backend.rb
+++ b/app/features/build/build_websocket_backend.rb
@@ -87,6 +87,9 @@ module FastlaneCI
 
       # Return async Rack response
       return ws.rack_response
+    rescue => ex
+      logger.error("Error in #{self.class} middleware: #{ex}")
+      logger.error(ex.backtrace.join("\n"))
     end
   end
 end


KrauseFx:
Actually nooooooo, this seems dangerous and seems to break other things...

Without this change:
screenshot 2018-04-05 11 57 32

---
With this change
screenshot 2018-04-05 11 57 34

KrauseFx:
cc @taquitos, I'm gonna close this, as I don't want to dive into this right now. But seems like this simple `rescue` caused other areas in the app to break

KrauseFx:
Oh, note the error message in black being very different

taquitos:
@KrauseFx close or nah?

KrauseFx:
Yes, close, but I'm so confused on why this happens, do you have an idea?

KrauseFx:
Yes, close, but I'm so confused on why this happens, do you have an idea?

Issue #583, opened 2018-04-05
[Tricky] Build output shows error message that lane can't be found


KrauseFx:
In the build output, chances are you see 
```
Could not find 'test'. Available lanes: ios bootstrap, ios test, ios beta, ios appstore
```
at least depending on how `platform` was set up, you can use https://github.com/themoji/ios to reproduce.

This is caused by https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/fast_file.rb#L168
```ruby
    def is_platform_block?(key)
      UI.crash!('No key given') unless key

      return false if self.runner.lanes.fetch(nil, {}).fetch(key.to_sym, nil)
      return true if self.runner.lanes[key.to_sym].kind_of?(Hash)

      if key.to_sym == :update
        # The user ran `fastlane update`, instead of `fastlane update_fastlane`
        # We're gonna be nice and understand what the user is trying to do
        require 'fastlane/one_off'
        Fastlane::OneOff.run(action: "update_fastlane", parameters: {})
      else
        UI.user_error!("Could not find '#{key}'. Available lanes: #{self.runner.available_lanes.join(', ')}")
      end
    end
```

When you run `fastlane` directly, you'll see that we catch the message in 
https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/lane_manager.rb#L19-L22
```ruby
      begin
        is_platform = ff.is_platform_block?(lane)
      rescue # rescue, because this raises an exception if it can't be found at all
      end
```
We'll still show it as part of `fastlane.ci` as we show the `user_error!` immediately, and not like in fastlane, only if we completely surface the exception.

This can be solved using multiple approaches. 
1. Show the error message only if the exception surfaced all the way up like we do it in fastlane (this might be good as we might have the same problem in other areas of the code base
1. Fix this particular instance, we shouldn't raise and swallow an exception. This is risky though, as changing the lane/platform management might cause other unexpected side effect we don't want

KrauseFx:
Actually this marks the whole build as build_failure due to https://github.com/fastlane/ci/blob/master/app/features/build_runner/build_runner_output_row.rb#L37-L43

```ruby
    # Did this particular message fail the build? (e.g. `user_error` or `build_error`)
    def did_fail_build?
      # The first time this method is called, we check if this row failed the build
      if @_did_fail_build.nil?
        @_did_fail_build = BUILD_FAIL_TYPES.include?(type)
      end
      return @_did_fail_build
    end
```

[...]

PR #585, opened 2018-04-05
Let fastlane exceptions bubble up

KrauseFx:
This will do a few things
- `UI.error` is an error, but not a critical one, it doesn't fail the build
- We let the exceptions bubble up before showing the message and/or failing the build with that particular message
- Show crashes in red also
- Stop using messages to detect a build failure for now, we have exception handling for that to allow catching & re-raising of exceptions

Fixes https://github.com/fastlane/ci/issues/583


KrauseFx:
Working on that feedback

KrauseFx:
Gemfile.lock is still being uploaded

PR #600, opened 2018-04-05
Don't duplicate output from fastlane now that UI.X is connected

taquitos:


app/shared/models/local_artifact_provider.rb
```diff
@@ -15,24 +15,21 @@ def default_root_path
       end
     end
 
-    # @return [Pathname]
```
KrauseFx: is that change related?

taquitos: Yeah, I combined them. Will pull this back out though

PR #611, opened 2018-04-06
Download just the fastfile from github

### LINKED ISSUE #610: Onboarding lane selection should be much faster
When adding a new project, selecting the branch is fast, but selecting the lane is really slow. We can do much better.

- Do we do a shallow clone?
- Additionally we should clone just a list of files, meaning let's check if `fastlane/Fastfile` or `.fastlane/Fastfile` is there, if not fallback to clone the whole repo

taquitos:
- Fixes #610
- Attempt to download fastfile from remote repo without checkout
- Checkout only if we couldn't download
- Update Gemfile.lock to newest fastfile_parser
- Also fixes remaining `=` that should be `==`

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,7 +6,7 @@ GIT
 
 GIT
   remote: https://github.com/fastlane/fastfile-parser
-  revision: 4ee3b130e4d42bd61ee4134708a48e05c00d44b2
+  revision: a9707d87eb6bfdc74c9fc3125f7eec917c8c1e49
   specs:
     fastfile-parser (0.0.0)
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -6,15 +6,15 @@ GIT
 
 GIT
   remote: https://github.com/fastlane/fastfile-parser
-  revision: 4ee3b130e4d42bd61ee4134708a48e05c00d44b2
+  revision: a9707d87eb6bfdc74c9fc3125f7eec917c8c1e49
```
taquitos: gotta pick up the latest revision for the api changes to fastfile-parser

Issue #296, opened 2018-03-14
Containerize fastlane.ci


taquitos:
It would be really nice to have a docker image or something that we could deploy. 

It's not a requirement since we expect most people to deploy directly on a Mac. This only benefits fastlane.ci running for fastlane.ci or any other project that doesn't require a Mac (for Xcode, and other build tools)

adellibovi:
Hi @taquitos, I created a PR (#619) to address what you are mentioning.

Eventually we could push the image to the DockerHub and/or supporting for development deployment too (which in my opinion is something nice to have in the future: not having to install specific version of Ruby and gems and still be able to contribute to the project).

taquitos:
Yey!

PR #619, opened 2018-04-07
Add initial support to Docker

adellibovi:
This solves  #296, it adds support to deploy fastlane.ci with "docker-compose up", currently it supports production mode only.


dlackty:
I would suggest adding a [`.dockerignore`](https://github.com/sameersbn/docker-gitlab/blob/master/.dockerignore) file?

adellibovi:
@dlackty what would you suggest to ignore? Currently, I would say that only `docs` folder could be ignored

taquitos:
It might be helpful to add in some basic instructions on how to use it with docker for people who aren't familiar 👍

taquitos:
@adellibovi I just pinging you on twitter 🎉

adellibovi:
@taquitos good idea! I will modify the Readme to include instruction to run with Docker 😄

adellibovi:
Made some changes on the `Readme`, please let me know how does that feels :)

dlackty:
@adellibovi ah yes, you're right and I just forgot that. 
In the long term I guessed we might need to separate this project itself and configuration repo just like Match or Fastlane itself does.

PR #622, opened 2018-04-07
[WIP] First implementation on GraphQL

minuscorp:
- [x] Simple `Project` type without complex nested types.
- [x] `Project` query by id (optional).
- [ ] `JobTrigger` types.
- [ ] `RepoConfig` type.
- [ ] `GitRepoConfig` type.
- [ ] `ArtifactProvider` type.


app/features-json/graph_ql/queries/project_query.rb
```diff
@@ -0,0 +1,20 @@
...
+
+module FastlaneCI
+  module Query
+    Projects = GraphQL::ObjectType.define do
+      name "Projects"
+      field :projects, types[Types::ProjectType] do
+        argument(:id, types.String)
+        description "Get a list of Projects"
+        resolve lambda do |_obj, args, _ctx|
+          if !args[:id].nil?
```
AndrewMcBurney: Could we do an `unless` here instead please?

app/features-json/graph_ql/schema.rb
```diff
@@ -0,0 +1,17 @@
+Dir[File.join(__dir__, "queries", "*.rb")].each { |file| require file }
+
+require "graphql"
+
+# Load FastlaneCISchema as a global.
+module FastlaneCI
+  # Wrapper class for the GraphQL schema of the application.
+  class GraphQLSchema
+    class << self
+      def schema
```
AndrewMcBurney: This is a bit of a nitpick - but could you please do `self.schema` instead of wrapping the method in `class << self`? [[1]](https://github.com/airbnb/ruby#no-class-self)

app/features-json/graph_ql/schema.rb
```diff
@@ -0,0 +1,17 @@
+Dir[File.join(__dir__, "queries", "*.rb")].each { |file| require file }
```
AndrewMcBurney: Nice ❤️

app/features-json/graph_ql/schema.rb
minuscorp: If I do that I think I should use `@@schema` which is one Rubocop's cop?

AndrewMcBurney: What's wrong with doing the following?

```ruby
class GraphQLSchema
  def self.schema
    @schema ||= GraphQL::Schema.define do
      query Query.Projects
    end
  end
end
```

app/features-json/graph_ql/queries/project_query.rb
AndrewMcBurney: I forgot about that - my bad 😳

app/features-json/graph_ql/schema.rb
minuscorp: Probably nothing, I'll check tomorrow morning, and I'll try to add some more types too 😁🚀

minuscorp:
Yeah, this is a quick demonstration on how could it work with angular and our current architecture, I'm not going to make a super huge PR without knowing that this may fit, I cc him @nakhbari and I'll talk with him this week 😊

app/features-json/graph_ql/queries/project_query.rb
```diff
@@ -0,0 +1,20 @@
+require_relative "../types/project_type"
+
+module FastlaneCI
+  module Query
+    Projects = GraphQL::ObjectType.define do
+      name "Projects"
+      field :projects, types[Types::ProjectType] do
+        argument(:id, types.String)
+        description "Get a list of Projects"
```
nakhbari: Should the description be for the ObjectType, rather than the field?

app/features-json/graph_ql/queries/project_query.rb
nakhbari: could we reverse the order of this condition?

```
if nil -> all projects
else -> get project
```

app/features-json/graph_ql/types/job_trigger_type.rb
```diff
@@ -0,0 +1,28 @@
+require "graphql"
+
+module FastlaneCI
+  # Definition of the different GraphQL types.
+  module Types
+    CommitJobTriggerType = GraphQL::ObjectType.define do
+      name "CommitJobTrigger"
+      field :branch, !types.String
```
nakhbari: Non-null ❤️

app/features-json/graph_ql/types/job_trigger_type.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    NightlyJobTrigger = GraphQL::ObjectType.define do
+      name "NightlyJobTrigger"
+      field :branch, !types.String
+      field :hour, types.Int
+      field :minute, types.Int
```
nakhbari: do these fields exist yet? If not, can we add them later?

app/features-json/graph_ql/types/project_type.rb
```diff
@@ -0,0 +1,20 @@
...
+
+module FastlaneCI
+  # Definition of the different GraphQL types.
+  module Types
+    # TODO: [WIP] We don't include nested Project's properties for now.
+    ProjectType = GraphQL::ObjectType.define do
+      name "Project"
+      description "fastlane.ci Project"
+      field :id, !types.String
+      field :name, types.String, property: :project_name
```
nakhbari: field alias 🔥

app/features-json/graph_ql/queries/project_query.rb
minuscorp: So the matter with this query is an `Object` per sé, so to diff the `Object` that mirrors the `Model` than the `Object` used as a `Query` I added that description, that being said, is totally optional and I can change it to whatevs

app/features-json/graph_ql/types/job_trigger_type.rb
minuscorp: Do they exist! `NightlyJobTrigger` has those `attr_accessor`

app/features-json/graph_ql/queries/project_query.rb
minuscorp: Because then you're skipping all the other parameters that might be present in the query itself (that are not added for now), like, projects with certain name regex, or something else.

minuscorp:
Are we going to use [Apollo](https://www.apollographql.com/docs/angular/basics/setup.html) for the integration between back and front end @nakhbari? What do you think? I think it has a strong community support and its [license](https://github.com/apollographql/apollo-client/blob/master/LICENSE) fits pretty well with ours if I'm not wrong

nakhbari:
@minuscorp Apollo looks cool. I like that you can watch a query. I think we can look more into it in the future. Right now it shouldn't be too different for us to just make the query ourselves. I think we should look into it if it proves to do a lot more heavy lifting for us. Let's move forward without it for now.

app/features-json/graph_ql/queries/project_query.rb
nakhbari: if the graphql endpoint shows the description when you select projects, all is good :)

nakhbari:
One thing we could really use right now is a relative date pipe within our Angular app. https://github.com/fastlane/ci/issues/666

Let me know if this is something you're interested in working with, I can give you pointers to documentation to get you started.

minuscorp:
I have to finish this too, if that one is not blocking either, assign it to me. I'm trying to get familiar with Angular😁

minuscorp:
Oh, I misread your previous comment, I have just finished this GraphQL thingy right now without being noticed about your review comment.
I don't really know why checks aren't passing, they both pass locally. 
So what is the conclusion? Is this not being merged for now?

Issue #454, opened 2018-03-30
Delete fastlane.ci user


taquitos:
We want to implement deleting a fastlane.ci user

PR #627, opened 2018-04-08
Remove views for provider credentials and users.

AndrewMcBurney:
Removes views for `users` and `provider_credentials`.

### Rational

- Some of the actions are currently broken
- The views are not the most aesthetically pleasing (I'm no designer :joy:)
- We plan to move to an Angular frontend in the future, where we can revisit these views
- We can still use the `post` actions in both the `UsersController` and `ProviderCredentials` controller, and extend the controllers for further usecases (i.e., adding a `delete` action for `users` https://github.com/fastlane/ci/issues/454)


taquitos:
I actually like being able to see basic user information, would it be possible to remove only the broken stuff for now?

PR #632, opened 2018-04-09
Add some better debug messaging

taquitos:
Removed repo setup step at startup since we don't need it anymore
Also fix missing require


taquitos:
We used to need a checkout of the repo before we could start tasks, but since we don't rely on it for anything other than testing permissions, we can remove it;

taquitos:
@KrauseFx aren't you supposed to be sleeping?

PR #636, opened 2018-04-10
Added a way to delete user from `users.json`

### LINKED ISSUE #454: Delete fastlane.ci user

AndrewMcBurney:
Closes #454.

### Changes 
- Introduces `/users_erb/delete/:id` action in `UsersController`
- Added `delete_user!` method in`json_user_data_source.rb`
- Added internal helper in `json_user_data_source.rb`: `find_user_index_and_existing_user(user:)`
- Added `json_user_data_source_spec.rb` with specs for deleting users

### DIFF
--- a/app/services/data_sources/json_user_data_source.rb
+++ b/app/services/data_sources/json_user_data_source.rb
@@ -131,7 +131,7 @@ module FastlaneCI
 
     # TODO: this isn't threadsafe
     def update_user!(user: nil)
-      user_index, existing_user = find_user_index_and_existing_user
+      user_index, existing_user = find_user_index_and_existing_user(user: user)
 
       if existing_user.nil?
         logger.debug("Couldn't update user #{user.email} because they don't exist")
@@ -146,7 +146,7 @@ module FastlaneCI
     end
 
     def delete_user!(user: nil)
-      user_index, existing_user = find_user_index_and_existing_user
+      user_index, existing_user = find_user_index_and_existing_user(user: user)
 
       if existing_user.nil?
         logger.debug("Couldn't delete user #{user.email} because they don't exist")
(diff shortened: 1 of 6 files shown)


app/services/data_sources/json_user_data_source.rb
```diff
@@ -180,5 +184,20 @@ def create_user!(id: nil, email: nil, password: nil, provider_credentials: [])
...
+    private
+
+    # Finds the index of the user, and returns an existing `user` if they exist
+    #
+    # @param  [User] `user` a user to lookup by `user.email`
+    # @return [Integer] `user_index` in the `users` array
+    # @return [User] `existing_user` in the `users.json` file
+    def find_user_index_and_existing_user(user:)
+      users.each.with_index do |old_user, index|
+        return [index, old_user] if old_user.email.casecmp(user.email.downcase).zero?
```
taquitos: maybe instead of email, we should compare ID? I know right now email is pretty much guaranteed to be unique, but ID is truly guaranteed to be unique

spec/services/data_sources/json_user_data_source_spec.rb
```diff
@@ -0,0 +1,78 @@
+require "spec_helper"
+require "app/shared/models/user"
+require "app/services/data_sources/json_user_data_source"
+
+describe FastlaneCI::JSONUserDataSource do
```
taquitos: ❤️:heart::heart:

app/services/data_sources/json_user_data_source.rb
taquitos: Yeah, I think I wrote that code you copied 😆 My original thought was that the update might be operating on a `User` that didn't yet contain an ID. But really, I should make that much more clear if that's what I was doing.

AndrewMcBurney: Ahh, gotcha! Thank you for clearing that up :)

PR #637, opened 2018-04-11
Default to empty array in `available_lanes` rather than `nil`.

AndrewMcBurney:
The problem was referenced in #605, in comments [[1]](https://github.com/fastlane/ci/issues/605#issuecomment-380424652), and [[2]](https://github.com/fastlane/ci/issues/605#issuecomment-380436556).

### DIFF
--- a/app/features/project/project_controller.rb
+++ b/app/features/project/project_controller.rb
@@ -252,7 +252,7 @@ module FastlaneCI
       locals = {
         project: project,
         title: "Project #{project.project_name}",
-        available_lanes: nil,
+        available_lanes: [],
         fastfile_path: nil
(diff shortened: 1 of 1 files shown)


AndrewMcBurney:
Yep, I agree it's just a bandaid. For clarification - the PR wasn't meant to fix issue #605, just to fix the startup issues users were experiencing mentioned in comments [[1]](https://github.com/fastlane/ci/issues/605#issuecomment-380424652), and [[2]](https://github.com/fastlane/ci/issues/605#issuecomment-380436556).

taquitos:
Ahhh, right, that makes sense.

PR #647, opened 2018-04-12
fix that can't find Fastfile when name of branch contains `/`

kgmyshin:
When the name of branch contains `/`, occur RuntimeError because not found FastFile 😂


kgmyshin:
I signed it!

Issue #657, opened 2018-04-12
Add TS Lint to Fastfile


nakhbari:
We have a TS linter set up, let's add it to the Fastfile so every frontend PR can be linted.

KrauseFx:
Ah nice, can you link me to the TS linter we're using? Is it prettier?

nakhbari:
This is the standard linter for TS.

https://palantir.github.io/tslint/

config:
https://github.com/fastlane/ci/blob/master/tslint.json

taquitos:
> frontend CL
😆 frontend *PR*

KrauseFx:
Thanks 👍

nakhbari:
Fixed by add tslint to the Fastfile #658

PR #658, opened 2018-04-12
add tslint to the Fastfile

nakhbari:
https://github.com/fastlane/ci/issues/657

Now we can add a project that runs tslint on the web typescript files!

##### Sample Output
```
/Library/Ruby/Gems/2.3.0/gems/fastlane-2.85.0/fastlane_core/lib/fastlane_core/ui/interface.rb:145:in `shell_error!': [!] Exit status of command 'cd ..;./node_modules/tslint/bin/tslint './web/**/*.ts'' was 2 instead of 0. (FastlaneCore::Interface::FastlaneShellError)

Warning: The 'deprecation' rule requires type information.
Warning: The 'no-use-before-declare' rule requires type information.

ERROR: web/app/dashboard/dashboard.component.ts[13, 57]: Missing semicolon
ERROR: web/app/models/project_summary.ts[1, 86]: Missing semicolon
```


KrauseFx:
Sorry for the noob question on this one, is TSLint like prettier?

fastlane/Fastfile
```diff
@@ -10,3 +10,9 @@ lane :rubocop do
   sh("cd ..;bundle install")
   sh("cd ..;bundle exec rubocop")
 end
+
+lane :tslint do
+  desc "TSLint the whole project to ensure no style violations"
+  sh("cd ..;npm install")
```
taquitos: how do we guarantee that `npm` is installed?

nakhbari:
I'm actually not too familiar with prettier. It seems like it's an automatic formatter? I have tslint configured in my VSCode to automatically format. I can provide instruction on this. Are there other functions prettier provides? I noticed there are articles on combining the two.

fastlane/Fastfile
nakhbari: good point. I following bundle as an example, but i guess there's a higher assumption that it's already installed if you can run fastlane.

Any suggestions? I'll look into it in the meantime.

nakhbari: @taquitos wouldn't the command fail, thus failing the fastlane execution? If so, that should be OK. doing a `brew install npm` might be overkill.

taquitos: Yeah, this is fine.
Let's add to the rake file whatever is the standard way of getting npm

KrauseFx:
I don't know enough about JS, I just heard good things about prettier and how many teams make use of it. Let's stick with what you already set up 👍

nakhbari:
@snatchev Can you help with adding this to git presubmit. Not really knowledgeable about that.

taquitos:
@nakhbari check out https://github.com/fastlane/ci/blob/master/.pre-commit
You can see how he does the rubocop check at https://github.com/fastlane/ci/blob/master/.pre-commit#L15

snatchev:
@nakhbari added to the pre-commit hook in https://github.com/fastlane/ci/pull/658/commits/03c3e32c4b36bc9da80c22263a8e14d64d21b6c6

One caveat: rubocop allows us to specify which ruby files to lint (in the `changed_files` array), and we only lint the ones that are about to be committed. I could not find a similar feature in tslint. Do you know if that is supported?

fastlane/Fastfile
nakhbari: done

nakhbari:
@snatchev I added a change to only do ng lint if there are modified ts files.

you can select multiple files with tslint, but ng lint does some extra stuff and it's a project based check for typing purposes.

snatchev:
@nakhbari 👍

Issue #406, opened 2018-03-28
Do we deal with passwords or API tokens?


KrauseFx:
Discussion started in https://github.com/fastlane/ci/pull/402#discussion_r177645737

My feeling is that we deal with API tokens, and should never encourage any user to provide their actual password. If that is accurate, we should reflect that in the environment variable names and config files

taquitos:
Yeah, I think you're probably right, just API Tokens, so the field is mislabeled

KrauseFx:
And if so, is there a way to prohibit the user from entering their password. This could be an after MVP task of course

taquitos:
I bet github tokens are a certain length. So at the very least, we could verify the length of the tokens and if it is incorrect, we could let them know they should use the token, not a password.

getaaron:
IIRC GitHub uses OAuth2 Bearer tokens 🐻 

https://tools.ietf.org/html/rfc6750

I don’t think there’s a guaranteed length so I’d be hesitant to include a strict length requirement to validate a token.

taquitos:
Yeah, the length is more of a loose guide, like if it length is < 20, we can be pretty confident they entered a password and we should just point out "Hey, are you sure this is a token? If so, that's cool, continue on your way"

KrauseFx:
Sounds like an easy win to just compare the length for now 👍

minuscorp:
From `Octokit` README:
```ruby
client = Octokit::Client.new(:access_token => "<your 40 char token>")
```

minuscorp:
Also, I'd recommend to migrate the `.keys` file (environment variables) into something more user-friendly (i.e., `json`, `yml`)

KrauseFx:
Yep, so keys it is. Let's not do any changes with `.keys` for now, we can solve this together with https://github.com/fastlane/ci/blob/master/docs/use_cases/edit_project_env_variables.md

KrauseFx:
And https://github.com/fastlane/ci/blob/master/docs/use_cases/edit_global_env_variables.md

KrauseFx:
No idea why I closed this, we still need to implement the check to avoid people entering their actual GitHub password

taquitos:
Done 🎉

PR #660, opened 2018-04-12
Require api_tokens to be exactly 40 chars

taquitos:
- GitHub documents this requirement in their API docs
https://github.com/octokit/octokit.rb#application-authentication
- Fix for #406


app/features/onboarding/onboarding_controller.rb
```diff
@@ -127,6 +127,18 @@ class OnboardingController < ControllerBase
...
+        api_token = params[:clone_user_api_token]
+        api_token.strip!
+        if api_token.length != 40
+          error_message = "Invalid API token, please ensure you entered an API token and not a password"
+          session[:message] = error_message
+          logger.error(error_message)
+          logger.error("API tokens are generated by GitHub and are 40 characters long")
+
+          redirect("#{HOME}/initial_clone_user")
+          return
```
AndrewMcBurney: Could you please remove the extra `return`?

taquitos: Does `redirect` cause the rest of the method to not be executed? I don't want the rest of the logic in this method to execute.

app/features/onboarding/onboarding_controller.rb
```diff
@@ -127,6 +127,18 @@ class OnboardingController < ControllerBase
     # 3) If the data is not valid, display an error message
     post "#{HOME}/initial_clone_user" do
       if valid_params?(params, post_parameter_list_for_clone_user_validation)
+        api_token = params[:clone_user_api_token]
+        api_token.strip!
```
KrauseFx: Let's append the `.strip` directly to line `:130` without the `!` to make it easier to understand for new contributors 👍

app/features/onboarding/onboarding_controller.rb
AndrewMcBurney: Yep! 🙂 

```ruby
require "sinatra"

get "/" do
  puts "Before redirect."
  redirect "/redirect_here"
  puts "After redirect."
end

get "/redirect_here" do
  puts "Redirected here."
  "Hello World!"
end
```

```
❯ ruby test.rb                                                                                                                                                                                                                             
== Sinatra (v2.0.1) has taken the stage on 4567 for development with backup from Thin
Thin web server (v1.7.2 codename Bachmanity)
Maximum connections set to 1024
Listening on localhost:4567, CTRL+C to stop
Before redirect.
::1 - - [12/Apr/2018:18:52:09 -0400] "GET / HTTP/1.1" 302 - 0.0072
Redirected here.
::1 - - [12/Apr/2018:18:52:09 -0400] "GET /redirect_here HTTP/1.1" 200 12 0.0006
::1 - - [12/Apr/2018:18:52:09 -0400] "GET /favicon.ico HTTP/1.1" 404 469 0.0008
```

taquitos: That's magical 🌟

AndrewMcBurney: 🔮

PR #662, opened 2018-04-12
Ensure post_run_action happens even after build runner fail

### LINKED ISSUE #467: Ensure the `post_run` method is called in BuildRunner
See https://github.com/fastlane/ci/pull/465/files#r178480662 for more context

taquitos:
Fixes #467


app/features/build_runner/build_runner.rb
```diff
@@ -283,8 +283,14 @@ def start
...
         logger.debug("Not using a workqueue for build runner #{self.class}, this is probably a bug")
-        work_block.call
-        post_run_block.call
+
+        begin
+          work_block.call
+        rescue StandardError => ex
+          logger.error(ex)
+        ensure
+          post_run_block.call
```
taquitos: this is already called in an ensure block if we're using a workqueue to execute it

KrauseFx: Should this GitHub comment be an actual code comment?

taquitos: Good point

PR #665, opened 2018-04-13
Improve fastfile finder for local and remote repo

### LINKED ISSUE #638: Fasfile finder should recurse repo looking for fastfile
If the fastfile isn't located at `repo/fastlane/fastfile` and we need to checkout the repo, let's make sure we recurse the entire project to find it.
- potentially might have multiple.
- let's use breadth-first search to find the parent-most file, and assume that the first one we find is the right one (for now)
- Addresses one issue found in https://github.com/fastlane/ci/issues/605

adellibovi:
Hello, this PR fixes #638.

Two main changes:
* the local search of a Fastlane file is done recursively using `Dir["**/Fastfile"]`. we will return `fastlane/Fastfile` if exists, otherwise the first found.
* the remote search has been done using the `Tree API`, which returns recursively all folder and files (there is probably a limit of nodes that can be returned, but it is not specified in GitHub API, anyway if it fails we will still relying on cloning and searching locally). As above, we will return `fastlane/Fastfile` if exists, otherwise the first found. 

A note: I preferred to use the `Tree API` over requesting for the `root` directory and making requests for each directory because it is far more efficient (i.e. one request vs one request per directory)

### DIFF
--- a/app/features/project/project_controller.rb
+++ b/app/features/project/project_controller.rb
@@ -132,14 +132,10 @@ module FastlaneCI
       )
       repo_config = GitHubRepoConfig.from_octokit_repo!(repo: selected_repo)
 
-      fastfile_parser = fastfile_peeker.fastfile_from_github(
-        repo_full_name: repo_config.full_name,
+      fastfile_parser = fastfile_peeker.fastfile(
+        repo_config: repo_config,
         sha_or_branch: branch
       )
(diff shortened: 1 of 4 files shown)


adellibovi:
Done @taquitos, have a look when you can 🍷  Thanks 😄

PR #675, opened 2018-04-17
Added view model for the Project in detail

nakhbari:


app/shared/json_convertible.rb
```diff
@@ -10,7 +10,7 @@ def self.included(base)
 
     # add these as instance methods
     module InstanceMethods
-      def to_json(options)
+      def to_json(options = {})
```
nakhbari: without this, i was getting the `Expected 1 arg but got 0` error

minuscorp: Nice catch!

PR #691, opened 2018-04-18
Add documentation around how specific triggers work

taquitos:


docs/Triggers.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Background
+
+The heart of any CI system is a set of triggers that kick off the builds when certain events happen, or on a schedule
+
+## Triggers and their events
+
+The events that we support at launch are:
+
+- `Nightly Trigger`: Build project builds on a schedule
```
nakhbari: Should this still be called Nightly if the time can change?

taquitos: I dunno, `Nightly` is really well known terminology for doing a build 1 time a day, but you're totally right...

KrauseFx: Yeah, I think we can just tweak it as we go, and start with nightly for now? In particular the `nightly` build will eventually support more complex time specifications I guess, like: run every hour

PR #692, opened 2018-04-18
Adds a spinner while loading the Projects

### LINKED ISSUE #668: Add Loading spinner while loading data on the dashboard
Parent Issue: #644 

We should show a spinner while the data is being loaded. 

Thinking we could slap on an angular spinner inside the table where the data will be loaded https://material.angular.io/components/progress-spinner/overview

minuscorp:
Fixes #668


KrauseFx:
Can you also upload a screenshot/gif to the PR?

minuscorp:
I removed them on purpose, it seemed like the headers rendered a bit earlier than the spinner and got a weird animation, so I removed the whole table, but can change it if requested 😄

PR #693, opened 2018-04-18
Fix imports for DashboardComponent Tests

nakhbari:


nakhbari:
These imports are getting annoying. Let's see if we can find a better method for common modules being imported.

minuscorp:
Do you think this would work? https://angular.io/guide/sharing-ngmodules

nakhbari:
@minuscorp yep that looks good. Maybe we could do one for just AngularMaterialShared. I think that's the one that has bulk.

nakhbari:
Working on that right now^

minuscorp:
Even Angular in their documentation snippet do [it](https://stackblitz.com/angular/ampvdvkdvoj?file=main.ts)

PR #694, opened 2018-04-18
Fix crash for when a PR is closed and branch deleted but we pull anyway

taquitos:


KrauseFx:
> Coverage decreased (-0.05%) to 55.813%
ha, ok, I'm updating the config to not fail the build

Issue #646, opened 2018-04-12
Pulling fork with merge conflicts doesn't succeed


taquitos:
Currently, when we checkout a PR that is from a fork, we pull the original repo, and then we pull the fork'd commit on top of it. If this commit leads to a merge conflict, we fail:
```
merge conflict issues:

 Error switching to a fork: https://github.com/minuscorp/ci.git, branch: graphql-basics
[2018-04-09 07:31:26] ERROR GitRepo:  git '--git-dir=/Users/liebowitz/.fastlane/ci/32903571-271e-4c87-9b32-11df3c82d5aa/builds/d5f21a1259425c06bd81ba507f5f64de9f70f9f6/.git' '--work-tree=/Users/liebowitz/.fastlane/ci/32903571-271e-4c87-9b32-11df3c82d5aa/builds/d5f21a1259425c06bd81ba507f5f64de9f70f9f6' pull 'https://github.com/minuscorp/ci.git' 'graphql-basics'  2>&1:From https://github.com/minuscorp/ci
 * branch            graphql-basics -> FETCH_HEAD
Auto-merging launch.rb
Auto-merging Gemfile.lock
Auto-merging Gemfile
CONFLICT (content): Merge conflict in Gemfile
Automatic merge failed; fix conflicts and then commit the result.
[2018-04-09 07:31:26] DEBUG FastlaneBuildRunner:  Unable to pull before checking out d5f21a1259425c06bd81ba507f5f64de9f70f9f6 from All rspecs pass, attempting checkout
[2018-04-09 07:31:26] DEBUG FastlaneBuildRunner:  Checking out commit d5f21a1259425c06bd81ba507f5f64de9f70f9f6 from All rspecs pass
[2018-04-09 07:31:26] DEBUG FastlaneBuildRunner:  Finished running All rspecs pass for d5f21a1259425c06bd81ba507f5f64de9f70f9f6
```

PR #706, opened 2018-04-20
 Use github's ref they set up for prs

taquitos:
Fix for https://github.com/fastlane/ci/issues/646
Instead of checking out master and pulling a fork on top of it, we can use the github feature where they setup a new ref for your PR. Then you can just fetch that into a branch.
No more tears for merge conflicts

Also forked the `ruby-git` temporarily while we wait and see if we can get some movement on https://github.com/ruby-git/ruby-git/pull/362

### DIFF
--- a/app/shared/fastfile_finder.rb
+++ b/app/shared/fastfile_finder.rb
@@ -38,7 +38,7 @@ module FastlaneCI
     end
 
     def self.find_prioritary_fastfile_path(paths:, path: nil)
-      path = path.nil? ? "" : "#{path}/"
+      path = path.nil? ? "" : "#{path}/".downcase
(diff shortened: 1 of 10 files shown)


app/shared/fastfile_finder.rb
```diff
@@ -38,7 +38,7 @@ def self.search_path(path:, relative_path: false)
     end
 
     def self.find_prioritary_fastfile_path(paths:, path: nil)
-      path = path.nil? ? "" : "#{path}/"
+      path = path.nil? ? "" : "#{path}/".downcase
```
taquitos: This fixes a bug that was just recently introduced

taquitos:
Merging now so we can make use of it on ci, but would still like 👀  on it

PR #710, opened 2018-04-20
Handle connection error when github can't be found

taquitos:
Prevent scheduler from getting into bad state when exception happens

### DIFF
--- a/app/shared/github_handler.rb
+++ b/app/shared/github_handler.rb
@@ -42,14 +42,37 @@ module FastlaneCI
 
     def github_action(client, &block)
       if client.kind_of?(Octokit::Client)
-        if client.rate_limit!.remaining.zero?
-          sleep_time = client.rate_limit!.resets_in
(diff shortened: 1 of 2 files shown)


app/shared/github_handler.rb
```diff
@@ -41,25 +41,19 @@ def self.included(klass)
...
-            sleep_time = client.rate_limit!.resets_in
-            logger.debug("Rate Limit exceeded, sleeping for #{sleep_time} seconds")
-            sleep(sleep_time)
-          end
-        rescue Octokit::TooManyRequests => ex
-          logger.error(ex)
-          raise ex
-        rescue Octokit::Unauthorized => ex # Maybe the token does not give access to rate limits.
-          logger.error(ex)
+        if client.rate_limit!.remaining.zero?
```
minuscorp: this must be in a `begin...rescue Octokit::Unauthorized => ex` block

PR #717, opened 2018-04-20
[WIP] Test Bundler Injection

minuscorp:


minuscorp:
🎉  @taquitos

taquitos:
from chat:

> Ohh a good test now: remove the dependency
> and see if it breaks
> from both Gemfile and Gemfile.lock, but keep the `require` and the code in the lane
> that way we can see if the original gems were actually restored, and the injected one removed

taquitos:
I feel like that last build should have failed, right?

minuscorp:
going!

minuscorp:
Seeing the logs, it doesn't even like to care that I modified the Fastfile 😔

taquitos:
What more do you think we need to do to ensure our builds are successful and use the correct gems?
Also, note that I think the work being done in https://github.com/fastlane/ci/pull/836 could help here too, since fastlane will have it's own process. Thoughts?

minuscorp:
I guess that in the moment that the builds are going to be a separate process, where _fastlane_ and other gems are going to be installed, this issue should be gone. AFAIK, at least we're installing the gems in the bundle and uninstalling them (hopefully).

taquitos:
I'm going to close this because we're pretty close to using a separate process for stuff and this should be a moot point.

PR #703, opened 2018-04-19
Make Bundler load Project's Gemfile dependencies

minuscorp:
Fixes #674

PR #718, opened 2018-04-20
Clean environment and install again after build

minuscorp:
Follows #703

### DIFF
--- a/app/features/build_runner/fastlane_build_runner.rb
+++ b/app/features/build_runner/fastlane_build_runner.rb
@@ -178,6 +178,7 @@ module FastlaneCI
               original_lockfile.write(original_lockfile_contents)
               original_lockfile.close
               Bundler.load.clean(options[:"dry-run"])
+              Bundler.reset!
               Bundler::Plugin.gemfile_install(Bundler.default_gemfile)
               definition = Bundler.definition
               definition.validate_runtime!
--- a/app/features/build_runner/fastlane_build_runner.rb
+++ b/app/features/build_runner/fastlane_build_runner.rb
@@ -98,8 +98,13 @@ module FastlaneCI
         # folder, and all the following code works
         # This is needed to load other configuration files, and also find Xcode projects
 
+        # This step is needed in case of the target Project's repo having its own gem
+        # dependencies. As we don't isolate the build process by now, we have to inject
+        # those dependencies into the CI system in order to work fine.
+        # The first step is to make a snapshot of the current state of the CI's Gemfile and Gemfile.lock
         original_gemfile_contents = File.read(Bundler.default_gemfile)
         original_lockfile_contents = File.read(Bundler.default_lockfile)
+
         # We call the safe (because is synchronized) Bundler's `chdir` and
         # install all the dependencies, if any.
         Bundler::SharedHelpers.chdir(repo.local_folder) do
@@ -110,12 +115,14 @@ module FastlaneCI
             begin
               gemfile_dir = Dir[File.join(Dir.pwd, "**", "Gemfile")].first
 
+              # In case the target repo has its own Gemfile, we parse its contents
               builder = Bundler::Dsl.new
               builder.eval_gemfile(gemfile_dir)
 
               # We already use local fastlane, so don't try to install it.
               project_dependencies = builder.dependencies.reject { |d| d.name == "fastlane" }
 
+              # Inject all other dependencies that might be needed by the target.
               added = Bundler::Injector.inject(project_dependencies, {})
               if added.any?
                 logger.info("Added to Gemfile:")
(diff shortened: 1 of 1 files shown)


app/features/build_runner/fastlane_build_runner.rb
```diff
@@ -177,6 +177,13 @@ def run(new_line_block:, completion_block:)
               original_lockfile = File.open(Bundler.default_lockfile, "wb")
               original_lockfile.write(original_lockfile_contents)
               original_lockfile.close
+              Bundler.load.clean(options[:"dry-run"])
+              Bundler.reset!
+              Bundler::Plugin.gemfile_install(Bundler.default_gemfile)
+              definition = Bundler.definition
+              definition.validate_runtime!
+              Bundler::Installer.install(Bundler.root, definition, options)
+              Bundler.require
```
nakhbari: drive by review. As a non-ruby dev this has me confused. Could we add some comments 🙏

minuscorp: Oh sorry, my bad

minuscorp:
The purpose of this PR (which is a follow up from the mentioned) is to make sure that we restore the dependencies that might have been required by the project being built.
`Bundler` does not provide a ton of docs about how to do this as it is not intended to be used mostly out of the cli.
I'd be good to wait for @taquitos to review this with me so we can experiment if this is working as intended 🙌

app/features/build_runner/fastlane_build_runner.rb
AndrewMcBurney: Could you please change the lines:

```ruby
original_gemfile = File.open(Bundler.default_gemfile, "wb")
original_gemfile.write(original_gemfile_contents)
original_gemfile.close
original_lockfile = File.open(Bundler.default_lockfile, "wb")
original_lockfile.write(original_lockfile_contents)
original_lockfile.close
```

to

```ruby
File.write(Bundler.default_gemfile, original_gemfile_contents)
File.write(Bundler.default_lockfile, original_lockfile_contents)
```

taquitos:
@minuscorp I think I'll be able to get to this in the next hour or so, thank you for working on it!
Before we merge, I'd love to see the changes @AndrewMcBurney suggested, and also if you could add your comments to the actual source code, that would be 💯 
Then once we merge, I can throw it up on the build machine and we can test it out!

minuscorp:
Alright! It's a little bit late here, tomorrow I'll tackle the review and add some comments to make it ready to merge as soon as possible🙌🙌

taquitos:
Awesome, have a good night!

app/features/build_runner/fastlane_build_runner.rb
```diff
@@ -128,6 +135,7 @@ def run(new_line_block:, completion_block:)
                 end.join("\n"))
               end
 
+              # Install the new Bundle and require all the new gems into the runtime.
```
taquitos: How do we handle multiple builds here? I'm concerned that we might get into a state where we inject dependencies at the same time as another build runner is removing them. How can we be sure that we don't get into this state?

minuscorp: AFAIK, being in the `Bundler::SharedHelpers.chdir` block, the process is synchronized, therefore this _shouldn't_ be an issue. But this will need a lot of testing tho.

taquitos: Yeah but we're talking about 3 parts here: `saving state, modifying it` -> `running build` -> `restoring state`
If this process isn't atomic throughout, then another build running might interlace operations

minuscorp: Two options:
1. Synchronize it (CI's `Bundle` is atomic, and then builds are too).
2. Do not synchronize it, erase the restoration process, just get along with the injected deps until we can really isolate the process.

minuscorp:
So as I've seen in the Swift code, you mostly spawn a `Thread` that communicates with the runner through a `Socket`. That kind of fits with the proposal I made somewhere I can't remember of using either a `Socket` for two-way communication, or a `IO.Pipe` for unidirectional communication. I think this should be managed by `TaskQueue` tho, expand its capabilities to manage this communication and isolation topics. Don't know what you think

taquitos:
I think expanding `taskqueue` to manage that stuff is probably too much for that component. 
Right now, we have a scheduler that polls GitHub, which uses the taskqueue to create a job that executes a `BuildRunner` when a build is needed. 

I actually think the `FastlaneBuildRunner` should be responsible for spinning up a `fastlane` instance and feeding it whatever data it needs to start. Then, like you mentioned, a socket for 2 way coms is 💯 just like we do in `fastlane.swift`

minuscorp:
I'll need to dig deeper into how the Swift fastlane works to be able to, as you say, make `FastlaneBuildRunner` responsible of spawning the correct instance (Bundle, Fastlane and communication) and so on. I'll work on this next week. If you want to formalize this in an issue I'll be grateful or if not, I can do it as soon as I can and assign it to myself

taquitos:
I can create an issue for it 👍

Issue #698, opened 2018-04-18
Status icon colors 🌈


nakhbari:
Parent Issue: #644 

Get the icons right!

PR #720, opened 2018-04-20
Status icons Component 🌈

nakhbari:
Issue: https://github.com/fastlane/ci/issues/698

icons and colors subject to change, but now that's super easy with this modular component!

#### Desktop Screenshot
image

#### Mobile Screenshot
image


taquitos:
What do you think about using something more like a smily-face shrugging instead of the yellow smily? That way it better communicates that we don't know what the state is yet, because we're waiting for the build to be finished.

nakhbari:
@taquitos The icon types and colors will change. @chemichelle is working on it. This is just to get the infra in.

taquitos:
Hawt. Love it.

web/app/common/components/status-icon/status-icon.component.html
```diff
@@ -0,0 +1,14 @@
+<ng-container [ngSwitch]="status">
+  <mat-icon *ngSwitchCase="BuildStatus.SUCCESS" class="fci-status-icon-success">
+    sentiment_very_satisfied
+  </mat-icon>
+  <mat-icon *ngSwitchCase="BuildStatus.PENDING" class="fci-status-icon-pending">
+    sentiment_neutral
```
taquitos: `sentiment_unknown` might be better?

taquitos: ehh, maybe not. Yo no se

nakhbari: These are all the strings provide by the material icon set https://material.io/icons/

`sentiment_unknown` isn't a provide icon.

Issue #667, opened 2018-04-13
Add button to create new project on dashboard


nakhbari:
Add button to create new project.

#### Wireframe
![image](https://user-images.githubusercontent.com/2754461/38759405-aa3154ea-3f2a-11e8-8c2a-a0e7c6a50e36.png)

nakhbari:
This is blocked until we have a new project page. Please don't start yet.

taquitos:
What about adding a row in the dashboard after the last project that is a link/button/something to `add project`? @chemichelle would that make the UX gods cry?

chemichelle:
@taquitos adding 'add project' button to the last row of the project will become problematic when users start having many projects. If you're concern is that the 'add project' button in the header will be hidden after users scroll down, we can play with subtle animations where the header becomes skinnier and remains sticky (see an example from my iPhone here: [https://drive.google.com/corp/drive/u/0/folders/1g1JMDL3WFfzwfuxlm7Wy3GUdG6Aud-PS](url) )

taquitos:
Cool, makes sense 👍 I do really like the button in the header, and we can even defer on the header animations since most people will have less than 10 projects 😄

PR #723, opened 2018-04-20
Added the add project button and match styling to wireframe

nakhbari:
Issue: #667 

#### Desktop Screenshot
image

#### Mobile Screenshot
image

### DIFF
--- a/web/app/dashboard/dashboard.component.html
+++ b/web/app/dashboard/dashboard.component.html
@@ -1,44 +1,44 @@
 <div class="fci-dashboard-header">
   <div fci-grid class="fci-dashboard-welcome">
-    <div fci-cell="6">
-      <div class="fci-dashboard-title"> Welcome to fastlane CI! </div>
+    <div fci-cell="12">
+      <div class="fci-dashboard-title"> Welcome to fastlane.ci! </div>
       <div class="fci-dashboard-subtitle">
-        Open source, self-hosted, mobile-optimized CI powered by fastlane brought to you by the fastlane team. </div>
+        Open source, self-hosted, mobile-optimized CI powered by fastlane brought to you by the fastlane team.
+      </div>
+      <button mat-raised-button color="primary">Add New Project</button>
     </div>
   </div>
 </div>
 
 <div fci-grid>
   <div fci-cell="12">
-    <mat-card class="fci-project-table">
-      <mat-progress-spinner mode="indeterminate" *ngIf="isLoading" class="fci-loading-spinner" diameter="30"></mat-progress-spinner>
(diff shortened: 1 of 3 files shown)


web/app/dashboard/dashboard.component.html
```diff
@@ -1,44 +1,44 @@
...
-          <mat-header-cell *matHeaderCellDef> Project Name </mat-header-cell>
-          <mat-cell *matCellDef="let project">
-            <fci-status-icon [status]="project.latestStatus"></fci-status-icon>
-            <span>{{project.name}}</span>
-          </mat-cell>
-        </ng-container>
+    <mat-table class="fci-project-table" [dataSource]="projects">
+      <!-- Name Column -->
+      <ng-container matColumnDef="name">
+        <mat-header-cell *matHeaderCellDef> Project Name </mat-header-cell>
```
taquitos: spaces around `Project Name` probably not intentional, right?

web/app/dashboard/dashboard.component.scss
```diff
@@ -1,29 +1,34 @@
...
     .fci-dashboard-welcome {
-        padding: 112px 0 52px 0;
+        padding: 75px 0 4px 0;
         .fci-dashboard-title {
-            color: rgba(0, 0, 0, 0.87);
             font-family: 'Google Sans', sans-serif;
-            font-size: 30px;
-            font-weight: 400;
+            color: rgba(0, 0, 0, 0.87);
+            font-size: 32px;
```
taquitos: pt instead of px for font sizes

web/app/dashboard/dashboard.component.scss
```diff
@@ -1,29 +1,34 @@
...
+            font-weight: 500;
             line-height: 40px;
-            margin-bottom: 8px;
+            margin-bottom: 12px;
```
taquitos: Would it make sense to use `em` instead of `px` for margins? I've always hated `px` and opt for `em` or `%`.

web/app/dashboard/dashboard.component.html
nakhbari: removed

nakhbari:
@taquitos and I discussed the benefits of `em` vs `pt` vs `%` vs `px`. We don't see a compelling reason to use anything other than `px`.

`px` is specifically demanded in the [Material Design Spec](https://material.io/guidelines/layout/units-measurements.html#units-measurements-density-independent-pixels-dp). It's commonly used by many, so it should be easier for contributors to understand it. Mocks are also using px so replicating them is easiest with `px`.

PR #726, opened 2018-04-22
Add DataService method to GET all available repos

### LINKED ISSUE #716: Add DataService method to GET all available repos
Parent issue #713

minuscorp:
Fixes #716

### DIFF
--- a/app/features-json/project_json_controller.rb
+++ b/app/features-json/project_json_controller.rb
@@ -1,6 +1,7 @@
 require_relative "../shared/authenticated_controller_base"
 require_relative "./view_models/project_summary_view_model"
 require_relative "./view_models/project_view_model"
+require_relative "./view_models/repo_view_model"
 
 module FastlaneCI
   # Controller for providing all data relating to projects
@@ -24,6 +25,17 @@ module FastlaneCI
       project_view_model = ProjectViewModel.new(project: project)
 
       return project_view_model.to_json
+    end
+
+    get "#{HOME}/repos" do
+      provider_credential = check_and_get_provider_credential
+      repos = FastlaneCI::GitHubService.new(provider_credential: provider_credential).repos
+
+      all_repos_view_models = repos.map do |repo|
+        RepoViewModel.new(repo: repo)
+      end
+
+      return all_repos_view_models.to_json
     end
   end
 end
--- a/app/features-json/project_json_controller.rb
+++ b/app/features-json/project_json_controller.rb
@@ -1,7 +1,6 @@
 require_relative "../shared/authenticated_controller_base"
 require_relative "./view_models/project_summary_view_model"
 require_relative "./view_models/project_view_model"
-require_relative "./view_models/repo_view_model"
 
 module FastlaneCI
   # Controller for providing all data relating to projects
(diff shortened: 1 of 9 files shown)


app/features-json/project_json_controller.rb
```diff
@@ -25,5 +26,16 @@ class ProjectJSONController < AuthenticatedControllerBase
 
       return project_view_model.to_json
     end
+
+    get "#{HOME}/repos" do
```
nakhbari: Repos are attached to the configured robot accounts right? Maybe we should put this in a Repo Controller since it's not specific to a project.

web/app/services/data.service.ts
```diff
@@ -24,4 +25,10 @@ export class DataService {
     return this.http.get<ProjectResponse>(url).pipe(
         map((project) => new Project(project)));
   }
+
+  getRepos(): Observable<Repository[]> {
+    const url = `${HOSTNAME}/repos`;
```
nakhbari: i like this path. I don't think it matches what we have in the controller, since it's currently `/projects/repos`. But by moving repos into it's own controller can can have `/repos` 👍

app/features-json/view_models/repo_view_model.rb
```diff
@@ -0,0 +1,23 @@
...
+module FastlaneCI
+  # View model to expose the basic info about repositories.
+  class RepoViewModel
+    include FastlaneCI::JSONConvertible
+
+    # @return [String]
+    attr_reader :full_name
+
+    # @return [String]
+    attr_reader :api_path
```
nakhbari: Do we need the api path yet? After looking at the info required for the add project dialog, i'm thinking that we just need the url and full name for now.

app/features-json/project_json_controller.rb
minuscorp: Yeah! That makes sense! I'll tackle all the review comments today!

app/features-json/repos_json_controller.rb
```diff
@@ -0,0 +1,20 @@
+require_relative "../shared/authenticated_controller_base"
+require_relative "./view_models/repo_view_model"
+
+module FastlaneCI
+  # Controller for providing all data relating to projects
+  class ProjectJSONController < AuthenticatedControllerBase
+    HOME = "/data/repos"
+
+    get HOME.to_s do
```
nakhbari: For my own knowledge, is this best practice for Ruby or `"#{HOME}"`?

app/features-json/repos_json_controller.rb
```diff
@@ -0,0 +1,20 @@
+require_relative "../shared/authenticated_controller_base"
+require_relative "./view_models/repo_view_model"
+
+module FastlaneCI
+  # Controller for providing all data relating to projects
+  class RepositoryJSONController < AuthenticatedControllerBase
+    HOME = "/data/repos"
+
+    get HOME.to_s do
```
nakhbari: For my own knowledge, is this best practice for Ruby or `"#{HOME}"`?

minuscorp: Rubocop: prefer `.to_s` over string interpolation 😆

PR #731, opened 2018-04-23
Added POST endpoint to allow creation of new project

nakhbari:
This works but I keep getting failure when trying to push to the config repo. Not sure if it's due to my poor internet connection. It gets synced later on. Error: https://gist.github.com/nakhbari/694c5c2fff6f3c54073f420c437ba970

### DIFF
--- a/app/features-json/project_json_controller.rb
+++ b/app/features-json/project_json_controller.rb
@@ -7,6 +7,13 @@ module FastlaneCI
   class ProjectJSONController < AuthenticatedControllerBase
     HOME = "/data/projects"
 
+    # Get the payload from the body
+    before do
+      request.body.rewind
+      body = request.body.read
+      @request_payload = JSON.parse(body) unless body.empty?
+    end
+
     get HOME do
       current_provider_credential = check_and_get_provider_credential
       current_user_config_service = self.current_user_config_service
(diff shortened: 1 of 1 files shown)


app/features-json/project_json_controller.rb
```diff
@@ -25,5 +32,80 @@ class ProjectJSONController < AuthenticatedControllerBase
 
       return project_view_model.to_json
     end
+
+    post HOME do
+      REQUIRED_PARAMS = Set["lane", "repo_org", "repo_name", "project_name", "trigger_type"]
+      has_required_params = REQUIRED_PARAMS.subset?(Set.new(@request_payload.keys))
+
+      if has_required_params
```
KrauseFx: Minor thing, if you use
```ruby
raise "something" unless has_required_params`
```
the whole rest of the method can be moved more to the left 🚔

app/features-json/project_json_controller.rb
```diff
@@ -25,5 +32,80 @@ class ProjectJSONController < AuthenticatedControllerBase
...
+        project_name = @request_payload["project_name"]
+        branch = @request_payload["branch"]
+        trigger_type = @request_payload["trigger_type"]
+        hour = @request_payload["hour"]
+        minute = @request_payload["minute"]
+
+        # TODO: Until we make a proper interface to attach JobTriggers to a Project, let's add a manual one for the
+        # selected branch.
+        # TODO: get default branch when there is no branch selected
+        triggers_to_add = [FastlaneCI::ManualJobTrigger.new(branch: branch.nil? ? "master" : branch)]
```
KrauseFx: Nice :+1:

app/features-json/project_json_controller.rb
```diff
@@ -25,5 +32,80 @@ class ProjectJSONController < AuthenticatedControllerBase
 
       return project_view_model.to_json
     end
+
+    post HOME do
+      REQUIRED_PARAMS = Set["lane", "repo_org", "repo_name", "project_name", "trigger_type"]
```
taquitos: this should be lower case

app/features-json/project_json_controller.rb
```diff
@@ -25,5 +32,80 @@ class ProjectJSONController < AuthenticatedControllerBase
...
+        provider_credential = check_and_get_provider_credential
+
+        github_service = FastlaneCI::GitHubService.new(provider_credential: provider_credential)
+
+        selected_repo = github_service.repos.detect do |repo|
+          @request_payload["repo_name"] == repo[:name] &&
+            @request_payload["repo_org"] == repo[:owner][:login]
+        end
+
+        repo_config = GitHubRepoConfig.from_octokit_repo!(repo: selected_repo)
```
KrauseFx: I could imagine the error message is pretty confusing if `selected_repo` is `nil` here, just a guess

app/features-json/project_json_controller.rb
nakhbari: Am i doing constants wrong?

app/features-json/project_json_controller.rb
```diff
@@ -25,5 +32,80 @@ class ProjectJSONController < AuthenticatedControllerBase
...
+
+        # We now have enough information to create the new project.
+        # TODO: add job_triggers here
+        # We shouldn't be blocking manual trigger builds
+        # if we do not provide an interface to add them.
+        project = Services.project_service.create_project!(
+          name: project_name,
+          repo_config: repo_config,
+          enabled: true,
+          platform: lane.split(" ").first,
```
taquitos: we have this logic now in a couple places `lane.split(" ")`
is there a nicer way we could abstract this out?

app/features-json/project_json_controller.rb
taquitos: Yeah, as far as I'm aware, it's not really idiomatic to Ruby to use constants like that. It can get confusing for maintainers because constants are typically defined on a class, so when somebody comes in to read this code and fix a bug, people might see `REQUIRED_PARAMS` in one place, and assume this is defined for the entire class.

app/features-json/project_json_controller.rb
nakhbari: done

app/features-json/project_json_controller.rb
nakhbari: i tried to do a split earlier on to make it cleaner

app/features-json/project_json_controller.rb
nakhbari: done

PR #696, opened 2018-04-18
[WIP] Rename environment_variable_service to dot_keys_variable_service

KrauseFx:
.keys manager is limited and only focuses on the non-user specific ENV variables. For all user managed ones, we're gonna use a new data source + service

More information: https://github.com/fastlane/ci/issues/643
We're gonna use the names `EnvironmentVariable` and `EnvironmentVariableService` for ENV variables that are provided by the user

Issue #643, opened 2018-04-11
Build a system that allows global, project and branch specific settings


KrauseFx:
It looks like for some things we'll have 3 levels of where the value comes from. Examples are: Xcode version, and environment variables. The levels are

- **Global setting**: Default global Xcode installation, which is used unless overwritten by any of the other settings
- **Project-level setting**: A project defines a particular Xcode version, as it wasn't migrated to the new version of Swift yet
- **Branch-specific setting**: The `swift-5-migration` branch need to use the new Xcode beta to be able to compile & run tests

All of the above should be applied to all settings where it makes sense. Out of the top of my head, this might just be Xcode version, and the environment variables 👍 

Existing issues:
- Edit project-specific environment variables: https://github.com/fastlane/ci/issues/441
- Edit the global fastlane.ci environment variables https://github.com/fastlane/ci/issues/439

christoff-1992:
Will this include upload parameters such as,

DELIVER_ITMSTRANSPORTER_ADDITIONAL_UPLOAD_PARAMETERS="-t DAV"

It would be useful to globally configure these as this is environment specific not project.

taquitos:
@christoff-1992 since we use `fastlane` internally, we should get this global behavior for free 🎉 
Our plan is to allow you to also specify any arbitrary environment variable at the project-level, as well as branch-level too.

Did you try using this environment variable and have issues in fastlane.ci?

christoff-1992:
@taquitos - No issues with the environment variable that sounds perfect to me :)

KrauseFx:
So my current thoughts are:

### Global ENV variables

Store them in `environment.json` in the root of the `ci-config` repo

### Project specific ENV variables
We have a few potential solutions:
- `[ci-config-repo]/projects/[project_id]/environment.json`: this will allow us to offer something similar for other types of configuration for the `project-level`. For branch specific options we could just use `[ci-config-repo]/projects/[project_id]/branches/master/environment.json` (store env variables in separate file per branch) or `[ci-config-repo]/projects/[project_id]/branches/environment.json` (store info of all branches in a single file)
- `environment.json` could have 2 keys in the root hash: `global`, `project`, kind of like:
```json
{
  "global": {
    "FASTLANE_DISABLE_YOLO": false
  },
  "project": {
      "[project_id]": {
          "env": {
              "FASTLANE_DISABLE_YOLO": true,
          },
          "branches": {
              "develop": 
[...]

PR #739, opened 2018-04-24
Rename environment_variable_service to dot_keys_variable_service

KrauseFx:
Same change as #696

---

.keys manager is limited and only focuses on the non-user specific ENV variables. For all user managed ones, we're gonna use a new data source + service

More information: #643
We're gonna use the names `EnvironmentVariable` and `EnvironmentVariableService` for ENV variables that are provided by the user


KrauseFx:
Self-approved to keep moving for now 🚔
Test it locally

PR #740, opened 2018-04-24
Update dot_keys file to work with new dotenv

KrauseFx:
This breaking change was introduced with a minor version release with https://github.com/bkeepers/dotenv/commit/7f82f73420bb46a06bb3a3e2052c3b3a1f34a0f5

It seems to only be related to `VARIABLE={NAME}` variables and about if they should be expanded. Since we use this just for `~/.fastlane/ci/.keys` we don't really care about this option


KrauseFx:
Not sure what you mean, this is unrelated to all other PRs from this morning @taquitos.

Issue #439, opened 2018-03-30
Edit the global fastlane.ci environment variables


KrauseFx:
https://github.com/fastlane/ci/blob/master/docs/use_cases/edit_global_env_variables.md

Related https://github.com/fastlane/ci/issues/441

- [x] Use the ENV variables during build time

PR #747, opened 2018-04-24
Add support for management ENV variables for fastlane.ci

KrauseFx:
This is an early WIP

As part of this PR I noticed how much duplicate code we have to write just to cover the basic CRUD operations, maybe something to look into post Beta release - for now this is fine, let's focus on building & shipping features.

- Related to https://github.com/fastlane/ci/issues/643 
- Fixes https://github.com/fastlane/ci/issues/439
- Implements use case [edit_global_env_variables.md](https://github.com/fastlane/ci/blob/master/docs/use_cases/edit_global_env_variables.md)


KrauseFx:
Not yet finished, need to spend some more time on the overhead called CRUD :) editing, deleting, finding existing ENV variables

app/services/data_sources/environment_data_source.rb
```diff
@@ -0,0 +1,24 @@
+module FastlaneCI
+  # Data source for global environment variables for fastlane.ci
+  class EnvironmentDataSource
```
snatchev: I don't love this interface. Ruby already has an interface to the environment using the `ENV` and a hash. Why not have a class extends a Hash or proxy to a hash like this? https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/base.rb#L22

KrauseFx: I used the same approach we use for all other model classes, e.g. https://github.com/fastlane/ci/blob/master/app/services/data_sources/user_data_source.rb, https://github.com/fastlane/ci/blob/master/app/services/config_data_sources/project_data_source.rb and https://github.com/fastlane/ci/blob/master/app/services/data_sources/notification_data_source.rb

snatchev: I suppose that is consistent with the `AbcDataSource` pattern.

But my reaction was surprise that the environment would come from something other than `ENV`, (which already has a well-established interface). Instead it's coming from a file DataSource, much like a regular config file.

Perhaps I don't understand the problem well enough.

KrauseFx: Ha, ok :) That makes sense. Did you have a chance to read https://github.com/fastlane/ci/blob/master/docs/use_cases/edit_global_env_variables.md, in particular the #security section: the goal is to store the environment variables for the user, similar to how you can set env variables on Heroku. This is just the basic implementation, we're gonna have to extend this class to support project and branch specific options also

app/shared/models/environment_variable.rb
```diff
@@ -0,0 +1,40 @@
+require "securerandom"
+
+module FastlaneCI
+  # EnvironmentVariable model representing a single entry of an env variable
+  class EnvironmentVariable
```
KrauseFx: This class has quite some custom code to do on-demand encryption/decryption

spec/services/configuration_respository_service_spec.rb
```diff
@@ -41,6 +41,7 @@
 
       expect(service).to receive(:create_remote_json_file).with("users.json", any_args)
       expect(service).to receive(:create_remote_json_file).with("projects.json")
+      expect(service).to receive(:create_remote_json_file).with("environment_variables.json")
```
KrauseFx: I want to add tests once I also cover https://github.com/fastlane/ci/issues/643 (the project/branch specific ENV variables)

app/services/environment_variable_service.rb
```diff
@@ -0,0 +1,78 @@
...
+    def initialize(environment_variable_data_source: nil)
+      unless environment_variable_data_source.nil?
+        unless environment_variable_data_source.class <= EnvironmentDataSource
+          raise "environment_variable_data_source must be descendant of #{EnvironmentDataSource.name}"
+        end
+      end
+
+      if environment_variable_data_source.nil?
+        # Default to JSONEnvironmentDataSource
+        # TODO: do we need `sample_data` here?
```
KrauseFx: I don't know what this is, taken from `user_service.rb`

app/services/data_sources/json_environment_variable_data_source.rb
```diff
@@ -0,0 +1,106 @@
...
+        @environment = JSON.parse(File.read(environment_file_path)).map do |environment_variable_hash|
+          EnvironmentVariable.from_json!(environment_variable_hash)
+        end
+      end
+    end
+
+    # TODO: this isn't threadsafe
+    def update_environment_variable!(environment_variable: nil)
+      existing = find_environment_variable(environment_variable_key: environment_variable.key)
+      existing.value = environment_variable.value
```
KrauseFx: This is so great, created an issue to migrate all the other data sources also https://github.com/fastlane/ci/issues/761

app/features/build_runner/build_runner.rb
```diff
@@ -200,6 +202,19 @@ def setup_build_specific_environment_variables
       env_mapping[:CI_BRANCH] = env_mapping[:GIT_BRANCH]
       # env_mapping[:CI_PULL_REQUEST] = nil # TODO: do we have the PR information here?
 
+      # Now that we have the CI specific ENV variables, let's go through the ENV variables
+      # the user defined in their configuration
+      Services.environment_variable_service.environment_variables.each do |environment_variable|
```
KrauseFx: This is so awesome, we can use the infrastructure we already build to inject `BUILD_ID`, `CI_BRANCH` and similar ENV variables

app/features/build_runner/build_runner.rb
```diff
@@ -200,6 +202,19 @@ def setup_build_specific_environment_variables
...
+        if env_mapping.key?(environment_variable.key.to_sym)
+          # TODO: this is probably large enough of an issue to use the fastlane.ci
+          #       notification system to show an error to the user
+          logger.error("Overwriting CI specific environment variable of key #{environment_variable.key} - " \
+            "this is not recommended")
+        end
+        env_mapping[environment_variable.key.to_sym] = environment_variable.value
+      end
+
+      # Finally, set all the ENV variables for the given build
```
KrauseFx: Looks like we're not properly cleaning up environment variables if the build crashes (or fails I guess), see https://github.com/fastlane/ci/issues/758 for more information. This is something we have to fix, as we never ever want to leak ENV variables

PR #756, opened 2018-04-24
Unify model classes so that all business logic related to a class resides in one file.

AndrewMcBurney:
Changes models with the `JSONConvertible` mixin to only have one class definiton, in the `/models` folder.


app/services/dot_keys_variable_service.rb
```diff
@@ -31,7 +31,7 @@ def reload_dot_env!
       return unless File.exist?(keys_file_path)
 
       require "dotenv"
-      ENV.update(Dotenv::Environment.new(keys_file_path, false))
+      ENV.update(Dotenv::Environment.new(keys_file_path))
```
AndrewMcBurney: This was causing a failing test case.

AndrewMcBurney: This was the cause of a failing test case.

AndrewMcBurney:
Sounds good to me @minuscorp, I'll make the changes 🙂

AndrewMcBurney:
I'm going to close this PR - I'll make the changes in another PR at another time.

PR #763, opened 2018-04-25
Fix UsersController not using named parameter

KrauseFx:
This probably crashed before, but I guess we're not using `Update user` anywhere yet


AndrewMcBurney:
This was a regression. It worked without the named parameter in the original implementation. We opted to remove the ability to update users from the views in https://github.com/fastlane/ci/pull/627, so this is essentially dead code. I remember having a discussion with @taquitos about keeping the CRUD actions - but I think it was on slack rather than GitHub.

In my initial commit I was going to nuke all the create/update actions, but we opted to keep them.

KrauseFx:
Alright sounds good, let's merge this for now 👍

PR #765, opened 2018-04-25
Load JWTAuth when loading Angular application and authenticate using JWT

### LINKED ISSUE #753: [backend] Login in Web Application
Parent Issue: #749 

- [x] JSON end point to login user and issue a JW Token
- [x] Enforce token auth all Data endpoints (except login of course)

minuscorp:
Fixes #753
captura de pantalla 2018-04-25 a las 20 57 08

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,9 @@ gem "google-cloud-storage", "~> 1.5.0"
 # Manage CI dependencies.
 gem "bundler", "~> 1.16.0"
 
+# Manage JWT authentication tokens.
+gem "jwt"
+
 # fastlane dependencies
 # TODO: point to minimum release instead of GitHub once
 #  we shipped a new release
--- a/Gemfile
+++ b/Gemfile
@@ -38,7 +38,7 @@ gem "google-cloud-storage", "~> 1.5.0"
 gem "bundler", "~> 1.16.0"
 
 # Manage JWT authentication tokens.
-gem "jwt"
+gem "jwt", "~> 2.1.0"
 
 # fastlane dependencies
 # TODO: point to minimum release instead of GitHub once
--- a/app/features-json/jwt_auth.rb
+++ b/app/features-json/jwt_auth.rb
@@ -1,0 +1,30 @@
+require_relative "../services/dot_keys_variable_service"
+
+require "jwt"
+
+module FastlaneCI
+  class JwtAuth
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      options = { algorithm: "HS256", iss: "fastlane.ci" }
+      bearer = env.fetch("HTTP_AUTHORIZATION", "").slice(7..-1)
+      payload, header = JWT.decode(bearer, FastlaneCI.dot_keys.encryption_key, true, options)
+
+      env[:scopes] = payload["scopes"]
+      env[:user] = payload["user"]
+
+      @app.call(env)
+    rescue JWT::DecodeError
+      [401, { "Content-Type" => "text/plain" }, ["A token must be passed."]]
+    rescue JWT::ExpiredSignature
+      [403, { "Content-Type" => "text/plain" }, ["The token has expired."]]
+    rescue JWT::InvalidIssuerError
+      [403, { "Content-Type" => "text/plain" }, ["The token does not have a valid issuer."]]
+    rescue JWT::InvalidIatError
+      [403, { "Content-Type" => "text/plain" }, ['The token does not have a valid "issued at" time.']]
+    end
+  end
+end
--- a/app/features-json/jwt_auth.rb
+++ b/app/features-json/jwt_auth.rb
@@ -13,8 +13,7 @@ module FastlaneCI
       options = { algorithm: "HS256", iss: "fastlane.ci" }
       bearer = env.fetch("HTTP_AUTHORIZATION", "").slice(7..-1)
       payload, = JWT.decode(bearer, FastlaneCI.dot_keys.encryption_key, true, options)
-
-      env[:scopes] = payload["scopes"]
+      
       env[:user] = payload["user"]
 
       @app.call(env)
--- a/app/features-json/jwt_auth.rb
+++ b/app/features-json/jwt_auth.rb
@@ -3,6 +3,7 @@ require_relative "../services/dot_keys_variable_service"
 require "jwt"
 
 module FastlaneCI
+  # API Middleware responsible of authenticate all the requests that uses it.
   class JwtAuth
     def initialize(app)
       @app = app
@@ -11,7 +12,7 @@ module FastlaneCI
     def call(env)
       options = { algorithm: "HS256", iss: "fastlane.ci" }
       bearer = env.fetch("HTTP_AUTHORIZATION", "").slice(7..-1)
-      payload, header = JWT.decode(bearer, FastlaneCI.dot_keys.encryption_key, true, options)
+      payload, = JWT.decode(bearer, FastlaneCI.dot_keys.encryption_key, true, options)
 
       env[:scopes] = payload["scopes"]
       env[:user] = payload["user"]
--- a/app/features-json/login_json_controller.rb
+++ b/app/features-json/login_json_controller.rb
@@ -1,12 +1,12 @@
-require_relative "../../shared/controller_base"
-require_relative "../../services/user_service"
-require_relative "../../services/dot_keys_variable_service"
-
+require_relative "../shared/controller_base"
+require_relative "../services/user_service"
+require_relative "../services/dot_keys_variable_service"
 
 require "jwt"
 require "json"
 
 module FastlaneCI
+  # Controller responsible of handling 
   class LoginJSONController < ControllerBase
     HOME = "/login"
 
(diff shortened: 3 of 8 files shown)


Gemfile
```diff
@@ -37,6 +37,9 @@ gem "google-cloud-storage", "~> 1.5.0"
 # Manage CI dependencies.
 gem "bundler", "~> 1.16.0"
 
+# Manage JWT authentication tokens.
+gem "jwt"
```
KrauseFx: Add a version number here also 😀

fastlane_app.rb
```diff
@@ -48,6 +49,7 @@ class FastlaneApp < Sinatra::Base
     # display of real-time output
     set(:server, "thin")
     if ENV["WEB_APP"]
+      use(FastlaneCI::JwtAuth)
```
nakhbari: does this run on every route? I think we just want this on our API calls right?

minuscorp: As it is inside the `ENV["WEB_APP"]`, it will only be loaded for the Angular application that middleware (to be tested, tho)

minuscorp: You're right, this should be included in every endpoint controller that uses `JwtAuth` instead of there, my bad.

minuscorp: Maybe `AuthenticatedControllerBase` is not useful for our API endpoints, and we should inherit from a custom `JWTAuthenticatedControllerBase` that uses the middleware.

nakhbari: I think ideally that would be the case, but switching to that enforcement right now would break our UI, since we do not have the token being sent yet. We could move the Authentication controller base work into another CL, with the Token verification and keep that open until the frontend is complete.

minuscorp: Yep, I just want to make sure that the `/login` endpoint generates the required token, even if we don't we use it for now. Later on, we can apply the middleware into the controllers when the frontend is finished. Is that ok?

nakhbari: Fine with me. I'll let one of the ruby experts give you the approval 😀

app/features-json/jwt_auth.rb
```diff
@@ -0,0 +1,31 @@
...
+
+module FastlaneCI
+  # API Middleware responsible of authenticate all the requests that uses it.
+  class JwtAuth
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      options = { algorithm: "HS256", iss: "fastlane.ci" }
```
nakhbari: @minuscorp could you expand on what led you to use HS256 instead of RS256. I'm not familiar with the benefits of one over the other. I have heard that RS256 is better. So i'm interested in known what your thoughts are here.

minuscorp: So this was not a completely thought, I saw that in closed-systems was safe to use `HS256` (the provider and the consumer of the JWT are the same source, both know the encryption key): https://stackoverflow.com/a/39239395/4161167 If we're thinking on opening the application and have several encryption keys depending on the `iss` (that's another topic for discussion)

app/features-json/login_json_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+require_relative "../services/user_service"
+require_relative "../services/dot_keys_variable_service"
+
+require "jwt"
+require "json"
+
+module FastlaneCI
+  # Controller responsible of handling the login process using JWT token.
+  class LoginJSONController < ControllerBase
+    HOME = "/login"
```
nakhbari: should be `/api/login`

app/features-json/login_json_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      payload = params
+      payload = JSON.parse(request.body.read).symbolize_keys unless params[:path]
+      email = payload[:email]
+      password = payload[:password]
+      user = Services.user_service.login(email: email, password: password)
+      if user.nil?
+        halt(401)
+      else
+        content_type(:json)
+        { token: token(user) }.to_json
```
nakhbari: can we add a TODO to also send the expiry of the token alongside the token.

app/features-json/login_json_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  class LoginJSONController < ControllerBase
+    HOME = "/login"
+
+    post HOME.to_s do
+      # Allow this endpoint to be requested with { Content-Type: application/json }
+      payload = params
+      payload = JSON.parse(request.body.read).symbolize_keys unless params[:path]
+      email = payload[:email]
+      password = payload[:password]
+      user = Services.user_service.login(email: email, password: password)
```
nakhbari: let's make this `username` instead of `email`

taquitos: Early on we decided that usernames will always be email to reduce complexity. I don't know if it's a good idea to change that. I'd like to hear more thoughts on this.

nakhbari: I don't particularly have a strong opinion on this. I just know that username gives us more flexibility. If we want to stick with emails that cool with me 👍

app/features-json/login_json_controller.rb
minuscorp: JWT has an expiry field [[1]](https://tools.ietf.org/html/rfc7519#section-4.1.4) that is already being sent (when you decode the token of course!) (now it lasts for one hour, again, can be changed 😄 )

app/features-json/login_json_controller.rb
taquitos: Cool, let's stick with email. We can open it up in the future to usernames. Part of our onboarding involves grabbing the GitHub account info and creating a user from their email. We don't default to an empty box.

app/features-json/login_json_controller.rb
taquitos: Does that mean the user needs to login after 1 hour or just a token refresh needed? If they need to login, we shouldn't expire their login but maybe every 6 months. We should rotate the auth token though, maybe daily, or weekly.

minuscorp: JWT does not provide a refresh authentication token, if we want to support that, we should use OAuth on top of JWT in order to provide the `refresh_token` to the client. But as a matter of now, I can just make the expiration time like you said, 6 months long. Another thing, the key used to encrypt the token should be known for both the server and the client, (using the environment variable) how we manage that key rotation if not only the server knows how's the key rotating?

nakhbari: @minuscorp Could you explain more on why the client needs to be able to decrypt the token?

minuscorp: To be able to decode the payload of the token (i.e., the user as you can see in the image of the PR)

minuscorp:
From the discussion with @nakhbari: 

1. JWT Token's payload is **not being encrypted at all** for now, the `key` is used to make sure no one can modify the payload of the token without the backend noticing it (MITM attacks or request spoofing).
2. Expire time of the token has been upgraded from **1 hour to 30 days**.
3. Under the `user` key of the payload, **we're sending back and forward the `User` primary key, so the backend has always a resource of checking that the relationships between the requested data and the requester are correct**.
4. Scopes are being removed for now, maybe in the future we want to make kind of different user roles and we can make it into the payload without any problem.

captura de pantalla 2018-05-01 a las 22 30 11

I don't know if I'm missing something, if I do, make sure to edit this comment 😄!

nakhbari:
Noting some high level points about JWTs for posterity:
- JWT (JSON web token) is usually encrypted with RS256 or HS256. HS256 only requires the secret in order to **verify** the token. The actual token is encoded in base64 and can be decoded without the secret.
- Any data can be passed along inside the payload of the token. Though we can pack anything in there, it's a good pattern to keep it constraint to the bare minimum. Any extra data increases the size of the token. Since this token is passed along with every request in the HTTP request header, an increase in size affects the latency of the request. So the smaller the better.

app/features-json/login_json_controller.rb
nakhbari: Let's make sure to change this before merging.

app/features-json/login_json_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+
+require "jwt"
+require "json"
+
+module FastlaneCI
+  # Controller responsible of handling the login process using JWT token.
+  class LoginJSONController < ControllerBase
+    include JSONController
+
+    HOME = "/login"
```
nakhbari: should be `/api/login`

app/features-json/login_json_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+require "json"
+
+module FastlaneCI
+  # Controller responsible of handling the login process using JWT token.
+  class LoginJSONController < ControllerBase
+    include JSONController
+
+    HOME = "/api"
+
+    post "#{HOME}/login" do
```
nakhbari: since this is the login controller, i do think that HOME = "/api/login"
we should constrain it to that namespace for now, expanding it if we need to in the future.

app/features-json/login_json_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    private
+
+    def token(user)
+      JWT.encode(payload(user), FastlaneCI.dot_keys.encryption_key, "HS256")
+    end
+
+    def payload(user)
+      {
+        # One month expire time is more than safe for now.
+        exp: Time.now.to_i + (60 * 60 * 24 * 30),
```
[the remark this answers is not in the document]
minuscorp: Login call should be done again

Issue #441, opened 2018-03-30
Edit project-specific environment variables


KrauseFx:
https://github.com/fastlane/ci/blob/master/docs/use_cases/edit_project_env_variables.md

Related https://github.com/fastlane/ci/issues/439

KrauseFx:
Fixed with https://github.com/fastlane/ci/pull/768

PR #768, opened 2018-04-27
Add support for project specific settings

KrauseFx:
- Fixes https://github.com/fastlane/ci/issues/643
- Fixes https://github.com/fastlane/ci/issues/441

This PR does the following
- Add support for project specific ENV variables. This is a crucial feature, which allows the user to set certain ENV variables, like the `FASTLANE_USER`, `APP_SCHEME`, `KEYCHAIN_PATH` and similar
- While those values can already be set on a system-wide level, being able to override the global ones per project is critical, and is made possible with this PR
- This required some custom code when reading/writing projects to ignore the `environment_variables` key of `Project` and use a separate `environment_variables.json` in the project's folder instead
- When editing a project, just using the regular `service.projects = ...` setter will automatically also trigger saving the ENV variables for the project 👍 
- This PR will also inject the project specific ENV variables into the build, and potentially overrides any existing ENV variables
- Just for debugging purposes, the ENV variables are all rendered in the project edit screen also
- We don't access the EnvironmentVariableService, but access the data source directly instead (see comment I added to this PR). I have the feeling this will allow us to be very flexible in the future, where a user might want to store ENV variables on a different source depending on the project.

screenshot 2018-04-30 06 02 43

---

screenshot 2018-04-30 06 02 38

### DIFF
--- a/app/services/config_data_sources/json_project_data_source.rb
+++ b/app/services/config_data_sources/json_project_data_source.rb
@@ -151,12 +151,15 @@ module FastlaneCI
 
     def projects=(projects)
       JSONProjectDataSource.projects_file_semaphore.synchronize do
-        # We have to ignore the `environment_variables` instance variable, as it's stored in a separate file
-        json_data = JSON.pretty_generate(projects.map do |project|
-          # While we iterate, also store the project specific configuration files
+        # First store the project specific configuration files, for every project
+        projects.each do |project|
           environment_variable_data_source = project_specific_environment_variables_data_source(project: project)
           environment_variable_data_source.environment_variables = project.environment_variables
+        end
 
+        # now store things into the actual projects.json
+        # We have to ignore the `environment_variables` instance variable, as it's stored in a separate file
+        json_data = JSON.pretty_generate(projects.map do |project|
           project.to_object_dictionary(ignore_instance_variables: [:@environment_variables])
         end)
         File.write(git_repo.file_path("projects.json"), json_data)
--- a/app/services/config_data_sources/json_project_data_source.rb
+++ b/app/services/config_data_sources/json_project_data_source.rb
@@ -120,14 +120,7 @@ module FastlaneCI
         saved_projects = JSON.parse(File.read(path)).map do |project_json|
           project = Project.from_json!(project_json)
 
-          # Now load the project specific files here
-          # that are in
-          #
-          #   `projects/[project_id]/[file].json
-          #
-          # more information on https://github.com/fastlane/ci/issues/643
-          project_specific_path = File.join(git_repo.file_path("projects"), project.id)
-          environment_variable_data_source = JSONEnvironmentDataSource.create(project_specific_path)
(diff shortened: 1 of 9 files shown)


app/services/config_data_sources/json_project_data_source.rb
```diff
@@ -119,6 +119,17 @@ def projects
...
+
+          # Now load the project specific files here
+          # that are in
+          #
+          #   `projects/[project_id]/[file].json
+          #
+          # more information on https://github.com/fastlane/ci/issues/643
+          project_specific_path = File.join(git_repo.file_path("projects"), project.id)
+          environment_variable_data_source = JSONEnvironmentDataSource.create(project_specific_path)
+          project.environment_variables = environment_variable_data_source.environment_variables
```
KrauseFx: Would love @taquitos' input on this - currently I'm creating a `JSONEnvironmentDataSource` for each project within the `json_project_data_source`, which would allow us to have a different data source for ENV variables (secrets) depending on the project. I wasn't sure if I should access the `EnvironmentVariableService` here or the ``JSONEnvironmentDataSource` directly, what's the approach here?

taquitos: This is a great idea! I think datasource to datasource communication is best here. The thinking is that you should be able to bundle the datasources together into a package with no dependences on the services. Thinking of it as a layer.

app/services/config_data_sources/json_project_data_source.rb
```diff
@@ -147,7 +151,15 @@ def job_triggers_from_hash_array(job_trigger_array: nil)
 
     def projects=(projects)
       JSONProjectDataSource.projects_file_semaphore.synchronize do
-        File.write(git_repo.file_path("projects.json"), JSON.pretty_generate(projects.map(&:to_object_dictionary)))
+        # We have to ignore the `environment_variables` instance variable, as it's stored in a separate file
+        json_data = JSON.pretty_generate(projects.map do |project|
```
taquitos: I would prefer to have this be a 2 step process, first is `projects.map`, then second be the `JSON.pretty_generate`

app/services/data_sources/json_environment_variable_data_source.rb
```diff
@@ -21,6 +21,9 @@ class << self
     end
 
     # can't have us reading and writing to a file at the same time
+    # TODO: currently we just have this global lock
```
taquitos: 💯

app/services/config_data_sources/json_project_data_source.rb
KrauseFx: Yeah, that's valid, will update tomorrow

app/services/data_sources/json_environment_variable_data_source.rb
KrauseFx: I actually created an issue for this also https://github.com/fastlane/ci/issues/771

Issue #713, opened 2018-04-20
Add New Project Button leads to Add Project Dialog


nakhbari:
Parent Issue: #644 

#### Pre-requisite
- [x] Add button to create new project on dashboard #667

#### Work (can do data items before frontend)
- [x] Add Endpoint to `ProjectJsonController` which handles project post requests and creates a new project. #724
- [x] Add Endpoint to GET available Git repos
- [x] Add `DataService` method to POST new project. #750 
- [x] Add `DataService` method to GET all available repos. #716 
- [x] Endpoint that gets a project's lanes

#### Frontend
- [x] [Create Angular material Dialog](https://material.angular.io/components/dialog/overview)
- [x] [Add select and other form components](https://material.angular.io/components/select/overview)
- [x] throw in the dialog going to fullscreen on mobile
- [x] Add Tests
- [x] Wire the close buttons to close the dialog
- [x] Wire the add project button to add project
- [x] Enable the add project endpoint to just use the repo full name
- [x] Add form data validation and errors 
- [x] Get Lane data from backend
- [x] Gather form inputs and use the `DataService` post method to create the project
- [ ] Add Branch form
- [ ] Reload lanes when branch is changed
- [ ] Add invalid input styling

#### Wireframe
![image](https://user-images.githubusercontent.com/2754461/39064704-4aa6fbb0-4484-11e8-806c-f45b9bf1f5f3.png)

nakhbari:
Gonna pick up the dialog frontend work.

PR #769, opened 2018-04-27
Add project dialog and modularize dashboard

nakhbari:
Added new Dialog for the Add Project flow.
Issue: #713

Some things to note here:
- We should try to be more "modular" having a module at each route/page will allow Angular to do some lazy loading. AKA Performance boost ⚡️ 
- `entryComponents` is used here to denote Components that won't be used in templates and only referenced programmatically. So since DashboardComponent is used in the RouterModule and not actually in any templates we can declare it there instead of the `exports`. 
- `exports` should be used to make components visible to any module that imports it.
- Eventually i'll break out the module some more, instead of using Shared Modules. We can start using Shared Modules when we really know what's being shared.

#### Screenshot
image

### DIFF
--- a/web/app/dashboard/dashboard.component.ts
+++ b/web/app/dashboard/dashboard.component.ts
@@ -1,7 +1,9 @@
 import {Component, OnInit} from '@angular/core';
+import {MAT_DIALOG_DATA, MatDialog, MatDialogRef} from '@angular/material';
 
 import {ProjectSummary, ProjectSummaryResponse} from '../models/project_summary';
 import {DataService} from '../services/data.service';
+import {AddProjectDialogComponent} from './add-project-dialog/add-project-dialog.component';
 
 @Component({
   selector: 'fci-dashboard',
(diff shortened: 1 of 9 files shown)


web/app/dashboard/dashboard.component.ts
```diff
@@ -14,12 +16,24 @@ export class DashboardComponent implements OnInit {
...
   ngOnInit() {
     this.dataService.getProjects().subscribe((projects) => {
       this.projects = projects;
       this.isLoading = false;
     });
   }
+
+  openAddProjectDialog() {
+    const dialogRef =
+        this.dialog.open(AddProjectDialogComponent, {width: '250px', data: {}});
```
taquitos: I feel like the default line width could be expanded to 120 before wrapping like this, but I'm not super familiar with web coding styling.

nakhbari: Yeah something i wanted to look into was encapsulating Google codestyle requirements in tslint. So we can just autoformat. But i'm punting that until later on. Hope that's OK.

taquitos: Yeah, sounds good. We can ignore style stuff and then autoformat later 🌮

PR #772, opened 2018-04-30
Remove unused `repos.json` file accessor

KrauseFx:


KrauseFx:
From what I understand, https://github.com/fastlane/ci/pull/459 will make more methods unnecessary that are in this file 👍

Issue #767, opened 2018-04-26
Make new frontend (Angular) default frontend instead of erb


taquitos:
We've made a lot more progress on our Angular hotness. Let's switch over to using that by default now, and fleshing out any missing parts!
- [ ] Add env var to opt out if necessary

PR #776, opened 2018-04-30
Make new frontend (Angular) default frontend instead of erb

nakhbari:
Fixes: #767 

Load the Angular app by default, and launch with ERB client when the `ERB_CLIENT` env key is set.

One drawback here is the lack of login implementation in the Angular app. This switch should motivate us to get these features done.

### DIFF
--- a/fastlane_app.rb
+++ b/fastlane_app.rb
@@ -47,7 +47,7 @@ module FastlaneCI
     # which is required to support web socket streams for the
     # display of real-time output
     set(:server, "thin")
-    if ENV["ERB_CLIENT"]
+    if ENV["FASTLANE_CI_ERB_CLIENT"]
       get "/" do
         if session[:user]
           redirect("/dashboard_erb")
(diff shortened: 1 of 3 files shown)


fastlane_app.rb
```diff
@@ -47,24 +47,24 @@ class FastlaneApp < Sinatra::Base
...
     # display of real-time output
     set(:server, "thin")
-    if ENV["WEB_APP"]
-      # Anything except a data route
-      get %r{/(?!data.*).*} do
-        # Use Angular Web App instead
-        send_file(File.join("public", ".dist", "index.html"))
-      end
-    else
+    if ENV["ERB_CLIENT"]
```
taquitos: let's make it "FASTLANE_CI_ERB_CLIENT"

nakhbari: done

taquitos:
🚢

KrauseFx:
Don't forget to delete the branches 👍

Issue #752, opened 2018-04-24
[frontend] Login in Web Application


nakhbari:
Parent Issue: #749 

- [x] Create Login Page Stub
- [x] Create Auth service that handles server login
- [x] Store JWT in local storage
- [x] Attach JWT to all outgoing HTTP requests
- [x] Add Listener to all data service methods to redirect to login if `401` is returned. Maybe `HttpInterceptor`?
- [x] Store expiry date of the token in order to determine if the user's token is still valid.

minuscorp:
> Check user authenticated on page load?
I've read that a good approach is to prompt for user login as a modal form (just as we do for creating a new project) whenever the authentication is wrong for certain request. That way we always make sure we're not redirecting the user to a login page whenever is not necessary (i.e., user is not logged in at all, it should be redirected to `/login`) and the user can continue where he left everything after the token being refreshed.

nakhbari:
Done, added a separate issue for account creation https://github.com/fastlane/ci/issues/1011

PR #781, opened 2018-04-30
Add Auth service which calls backend to login

nakhbari:
Issue: #752 

- [x] Create Auth service that handles server login
- [x] Store JWT in local storage

### DIFF
--- a/web/app/services/auth.service.spec.ts
+++ b/web/app/services/auth.service.spec.ts
@@ -1,0 +1,84 @@
+import {HttpClientTestingModule, HttpTestingController} from '@angular/common/http/testing';
+import {inject, TestBed} from '@angular/core/testing';
+import {mockLoginResponse} from '../common/test_helpers/mock_login_data';
+import {AuthService, LoginRequest, LoginResponse} from './auth.service';
+
+const FAKE_LOGIN_REQUEST: LoginRequest = {
+  username: 'tacoRocat',
+  password: 'wholetthedogsout'
+};
+
+describe('AuthService', () => {
+  let mockHttp: HttpTestingController;
+  let authService: AuthService;
+  let mockStorage = {};
+
+  // TODO: generalize this as a test helper
+  const mockLocalStorage = {
+    getItem: (key: string): string => {
+      return key in mockStorage ? mockStorage[key] : null;
+    },
+    setItem: (key: string, value: string) => {
+      mockStorage[key] = `${value}`;
+    },
+    removeItem: (key: string) => {
+      delete mockStorage[key];
+    },
+    clear: () => {
+      mockStorage = {};
+    }
+  };
+
+  beforeEach(() => {
+    spyOn(localStorage, 'getItem').and.callFake(mockLocalStorage.getItem);
+    spyOn(localStorage, 'setItem').and.callFake(mockLocalStorage.setItem);
+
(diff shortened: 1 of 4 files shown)


web/app/services/auth.service.spec.ts
```diff
@@ -0,0 +1,84 @@
+import {HttpClientTestingModule, HttpTestingController} from '@angular/common/http/testing';
+import {inject, TestBed} from '@angular/core/testing';
+import {mockLoginResponse} from '../common/test_helpers/mock_login_data';
+import {AuthService, LoginRequest, LoginResponse} from './auth.service';
+
+const FAKE_LOGIN_REQUEST: LoginRequest = {
+  username: 'tacoRocat',
```
minuscorp: i think the parameter is `email`, but I can change it in the backend. Also, how are you sending this parameters? (`json/application`, `form-data`?)

web/app/services/auth.service.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    mockLocalStorage.clear();
+  });
+
+  it('should should make login request to backend', () => {
+    let loginResponse: LoginResponse;
+    authService.login(FAKE_LOGIN_REQUEST).subscribe((response) => {
+      loginResponse = response;
+    });
+
+    const loginRequest = mockHttp.expectOne('/data/login');
```
minuscorp: it should be just  `/login` (?)

web/app/services/auth.service.ts
```diff
@@ -0,0 +1,48 @@
+import {HttpClient} from '@angular/common/http';
+import {Injectable} from '@angular/core';
+import * as moment from 'moment';
+import {Observable} from 'rxjs/Observable';
+import {shareReplay, tap} from 'rxjs/operators';
+import {LocalStorageKeys} from '../common/constants';
+
+// Auth server is currently locally hosted.
+const HOSTNAME = '/data';
```
minuscorp: Yeah, in the backend is not under `/data` path, is just `/login`

web/app/services/auth.service.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    expect(loginRequest.request.body).toBe(FAKE_LOGIN_REQUEST);
+    loginRequest.flush(mockLoginResponse);
+
+    expect(loginResponse.token).toBe('12345');
+  });
+
+  it('should store the token in local storage', () => {
+    authService.login(FAKE_LOGIN_REQUEST).subscribe();
+
+    const loginRequest = mockHttp.expectOne('/data/login');
```
minuscorp: dito

web/app/services/auth.service.spec.ts
minuscorp: So we should expose the whole json API under the `/api` prefix?

nakhbari: I'm good with that. Let's start with `/login`. Leave the rest in `/data` for now

web/app/services/auth.service.spec.ts
nakhbari: this would be a part of the HTTP request's body.

web/app/services/auth.service.spec.ts
nakhbari: We can change this in the front-end once the backend is ready

PR #782, opened 2018-04-30
Attach Auth token to header on every HTTP request

nakhbari:
Issue: #752 

Depends on #781 

- [x] Attach JWT to all outgoing HTTP requests

### DIFF
--- a/web/app/services/auth.interceptor.ts
+++ b/web/app/services/auth.interceptor.ts
@@ -1,0 +1,27 @@
+import {HttpEvent, HttpHandler, HttpInterceptor, HttpRequest} from '@angular/common/http';
+import {Injectable} from '@angular/core';
+import {Observable} from 'rxjs/Observable';
+import {LocalStorageKeys} from '../common/constants';
+
+/**
+ * Intercepts the outgoing HTTP requests and attaches the stored JSON web token
+ * to the HTTP header as an Auth Bearer. This saves us from having to attach the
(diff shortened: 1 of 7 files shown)


web/app/services/auth.interceptor.ts
```diff
@@ -0,0 +1,27 @@
...
+@Injectable()
+export class AuthInterceptor implements HttpInterceptor {
+  intercept(req: HttpRequest<any>, next: HttpHandler):
+      Observable<HttpEvent<any>> {
+    const token = localStorage.getItem(LocalStorageKeys.AUTH_TOKEN);
+
+    // Attach auth token to request if it exists. Otherwise, handle it normally.
+    if (token) {
+      const authorizedRequest = req.clone(
+          {headers: req.headers.set('Authorization', `Bearer ${token}`)});
```
minuscorp: might receive always a `403` when the token has expired, maybe this `Interceptor` should try to fetch a new token?

nakhbari: eventually i'd like to store the expiry of the token to determine if the user is still logged in. If the token is expired we should block the requests, until they log in.

taquitos: @nakhbari let's create an issue to track this idea 👍

nakhbari: it's one of the subtasks in #752

Issue #445, opened 2018-03-30
fastlane.ci update flow


KrauseFx:
https://github.com/fastlane/ci/blob/master/docs/use_cases/update_fastlane_ci_via_ruby_gem.md

KrauseFx:
Yes, I guess this is something we need to solve before a beta release

KrauseFx:
Working on this now 👍

KrauseFx:
This is implemented now, a few pieces are missing, in particular asking the user if they want to enable auto-update. Also we have to store the user's selection, so we'd need https://github.com/fastlane/ci/issues/870 as well

PR #793, opened 2018-05-02
Add UpdateFastlaneCIService class

KrauseFx:
This PR is ready to be merged - however UpdateFastlaneCIService needs some more updating as soon as we decided on how exactly we're gonna distribute fastlane.ci via RubyGems. In particular how to launch up the server is going to be relevant

As part of UpdateFastlaneCIService, we'll also need some code to check if any background workers or build workers are still running. I left the code empty for now, as we haven't yet finished our build workers setup, and we'd end up re-implementing things if we wrote it today

- Related to [update_fastlane_ci_via_ruby_gem.md](https://github.com/fastlane/ci/blob/master/docs/use_cases/update_fastlane_ci_via_ruby_gem.md) and https://github.com/fastlane/ci/issues/445

### DIFF
--- a/app/services/update_fastlane_ci_service.rb
+++ b/app/services/update_fastlane_ci_service.rb
@@ -65,7 +65,9 @@ module FastlaneCI
       end
 
       # suppress updater output - very noisy
-      Gem::DefaultUserInteraction.ui = Gem::SilentUI.new
+      unless ENV["FASTLANE_CI_VERBOSE"]
+        Gem::DefaultUserInteraction.ui = Gem::SilentUI.new
+      end
 
       update_needed.each do |tool_info|
         tool = tool_info[0]
--- a/app/services/update_fastlane_ci_service.rb
+++ b/app/services/update_fastlane_ci_service.rb
@@ -1,0 +1,93 @@
+require_relative "../shared/logging_module"
+require_relative "../services/services"
+
+require "rubygems/spec_fetcher"
+require "rubygems/command_manager"
+
+module FastlaneCI
+  # Self-update fastlane.ci, this is called either on manual
+  # user action, or through a background worker
+  class UpdateFastlaneCIService
+    include FastlaneCI::Logging
+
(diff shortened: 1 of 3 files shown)


app/services/update_fastlane_ci_service.rb
```diff
@@ -0,0 +1,93 @@
...
+      highest_versions = updater.highest_installed_gems.keep_if { |key| tools_to_update.include?(key) }
+      update_needed = updater.which_to_update(highest_versions, tools_to_update)
+
+      if update_needed.count == 0
+        logger.info("Nothing to update for fastlane.ci")
+        return
+      end
+
+      # suppress updater output - very noisy
+      Gem::DefaultUserInteraction.ui = Gem::SilentUI.new
```
taquitos: Maybe we should suppress by default, but if we have debug enabled, it should be verbose.

app/workers/check_for_fastlane_ci_update_worker.rb
```diff
@@ -20,13 +20,23 @@ def initialize
...
+
+      # Auto-update enabled code:
+      # FastlaneCI::Services.update_fastlane_ci_service.update_fastlane_ci
+
+      # Manual update enabled code:
+      Services.notification_service.create_notification!(
+        priority: Notification::PRIORITIES[:normal],
+        name: "New version of fastlane.ci available",
+        message: "Please open the fastlane.ci system settings, and hit the \"Update fastlane\" button",
+        details: "Please open the fastlane.ci system settings, and hit the \"Update fastlane\" button"
```
taquitos: Notification details are meant to provide extra context around the message. If no extra context is available, we can just pass nil.

Here's an example of how `details` could be used: https://github.com/fastlane/ci/blob/master/app/shared/models/git_repo.rb#L178

KrauseFx: Got it, thanks, that's super useful!

PR #794, opened 2018-05-02
Login Page logs the user in

nakhbari:
Issue: #752 

Wire up the login page to actually try to login.

### DIFF
--- a/web/app/services/auth.service.ts
+++ b/web/app/services/auth.service.ts
@@ -6,14 +6,14 @@ import {shareReplay, tap} from 'rxjs/operators';
 import {LocalStorageKeys} from '../common/constants';
 
 // Auth server is currently locally hosted.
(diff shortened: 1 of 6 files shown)


web/app/services/auth.service.ts
```diff
@@ -6,14 +6,14 @@ import {shareReplay, tap} from 'rxjs/operators';
 import {LocalStorageKeys} from '../common/constants';
 
 // Auth server is currently locally hosted.
-const HOSTNAME = '/data';
+const HOSTNAME = '/api';
```
taquitos: `HOSTNAME` variable probably needs to be renamed to something like `API_ROOT`, right?

nakhbari: that's a good point. HOSTNAME in the future, i'm envisioning being a config var you can adjust when spinning the server up.

PR #800, opened 2018-05-03
Add verification of proper setup of build runner

KrauseFx:
This will save us debugging time in the future

### DIFF
--- a/app/features/build_runner/fastlane_build_runner.rb
+++ b/app/features/build_runner/fastlane_build_runner.rb
@@ -39,6 +39,10 @@ module FastlaneCI
 
     # completion_block is called with an array of artifacts
     def run(new_line_block:, completion_block:)
+      if lane.nil?
+        raise "Before calling `.run` on #{self}, you have to call `.setup` to finish preparing the BuildRunner"
+      end
+
       artifacts_paths = [] # first thing we do, as we access it in the `ensure` block of this method
       require "fastlane"
 


app/features/build_runner/fastlane_build_runner.rb
```diff
@@ -39,6 +39,10 @@ def setup(parameters: nil)
 
     # completion_block is called with an array of artifacts
     def run(new_line_block:, completion_block:)
+      if lane.nil?
+        raise "Before calling `.run` on #{self}, you have to call `.setup` to finish preparing the BuildRunner"
```
taquitos: Is there anything we should do the with the `completion_block`? Does anything rely on it being called?

KrauseFx: `completion_block` is already a required parameter it seems? Note that we check if `self.lane` was defined, which is set as part of `setup`, so we can ensure the setup was properly executed.

KrauseFx: Also `run` is called from the parent class only, not from the outside world it seems, coming from calling `start` from the `BuildRunner` object

KrauseFx:
Ready to be merged from my side 👍

KrauseFx:
Gonna trigger that build again

AndrewMcBurney:
Hey @KrauseFx, I reran the build manually, and the checks are ✅. I think we're good to :shipit:

KrauseFx:
Beautiful, thanks @AndrewMcBurney 👍

PR #803, opened 2018-05-03
Add Project dialog makes the add project call

nakhbari:
issue: #769

### DIFF
--- a/app/features-json/project_json_controller.rb
+++ b/app/features-json/project_json_controller.rb
@@ -45,8 +45,7 @@ module FastlaneCI
       github_service = FastlaneCI::GitHubService.new(provider_credential: provider_credential)
 
       selected_repo = github_service.repos.detect do |repo|
-        @request_payload["repo_name"] == repo[:name] &&
-          @request_payload["repo_org"] == repo[:owner][:login]
+        @request_payload["repo_name"] == repo[:full_name]
       end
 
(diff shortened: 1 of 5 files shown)


app/features-json/project_json_controller.rb
```diff
@@ -45,8 +45,7 @@ class ProjectJSONController < AuthenticatedControllerBase
       github_service = FastlaneCI::GitHubService.new(provider_credential: provider_credential)
 
       selected_repo = github_service.repos.detect do |repo|
-        @request_payload["repo_name"] == repo[:name] &&
-          @request_payload["repo_org"] == repo[:owner][:login]
+        @request_payload["repo_name"] == repo[:full_name]
```
KrauseFx: It looks like in `project_controller`, we still find the repo by comparing the name and the owner. Are you sure this change works? If so, let's also update `project_controller` to make sure the code base is consistent.

Do I understand correctly that `repo[:name]` would be `fastlane/fastlane` in this case?

nakhbari: IIRC repo name would be like `ci`. full name would be `fastlane/ci`

taquitos: Yup, we're tied more to GitHub's description of a repo here, they think of repo as the name, like `fastlane` whereas the full name is `fastlane/fastlane` or `taquitos/fastlane` for my fork.

PR #804, opened 2018-05-04
Don't `delete` user at `user_index`, simply `delete` the record.

### LINKED ISSUE #761: Remove unnecessary complexity in JSONUserDataSource
https://github.com/fastlane/ci/blob/8f014bcfe5515580a6870ec9a354dfda4117f6f4/app/services/data_sources/json_user_data_source.rb#L190-L201

The above method is used to update and to delete a user

https://github.com/fastlane/ci/blob/8f014bcfe5515580a6870ec9a354dfda4117f6f4/app/services/data_sources/json_user_data_source.rb#L132-L146

However in Ruby, we don't need to know the index, we can just directly call `array_name.delete(object)`. 

In combination with https://github.com/fastlane/ci/issues/744, we don't even have to check if the object exists as part of the `JSONUserDataSource`. 

Instead we could just replace all of this with something as simple as 

```ruby
users.delete(user)
```

I did the same for the environment variable service with https://github.com/fastlane/ci/pull/747 and it works 👍

AndrewMcBurney:
Closes #761.

### DIFF
--- a/app/services/data_sources/json_user_data_source.rb
+++ b/app/services/data_sources/json_user_data_source.rb
@@ -146,14 +146,12 @@ module FastlaneCI
     end
 
     def delete_user!(user: nil)
-      user_index, existing_user = find_user_index_and_existing_user(user: user)
-
-      if existing_user.nil?
+      if find_user(id: user.id).nil?
         logger.debug("Couldn't delete user #{user.email} because they don't exist")
         raise "Couldn't delete user #{user.email} because they don't exist"
       else
-        users.delete_at(user_index)
-        logger.debug("Deleted user #{existing_user.email}, writing out users.json to #{user_file_path}")
+        users.delete(user)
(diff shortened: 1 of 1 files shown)


app/services/data_sources/json_user_data_source.rb
```diff
@@ -146,14 +146,12 @@ def update_user!(user: nil)
     end
 
     def delete_user!(user: nil)
-      user_index, existing_user = find_user_index_and_existing_user(user: user)
```
KrauseFx: From what I understand, we can remove this function completely, once we also update `update_user!` or anywhere else where we use `find_user_index_and_existing_user` 👍

AndrewMcBurney: Hey @KrauseFx , I updated the `update_user` method, and added specs for it. Could you let me know how you feel about the new implementation? Thanks! 💯

PR #805, opened 2018-05-04
Favor `max_by` instead of `sort_by`, `last`.

AndrewMcBurney:
screen shot 2018-05-03 at 10 51 35 pm


AndrewMcBurney:
Yes 🔮 :joy:

PR #808, opened 2018-05-04
[Proposal] Fix crash in `GitRepo#switch_to_git_fork`

KrauseFx:
I'm not sure if that's what we wanted to do, and also wasn't able to test this, but seems like this method didn't work before. Advice appreciated :+1:


taquitos:
Yeah, I refactored the logic to stop using this method, but kept it in just in-case we decided to use it again.

KrauseFx:
Thanks for merging 👍

PR #809, opened 2018-05-04
[WIP] More build metadata

KrauseFx:
Open tasks
- [ ] Persisting of the nested object (`GitForkConfig`)
- [ ] Loading of `GitForkConfig`
- [ ] Showing of branch information in build list
- [ ] Remove `sha` attribute of `Build`
- [ ] More testing
- [ ] Write unit tests

### DIFF
--- a/app/services/code_hosting/github_open_pr.rb
+++ b/app/services/code_hosting/github_open_pr.rb
@@ -1,11 +1,29 @@
 module FastlaneCI
   # Encapsulates the data we need to know how to checkout a branch from a fork
   # We can probably use this for all the things, though.
+
+  # Example data
+  # => #<FastlaneCI::GitHubOpenPR:0x007f9b65c20990
+  #         @branch="joshdholtz-allow-xcodebuild-build-settings-to-use-inherit",
+  #         @clone_url="https://github.com/fastlane/fastlane.git",
+  #         @current_sha="a0a97c934174474fe7b8c87837eb0176f1723d1d",
+  #         @number=12443,
+  #         @repo_full_name="fastlane/fastlane">
+  #
   class GitHubOpenPR
+    # @return [String]: The URL to clone the repo for this PR
     attr_reader :clone_url
+
+    # @return [String]: The branch that is used for this particular PR
     attr_reader :branch
+
+    # @return [String]: The full nume (slug) of the repo (e.g. "fastlane/ci")
     attr_reader :repo_full_name
+
+    # @return [Number]: The number of this particular GitHub PR
     attr_reader :number
+
+    # @return [String]: The SHA for the most recent commit for this PR
     attr_reader :current_sha
 
     def initialize(current_sha:, branch:, repo_full_name:, number:, clone_url:)
--- a/app/shared/models/build.rb
+++ b/app/shared/models/build.rb
@@ -54,6 +54,11 @@ module FastlaneCI
     # TODO: We currently don't use/store/support parameters (yet) https://github.com/fastlane/ci/issues/783
     attr_accessor :parameters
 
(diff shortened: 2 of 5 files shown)


app/features/build/views/build.erb
```diff
@@ -31,6 +31,23 @@
       </a>
       </td>
     </tr>
+    <% if build.git_fork_config %>
+      <% if build.git_fork_config.branch.to_s.length > 0 %>
```
AndrewMcBurney: This is a nitpick, but could we potentially use [`positive?`](https://ruby-doc.org/core-2.3.0/Numeric.html#method-i-positive-3F) instead? If you prefer `> 0`, that's cool too.

So:

```erb
<% if build.git_fork_config.branch.to_s.length.positive? %>
```

app/features/project/views/project_info.erb
```diff
@@ -23,7 +23,7 @@
         </td>
         <td>
           <a href="<%= build.link_to_remote_commit %>" target="_blank" class="commit-sha">
-            <%= build.sha[0...7] %>
+            <%= build.human_friendly_branch_information %>
```
AndrewMcBurney: 👍

app/services/code_hosting/github_open_pr.rb
```diff
@@ -1,11 +1,29 @@
 module FastlaneCI
   # Encapsulates the data we need to know how to checkout a branch from a fork
   # We can probably use this for all the things, though.
+
+  # Example data
```
AndrewMcBurney: Beautiful documentation in this class 👍

app/shared/models/build.rb
```diff
@@ -96,5 +102,16 @@ def status=(new_value)
     def link_to_remote_commit
       project.repo_config.link_to_remote_commit(sha)
     end
+
+    # This method will return the branch name (if available)
+    # and automatically fallback to the short (8 char) git sha. Either way
+    # you'll have something nice to show to the user
+    def human_friendly_branch_information
+      if git_fork_config.branch.to_s.length > 0
```
AndrewMcBurney: Another nitpick, but could we rewrite this as:

```ruby
def human_friendly_branch_information
  return git_fork_config.branch if git_fork_config.branch.to_s.length.positive?
  return sha[0...7]
end
```

app/shared/models/build.rb
```diff
@@ -54,38 +51,47 @@ class Build
     # TODO: We currently don't use/store/support parameters (yet) https://github.com/fastlane/ci/issues/783
     attr_accessor :parameters
 
+    # @return [String] contains all information to check out that specific remote, branch, sha, ref again
+    #                  this information will be used when re-running an old build
+    #                  see https://github.com/fastlane/ci/issues/481 for more details
+    attr_accessor :git_fork_config
```
AndrewMcBurney: Is this mutated outside of where it's set in the constructor? If not, could we make this an `attr_reader` instead?

KrauseFx:
Thanks for the feedback @AndrewMcBurney 👍 Really appreciated, will address it later this weekend or Monday 👍

app/shared/models/build.rb
KrauseFx: Nice catch

app/services/code_hosting/github_open_pr.rb
KrauseFx: Yesss, the way we'd want every class ideally 👍

app/features/build/views/build.erb
KrauseFx: Thanks, yes, I prefer `> 0`, as it's easier to read/parse when skimming through source code

AndrewMcBurney: Sounds good, I'll use that convention too from now on. :)

PR #812, opened 2018-05-04
Serve web app as the fallback for all routes that haven't been defined

nakhbari:
Shout out to @joshdholtz (no relations to steve holt) for this idea

image


joshdholtz:
That @joshdholtz is such a great guy ❤️

PR #813, opened 2018-05-04
Check the expiry date when determining if the user is logged in

nakhbari:
Issues: #752

### DIFF
--- a/web/app/services/auth.service.ts
+++ b/web/app/services/auth.service.ts
@@ -17,6 +17,22 @@ export interface LoginRequest {
   password: string;
 }
 
+interface DecodedJwt {
+  exp: number;   // Token Expiry date in seconds
+  iat: number;   // Token Issue date in seconds
+  iss: string;   // Token Issuer
+  user: string;  // User ID
+}
+
+function getTokenExpiryDate(token: string): Date {
+  const base64Url = token.split('.')[1];
+  const base64 = base64Url.replace('-', '+').replace('_', '/');
+  const decodedJwt: DecodedJwt = JSON.parse(window.atob(base64));
+
+  // Need to convert from seconds to milliseconds
+  return new Date(decodedJwt.exp * 1000);
+}
+
 @Injectable()
 export class AuthService {
   constructor(private http: HttpClient) {}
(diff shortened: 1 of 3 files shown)


web/app/services/auth.service.ts
```diff
@@ -17,6 +17,22 @@ export interface LoginRequest {
   password: string;
 }
 
+interface DecodedJwt {
+  exp: number;   // Token Expiry date in seconds
+  iat: number;   // Token Issue date in seconds
```
taquitos: Is there any way we can make these property names more human readable?

nakhbari: yep, this is what we're sending from the backend. So we need to change that.

minuscorp: Those are actually the names given by the JWT specification, so they're required to be as they are named☹️

taquitos: In what way are they required to be named that? Isn't it just something we can work with? An in:
`iat: number` -> `token_issue_date_seconds: number`
and then when we actually go to transform it into a token, we make sure to map it back to `iat`?
If that's possible, I'd prefer to do that. Debugging problems with token expiration ends up being really difficult with such short variable names because you have to keep mapping it in your head if you don't do it in code.

minuscorp: Yeah of course, I meant how it worked from the backend side 😅

nakhbari: So we can change all of these names in the backend. We're not bound to any of them for JWTs to work.

https://tools.ietf.org/html/rfc7519
>4.1.  Registered Claim Names
>
>   The following Claim Names are registered in the IANA "JSON Web Token
   Claims" registry established by Section 10.1.  **None of the claims
   defined below are intended to be mandatory to use or implement in all
   cases**, but rather they provide a starting point for a set of useful,
   interoperable claims.  Applications using JWTs should define which
   specific claims they use and when they are required or optional.  All
   the names are short because a core goal of JWTs is for the
   representation to be compact.

nakhbari: @taquitos can we do the renaming in another PR?

taquitos: Yup!

minuscorp: I meant that the JWT gem we are using look for this kind of parameter naming convention in order to parse the token content and raise the corresponding exceptions, I don't know if that can be changed or parsing using plain methods instead.

PR #818, opened 2018-05-07
Update update_xcode_installation.md

orta:
minor typo.

FWIW, I think you should ask on install that's when it's totally normal to require a password and you can be very specific in the wording about what it's for


KrauseFx:
I thought more about this, and I feel like having this as a step during onboarding, that's optional, something along the lines of

> Hey, so if you want to use fastlane.ci with iOS projects, we can automatically manage Xcode for you, do you want to set this up?

And after confirmation, we'll provide instructions on how to provide the password, probably a shell command to run, or maybe optionally also providing this via a web UI

Issue #824, opened 2018-05-07
nil shows up in build list


taquitos:
```
2018-05-07 15:07:20] DEBUG Launch:  Searching for open PRs with no status and starting a build for them
Thread terminated with exception: undefined method `sha' for nil:NilClass
  /Users/jliebowitz/dev/ci/app/shared/models/build.rb:92:in `sha'
  /Users/jliebowitz/dev/ci/app/services/build_service.rb:34:in `map'
  /Users/jliebowitz/dev/ci/app/services/build_service.rb:34:in `pending_build_shas_needing_rebuilds'
  /Users/jliebowitz/dev/ci/launch.rb:216:in `block in run_pending_github_builds'
```

KrauseFx:
Sounds good, closing and moving to https://github.com/fastlane/ci/pull/825

PR #825, opened 2018-05-07
Better messaging when build stuff is missing and we crash on nil

taquitos:
Better messaging for #824


taquitos:
@KrauseFx yeah, I remember that discussion. I think it makes sense in the long run. I was worried some early adopters would get weird `nil` errors (like I did) and not understand what could be going on, so going right to the root and raising an exception that people can understand might be a good thing. That being said, it might be better to just see what happens. Closing 👍

KrauseFx:
Cool, thanks 👍

PR #833, opened 2018-05-08
Allow JSONConvertible to decode objects with required initialization parameters

### LINKED ISSUE #822: Change JSON deserialization to allow required parameters in model object constructors for models with the `JSONConvertible` mixin.
I came across this issue when tackling #817.

[This line](https://github.com/fastlane/ci/blob/master/app/shared/json_convertible.rb#L73) in `from_json!` in the `JSONConvertible` module creates a new object instance, with no parameters.

Hence, we can not have required parameters in our model constructors. Right now, we have [a bunch of optional parameters](https://github.com/fastlane/ci/blob/master/app/shared/models/user.rb#L14) with default values (usually set to `nil`).

Currently, in our deseralization method, we use `attr_accessor`s to set the values of the instance variables after object instantiation.

Some problems with this approch are:
- Some of these properties should be `attr_reader` only (such as `id`). Currently, a model's `id` could be arbitrarily set wherever since we're exposing a setter method.
- A model could be instantiated without the necessary parameters, and not fail explicitly (i.e., creating a `User` without an `email`, or without a `password_hash`).

screen shot 2018-05-07 at 1 25 37 pm

minuscorp:
Fixes #822

### DIFF
--- a/spec/shared/json_convertible_spec.rb
+++ b/spec/shared/json_convertible_spec.rb
@@ -87,6 +87,19 @@ class MockJSONConvertibleWithRequiredParams
   end
 end
 
+class MockJSONConvertibleWithMixedParams
+  include FastlaneCI::JSONConvertible
+
+  attr_reader :one_attribute
+
+  attr_accessor :other_attribute
+
+  def initialize(one_attribute:, other_attribute: nil)
+    @one_attribute = one_attribute
+    self.other_attribute = other_attribute
+  end
+end
+
 module FastlaneCI
   describe JSONConvertible do
     let (:mock_object) { MockJSONConvertible.new(one_attribute: "Hello", other_attribute: Time.at(0)) }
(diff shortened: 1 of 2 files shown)


taquitos:
Once this is merged, where do you think it would be helpful to make use of this?

AndrewMcBurney:
Hey @taquitos, I described a good use case for it in https://github.com/fastlane/ci/issues/822#issue-320906594.

spec/shared/json_convertible_spec.rb
```diff
@@ -77,6 +77,16 @@ def self.map_enumerable_type(enumerable_property_name: nil, current_json_object:
...
+
+  attr_reader :one_attribute
+
+  def initialize(one_attribute:)
+    @one_attribute = one_attribute
+  end
+end
+
 module FastlaneCI
   describe JSONConvertible do
```
taquitos: I think we need a few more specs to be safe:
1. flex the various failure cases (how many failure cases are there?)
2. objects that have a mix of required and non-required params
3. objects that have multiple required params
4. objects that have a mix of multiple required and multiple non-required params

minuscorp: Seems legit to me, will try to get this done for the weekend!

minuscorp:
Ready to merge @taquitos ?

Issue #733, opened 2018-04-23
isolate fastlane into separate process


taquitos:
- [x] FastlaneRunner to be responsible for launching a new `fastlane` process
- [x] Once fastlane process is up and running, `FastlaneRunner` to connect and start passing commands
- [x] Output from `fastlane` to be piped back to `FastlaneRunner` in real-time so we can display output
- [ ] End result should be two components, one contains sinatra and serves the end points, and the other is all the macOS-required logic, such as the `fastlane` tooling.

The idea here is that the `sinatra` server doesn't require macOS, so we should keep it that way and allow people to run it on a linux machine. When we do this, that means people can host the frontend + web server on a really cheap but fast linux machine, or in a virtual machine on any hardware. At that point, we can have macs be 100% dedicated to being `build executor`s. These `build executors` will be a separate ruby process that the web-server can talk to using `grpc` or something like that, so people can scale their build environment horizontally by buying a new mac.

Check out `fastlane.swift`: https://github.com/fastlane/fastlane/tree/master/fastlane/swift
We do very similar things here, and we should leverage the existing code.
Also, take a look at: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/commands_generator.rb#L139
This is where we start a `socket_server` and listen for connections.

Some rough documentation on fastlane.swift, might be slightly outdated:
```
How does FastlaneRunner call fastlane?
FastlaneRunner connects to a socket server and sends a command, then waits for a response.
Request payload:
{
	"commandID": "",
	"methodName": "installXcodePlugin",
	"args": [{
		"name": "url",
		"value": "http://pluginRepo.com/my_plugin",
"value_type": ""  


	}, {
		"name": "github",
		"value": ""
	}]
}

commandID is meant to help us track invocations locally, but isn’t used currently.
methodName is the action/tool name. 
args is a list of name/value pairs where the name is the parameter name, and the value is the actual value of the argument you wish to pass.
	name parameter name
	value value of the argument we are passing to the method
value_type is used in cases where the parameter is actually meant to be a closure. *This should be extended to support value type encodings like Array, and Hash*

A response payload sent from fastlane back to FastlaneRunner:
{
	"payload": {
		"status": "ready_for_next",
		"return_object": "",
		"closure_argument_value": ""
	}
}
	
status can be 
**ready_for_next**
Will include whatever return object the action returns, or empty
**failure**
Returns failure_information which is the stack trace of the exception thrown from attempting to run the command
```

[...]

Issue #820, opened 2018-05-07
Define version numbers for each dependency in our Gemfile


KrauseFx:
The goal is to define each dependency like this:
```ruby
gem "thin", ">= 1.7.2", "< 2.0.0"
```

Do not use `~` and of course don't define any version at all. The reason why we don't want to use `~>` is that it doesn't work nicely with new feature releases, usually we're fine with new feature releases (minor), but not with breaking releases (major), so by specifying the `>=` in combination with `<` we can accomplish exactly that 👍

snatchev:
In the example above, I believe the same can be accomplished with:
```ruby
gem "thin", "~> 1.7"
```

As far as I understand, leaving off the patch bit, you will get any gem version from 1.7.x up to, but not including 2.x.x.

I do agree with the philosophy of including feature versions in the lockfile. It would be neat to have a cop to enforce this :)

KrauseFx:
Yes, you're right, this works, however it comes with one problem: e.g. we tested with `1.7.2` that includes a critical bugfix, there is no way to specify with `~`, to have `1.7.2` and up, but below `2.0`

snatchev:
Ah, yes! Perfect.

KrauseFx:
This is resolved now 🥇

PR #836, opened 2018-05-08
fastlane actions run in separate process

snatchev:
Context
---
Refer to #733 for more context. This PR introduces an executable process `runner` that is intended to run on a machine that listens for fastlane actions. When a request comes in, the runner spawns a process and streams back the logs and exit status. It is also responsible for setting up the `ENV` for the command.

We are using grpc for the communication because we may eventually want put the runner on separate machines and http2 provides security, bi-directional streaming, and the ability to be proxied easily.


Gemfile
```diff
@@ -45,6 +45,9 @@ gem "jwt", "~> 2.1.0"
 # TODO: point to minimum release instead of GitHub once
 #  we shipped a new release
 
+# Interprocess communication
+gem "grpc", "~> 1.11.0"
```
KrauseFx: Make sure to follow https://github.com/fastlane/ci/issues/820 👍

minuscorp:
This is absolutely awesome!

protos/runner_services_pb.rb
```diff
@@ -0,0 +1,21 @@
...
+require "runner_pb"
+
+module FastlaneCI
+  module Runner
+    class Service
+      include GRPC::GenericService
+
+      self.marshal_class_method = :encode
+      self.unmarshal_class_method = :decode
+      self.service_name = "fastlane_c_i.Runner"
```
KrauseFx: Maybe rename to `fastlane_ci` 👍

protos/runner_pb.rb
```diff
@@ -0,0 +1,30 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
+# source: runner.proto
+
+require "google/protobuf"
+
+Google::Protobuf::DescriptorPool.generated_pool.build do
+  add_message "fastlane_c_i.Log" do
```
KrauseFx: This is auto-generated, right? can we add code comments to this file?

KrauseFx: Ideally to describe each `message`, `enum` add the attributes

protos/runner.proto
```diff
@@ -0,0 +1,34 @@
...
+
+syntax = "proto3";
+
+package fastlane_c_i;
+
+service Runner {
+  rpc Spawn(Command) returns (stream Log) {}
+}
+
+message Log {
```
KrauseFx: Same as other comment about comments 👍

protos/runner.proto
```diff
@@ -0,0 +1,34 @@
...
+  }
+  Level level = 2;
+  int32 status = 3;
+}
+
+message Command {
+  string bin = 1;
+  repeated string parameters = 2;
+  map<string,string> env = 3;
+}
```
KrauseFx: Missing new line

runner/client.rb
```diff
@@ -0,0 +1,26 @@
...
+      end
+    end
+  end
+end
+
+if $0 == __FILE__
+  client = FastlaneCI::Runner::Client.new
+  logs = client.request_spawn("ping", "-c", "20", "google.com")
+  logs.each do |log|
+    puts log.message
```
KrauseFx: Amazing 😍

protos/runner_pb.rb
```diff
@@ -0,0 +1,30 @@
...
+    optional :message, :string, 1
+    optional :level, :enum, 2, "fastlane_c_i.Log.Level"
```
KrauseFx: What's the reason those are all optional?

protos/runner_services_pb.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      rpc :Spawn, Command, stream(Log)
```
KrauseFx: What does this do, is that the `Command` we defined in the proto pufs?

runner/server.rb
```diff
@@ -0,0 +1,57 @@
...
+      # @input FastlaneCI::Runner::Command
+      # @output Enumerable::Lazy<FastlaneCI::Runner::Log> A lazy enumerable with log lines.
+      def spawn(command, _call)
+        @logger.info("spawning process with command: #{command.bin} #{command.parameters}, env: #{command.env.to_h}")
+        _stdin, stdouterr, wait_thrd = Open3.popen2e(command.env.to_h, command.bin, *command.parameters)
+        @logger.info("spawned process with pid: #{wait_thrd.pid}")
+
+        # convert every line from stdout to a Log object in a lazy stream
+        stdouterr.each_line.lazy.flat_map do |line|
+          Log.new(message: line)
```
KrauseFx: Amazing!

protos/runner_services_pb.rb
snatchev: So the package name of `fastlane_c_i` comes from the proto. I chose that particular string because it gets generated to `module FastlaneCI` in the ruby source. `fastlane_c_i` is only visible in the proto definition. I don't have a particular preference either way. I am open to changing it to anything, but I do want to avoid a namespace `FastlaneCi` existing in addition to `FastlaneCI`

protos/runner_services_pb.rb
snatchev: Yes, this is the ruby struct defined in the proto. I don't quite understand this syntax, but I read it as:
define an rpc method on this service called `Spawn` that accepts a `Command` and returns a stream of `Log`s

runner/client.rb
```diff
@@ -0,0 +1,26 @@
+require_relative "runner"
+
+module FastlaneCI
+  module Runner
+    ##
+    # A sample client that can be used to make a request to the server.
+    class Client
+      def initialize
+        @stub = Stub.new("#{HOST}:#{PORT}", :this_channel_is_insecure)
```
taquitos: 🔓

taquitos:
This is so awesome. I'm so excited, really amazing stuff here. This will really help us with process isolation and all the goodness that comes with that!!!

agent/server.rb
```diff
@@ -0,0 +1,81 @@
...
+        penum.lazy.flat_map do |line, status|
+          # proto3 doesn't have nullable fields, afaik
+          Log.new(message: line, status: (status || 0))
+        end
+      end
+    end
+  end
+end
+
+# rubocop:disable all
```
AndrewMcBurney: Why are you disabling all cops? Could you just disable the cops that are failing, please?

protos/agent_pb.rb
```diff
@@ -0,0 +1,30 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
+# source: agent.proto
+
+require 'google/protobuf'
```
AndrewMcBurney: Could you use double quotes here, please?

protos/agent_services_pb.rb
```diff
@@ -0,0 +1,30 @@
...
+# *
+# proto definitions for auto-generated client, server, and messages to pass using grpc.
+#
+# if you make any changes to this file, run
+#   `rake proto`
+# to regenerate the ruby code.
+#
+
+require 'grpc'
+require 'agent_pb'
```
AndrewMcBurney: Could you use double quotes for these two `require`s, please?

protos/agent_services_pb.rb
```diff
@@ -0,0 +1,30 @@
...
+
+module FastlaneCI
+  module Agent
+    class Service
+
+      include GRPC::GenericService
+
+      self.marshal_class_method = :encode
+      self.unmarshal_class_method = :decode
+      self.service_name = 'fastlane_c_i.Agent'
```
AndrewMcBurney: Could you use double quotes here, please?

spec/agent/server_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it "will returns lines of text from the file" do
+      expect(penum.next).to eq("this is a line of text\n")
+    end
+
+    it "enumerates lines from a file until the thread dies and returns its status" do
+      allow(thread).to receive(:alive?).and_return(false)
+      allow(thread_value).to receive(:exitstatus).and_return(0)
+      allow(io).to receive(:close)
+
+      expect { |block| penum.each(&block) }.to yield_with_args("\4", 0)
```
AndrewMcBurney: Nice! These specs helped me understand the code a lot better 👍

protos/agent_services_pb.rb
snatchev: Sorry, can't do anything about the code style in these files. They are auto-generated by `grpc`.

AndrewMcBurney: Gotcha, are the files generated excluded from rubocop in the `.rubocop.yml`?

agent/server.rb
snatchev: Good call. I think I will end up removing all of this debugging code from 68 till EOF.

snatchev:
Nope! I forgot about that. Thanks for the heads up.

Rakefile
```diff
@@ -35,6 +35,10 @@ namespace :docker do
   end
 end
 
+task :proto do
+  sh "bundle exec grpc_tools_ruby_protoc -I protos/ --ruby_out=./protos --grpc_out=./protos protos/*.proto"
```
nakhbari: Does this create ruby libs for the proto files?

protos/agent.proto
```diff
@@ -0,0 +1,45 @@
...
+
+package fastlane_c_i;
+
+service Agent {
+  rpc Spawn(Command) returns (stream Log) {}
+}
+
+message Log {
+  // the text contents of the log line.
+  string message = 1;
```
nakhbari: Could we mark what is Optional and Required?

`// Optional`
`// Required`

Rakefile
snatchev: It does!

protos/agent.proto
snatchev: Everything is optional in proto3.

nakhbari: Yes, but some fields are really required for a service to function appropriately. I'm assuming at some point we'll want to fail the gRPC if the proto in the request is not valid (ex. missing the `bin` field).

snatchev: Hmm, that's a good point. Right now, if you don't supply the value, it will default to an empty-string, and I think that will just return a 126 exit code.

Rakefile
taquitos: Let's name this `generate_protos` or something like that

agent/server.rb
```diff
@@ -0,0 +1,76 @@
...
+      # that will yield back lines from the stdout/err of the process
+      # as well as the exit status when it is complete.
+      class ProcessOutputEnumerator
+        extend Forwardable
+        include Enumerable
+
+        def_delegators :@enumerator, :each, :next
+
+        def initialize(io, thread)
+          @enumerator = Enumerator.new do |y|
```
taquitos: What's y?

agent/server.rb
```diff
@@ -0,0 +1,76 @@
...
+            y.yield(io.gets) while thread.alive?
+            io.close
+            y.yield(EOT_CHAR, thread.value.exitstatus)
+          end
+        end
+      end
+
+      def self.server
+        GRPC::RpcServer.new.tap do |s|
+          s.add_http2_port("#{HOST}:#{PORT}", :this_port_is_insecure)
```
taquitos: heh, `s` maybe some more descriptive names would make this easier to understand

docs/WorkerAgent.md
```diff
@@ -0,0 +1,44 @@
...
+## Process Environment
+Each GRPC request to the Agent should contain all of the necessary configuration information required to run the Lane successfully. This can be accomplished by supplying an environment (the ENV variables) to the child process when it is spawned by the Agent.
+
+<img src="WorkerAgent.svg">
+
+For more on using the environment for configuration (https://12factor.net/config)
+The Agent may have to fetch resources (such as source code) from other services but it should not have to fetch any configuration data.
+
+## Permission model
+The Agent will be required to run arbitrary Ruby code on the Worker. It may also be required to perform operations that require root such as installing Xcode. While there will be other security mechanisms that will enforce strict security against malicious users, we can help ourselves by preventing accidental (or incidental) damage to the Worker by running the Agents under a special user that belongs to the sudoers group. This will allow us to whitelist the commands and permissions the user can perform on a Worker.
```
taquitos: sudoers group 💯

docs/WorkerAgent.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Throttling
+Each agent is capable of spawning as many processes to handle the requests as it wants, but there are stability considerations that we should take into account. Lane execution  may be very resource heavy and launching too many of them may make the Worker kill the process prematurely. There are two possibilities for implementing throttling:
+1. Implement a serial queue on the Agent
+2. The Agent reports its capacity to the Server and rejects requests that cannot be fulfilled.
+
+Given that there will probably be a queueing service in the Server that will be responsible for distributing workload across many workers, it seems like the Agent should not implement another queue. The logic for job queueing should live somewhere higher in the stack and the Agent should simply be able to report its capacity when queried.
```
taquitos: This is great to point out. You're right, the job queueing should live higher up, and the Agent should report its capacity. We should always set this as a capacity of 1, since fastlane can't run in parallel on the same machine. That being said, we might want to define this specifically for capability.
Would it make sense to, instead of saying "capacity for 1" the idea would be "capacity for 1 fastlane job" so that later on we could report back "1 fastlane job, 1 download job, 1 whatever job" so that the Agent could potentially perform other work along with fastlane?
I'd think it could be like
```
 [
  { "fastlane" : 1},
  { "download" : 2},
  { "someOtherJobType" : 1}
]
```

docs/WorkerAgent.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Persistence
+Workers should be ephemeral and all artifacts need to be placed somewhere with strong persistence during the execution. Logs should be streamed as part of the GRPC response. After a successful fastlane run, other artifacts such as IPA files and screenshots should be persisted outside of the Worker. Additionally the raw log file should be persisted as well, to access them in the future.
```
taquitos: 💯 
Any thoughts on the best way to do this? Should we stick with GCS for now? Maybe we pass in a bucket, api key, etc along with a path?

docs/WorkerAgent.svg
```diff
@@ -0,0 +1 @@
...="matrix(0.05,0,0,0.05,43.95,0)" xlink:href="#x"/><use transform="matrix(0.05,0,0,0.05,53.95,0)" xlink:href="#A"/><use transform="matrix(0.05,0,0,0.05,62.95,0)" xlink:href="#A"/></g><g id="q"><use transform="matrix(0.05,0,0,0.05,0,0)" xlink:href="#B"/><use transform="matrix(0.05,0,0,0.05,5,0)" xlink:href="#C"/><use transform="matrix(0.05,0,0,0.05,15,0)" xlink:href="#A"/><use transform="matrix(0.05,0,0,0.05,24,0)" xlink:href="#t"/><use transform="matrix(0.05,0,0,0.05,29,0)" xlink:href="#D"/><use transform="matrix(0.05,0,0,0.05,32.95,0)" xlink:href="#C"/><use transform="matrix(0.05,0,0,0.05,42.95,0)" xlink:href="#E"/><use transform="matrix(0.05,0,0,0.05,52.95,0)" xlink:href="#x"/></g><g id="r"><use transform="matrix(0.05,0,0,0.05,0,0)" xlink:href="#D"/><use transform="matrix(0.05,0,0,0.05,3.95,0)" xlink:href="#C"/><use transform="matrix(0.05,0,0,0.05,13.949999999999998,0)" xlink:href="#E"/><use transform="matrix(0.05,0,0,0.05,23.95,0)" xlink:href="#x"/></g><g id="s"><use transform="matrix(0.05,0,0,0.05,0,0)" xlink:href="#u"/><use transform="matrix(0.05,0,0,0.05,5.95,0)" xlink:href="#z"/><use transform="matrix(0.05,0,0,0.05,15.949999999999998,0)" xlink:href="#E"/></g></defs></g></svg>
```
taquitos: Nice graphic

docs/WorkerAgent.md
snatchev: I was thinking the same thing, that some jobs maybe be parallelizable and there are a few ways to implement this with different tradeoffs. I didn't feel comfortable committing to any one of them in code at this point, but I did want to document it.

docs/WorkerAgent.md
minuscorp: Depending on which `ArtifactProvider` is the user using. So my thoughts here is to provide the ability for those `ArtifactProviders` is to store the artifacts in the parent process by adding a functionality to send the artifact to the parent and him storing it, or provide the ability to each worker with all the information needed to store the artifacts by themselves.

snatchev: I think there is a high-level open question here. In my mind, there are 2 options.

1.) A lane that creates an artifact puts it somewhere in a temporary place that the Agent knows about. It is the job of the Agent to deliver the artifact in a permanent place. This would be something configured in the webapp.
2.) The lane defines where the artifacts go, possibly using some new fastlane actions. This would be configured in the Fastfile.

I think we should revisit this in the near feature when we start scoping out artifacts as a whole.

taquitos: Cool, I actually don't think fastlane should care at all (or lanes). It's totally a CI thing, so it's more of a CI general setting that a project can override. 

I think it might make sense for each build runner to do the storage

docs/WorkerAgent.md
taquitos: 🆒

docs/WorkerAgent.md
minuscorp: So for the `LocalArtifactProvider` and others, we have to add a functionality that allows the Agent to send the arftifacts to the main server. IMO, I think the main node should gather all the artifacts from the Agents, and store them (as the main node knows which `ArtifactProvider` and its configuration is)

PR #837, opened 2018-05-09
Use Faraday::Request::Retry to prevent crashes due to poor Internet connectivity

### LINKED ISSUE #798: Intermittent connection error causes server to become unstable
`[2018-05-02 17:10:06] DEBUG GitHubService:  Failed to open TCP connection to api.github.com:443 (getaddrinfo: nodename nor servname provided, or not known)`

`[2018-05-02 17:12:37] ERROR CheckForNewCommitsOnGithubWorker:  [FastlaneCI::CheckForNewCommitsOnGithubWorker Exception], work unit caused exception: Failed to open TCP connection to api.github.com:443 (getaddrinfo: nodename nor servname provided, or not known): 
`

minuscorp:
Fixes #798

### DIFF
--- a/app/shared/github_handler.rb
+++ b/app/shared/github_handler.rb
@@ -112,7 +112,7 @@ module FastlaneCI
           Faraday::Adapter::NetHttp,
           Faraday::Request::Retry,
           max: 1000,
-          interval: 0.05,
+          interval: 0.5,
           interval_randomness: 0.5,
           backoff_factor: 2,
           exceptions: [
--- a/app/shared/github_handler.rb
+++ b/app/shared/github_handler.rb
@@ -1,5 +1,7 @@
 require "octokit"
 require "faraday"
+require "socket"
+require "net/http"
 require_relative "logging_module"
 
 # TODO: eventually we'll want to consider handling all these:
@@ -42,6 +44,24 @@ module FastlaneCI
 
     def github_action(client, &block)
       if client.kind_of?(Octokit::Client)
+        # Inject Faraday::Request::Retry to the client if necessary
+        unless client.middleware.handlers.include?(Faraday::Request::Retry)
+          middleware_dup = client.middleware.dup
+          client.middleware = middleware_dup
+          client.middleware.insert_before(Faraday::Adapter::NetHttp,
+                                          Faraday::Request::Retry,
+                                          max: 1000,
+                                          interval: 0.05,
+                                          interval_randomness: 0.5,
+                                          backoff_factor: 2,
+                                          exceptions: [
+                                            Errno::ETIMEDOUT,
+                                            "Timeout::Error",
+                                            Faraday::Error::TimeoutError,
+                                            Faraday::Error::RetriableResponse,
+                                            SocketError
+                                          ])
+        end
         # `rate_limit_retry_count` retains the variables through iterations so we assign to 0 the first time.
         rate_limit_retry_count ||= 0
         begin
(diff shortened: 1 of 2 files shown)


app/shared/models/git_repo.rb
```diff
@@ -87,6 +88,21 @@ def pushes_disabled?
       def load_octokit_cache_stack
         @stack ||= Faraday::RackBuilder.new do |builder|
           builder.use(Faraday::HttpCache, serializer: Marshal, shared_cache: false)
+          # In order to have access to SocketError and ETIMEDOUT exceptions.
+          require "socket"
+          require "net/http"
+          builder.use(Faraday::Request::Retry,
```
taquitos: Instead of applying that logic here, I think we should improve the retry logic here: https://github.com/fastlane/ci/blob/master/app/shared/github_handler.rb#L36

That way we don't have multiple layers doing their own retry/backoff logic. 
It's better that we have custom control over how this works in 1 single place.

app/shared/github_handler.rb
```diff
@@ -42,6 +44,24 @@ def self.included(klass)
 
     def github_action(client, &block)
       if client.kind_of?(Octokit::Client)
+        # Inject Faraday::Request::Retry to the client if necessary
+        unless client.middleware.handlers.include?(Faraday::Request::Retry)
```
taquitos: I think it could be worthwhile to pull this out into a separate method instead of inline in the `github_action` method.

app/shared/github_handler.rb
```diff
@@ -42,6 +44,24 @@ def self.included(klass)
...
+          middleware_dup = client.middleware.dup
+          client.middleware = middleware_dup
+          client.middleware.insert_before(Faraday::Adapter::NetHttp,
+                                          Faraday::Request::Retry,
+                                          max: 1000,
+                                          interval: 0.05,
```
taquitos: I wonder if a larger interval value would be better.

app/shared/github_handler.rb
```diff
@@ -42,6 +44,24 @@ def self.included(klass)
 
     def github_action(client, &block)
       if client.kind_of?(Octokit::Client)
+        # Inject Faraday::Request::Retry to the client if necessary
+        unless client.middleware.handlers.include?(Faraday::Request::Retry)
+          middleware_dup = client.middleware.dup
+          client.middleware = middleware_dup
+          client.middleware.insert_before(Faraday::Adapter::NetHttp,
```
AndrewMcBurney: This is a style nitpick, but could you please put the first parameter on a new line, and indent the parameters 2 spaces out?

i.e.,

```ruby
client.middleware.insert_before(
  Faraday::Adapter::NetHttp,
  Faraday::Request::Retry,
  # ...
)
```

app/shared/github_handler.rb
```diff
@@ -42,6 +44,8 @@ def self.included(klass)
 
     def github_action(client, &block)
       if client.kind_of?(Octokit::Client)
+        # Inject Faraday::Request::Retry to the client if necessary
+        client = inject_retry_middleware(client)
```
taquitos: ❤️

app/shared/github_handler.rb
```diff
@@ -97,5 +101,30 @@ def github_action(client, &block)
...
+    private
+
+    def inject_retry_middleware(client)
+      unless client.middleware.handlers.include?(Faraday::Request::Retry)
+        middleware_dup = client.middleware.dup
+        client.middleware = middleware_dup
+        client.middleware.insert_before(
+          Faraday::Adapter::NetHttp,
+          Faraday::Request::Retry,
+          max: 1000,
```
taquitos: I think max should be lower here. We probably shouldn't go nuts since 1000 tries + exponential backoff means we'll never stop retrying. I think 3 tries is enough, then we'll need to catch the resulting exception.

minuscorp: Fair enough! Will change!

PR #838, opened 2018-05-09
Add doc to start discussion on how to handle errors

taquitos:


docs/HandlingErrors.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Background
+
+Many different errors can happen at various different points. Having a system to effectively present errors and the details needed to know what to do, and when, is critically important
+
+## Error situations
+
+1. Something broke, and it might be transient, you can try that action again
+1. Some 3rd party dependency failed, and you’ll need to figure out what to do (e.g. GitHub is down, or Xcode couldn't be installed)
+1. You hit a fastlane.ci or fastlane bug
```
nakhbari: Maybe we should break up fastlane bugs and fastlane ci bugs, since they have different debugging paths.

docs/HandlingErrors.md
```diff
@@ -0,0 +1,33 @@
...
+
+Some errors are recoverable by just retrying (we do that automatically for things that make sense). Some are recoverable only when somebody logs into the box and fixes the problem. Some problems are not recoverable at all.
+
+## Presenting Errors
+
+Current thinking about errors, they:
+
+1. Show up and stay around until dismissed on a given page, e.g. fastlane broke on this build, but if you want, you can retry it, but you only want this to persist on the project build page
```
nakhbari: Is this a notification for an error wherever you are in the UI? Or just how it's displayed in the build details?

docs/HandlingErrors.md
taquitos: Yeah, potentially, we could separate it by `fastlane.ci` and `build runner`

docs/HandlingErrors.md
nakhbari: So the assumption is that if there is a CI error on one build it is a problem for all builds? If that's true this sounds good to me.

taquitos: Yup, while it's not guaranteed, it's most likely that a runner problem is a problem for all.

taquitos:
@snatchev I'd say, storing is already more or less covered by the notification system that @AndrewMcBurney built (https://github.com/fastlane/ci/issues/84 and https://github.com/fastlane/ci/pull/142). I started using it for messages and it's nice, you can see an example in https://github.com/fastlane/ci/blob/master/app/shared/models/git_repo.rb#L179. I think this is more for defining what errors/potential errors we want to display and how.

docs/HandlingErrors.md
nakhbari: Can we add this assumption to the doc?

nakhbari (approved):

Issue #830, opened 2018-05-08
Can not run in docker


chaim1986:
mac-pro:ci chaim$ docker  version 
Client:
 Version:      18.03.1-ce
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   9ee9f40
 Built:        Thu Apr 26 07:13:02 2018
 OS/Arch:      darwin/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.03.1-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.5
  Git commit:   9ee9f40
  Built:        Thu Apr 26 07:22:38 2018
  OS/Arch:      linux/amd64
  Experimental: true
mac-pro:ci chaim$ 

------error log . --- 

mac-pro:ci chaim$ docker-compose up --build
Building web
Step 1/9 : FROM ruby:2.5
 ---> fb664b54b956
Step 2/9 : ENV LANG C.UTF-8
 ---> Using cache
 ---> 2c19102643b6
Step 3/9 : RUN bundle config --global frozen 1
 ---> Using cache
 ---> 039cb016eb93
Step 4/9 : WORKDIR /usr/src/app
 ---> Using cache
 ---> 7503a14ad782
Step 5/9 : COPY Gemfile Gemfile.lock ./
 ---> Using cache
 ---> 12bbaddd71a9
Step 6/9 : RUN bundle install
 ---> Using cache
 ---> d01316a66c22
Step 7/9 : COPY . .
 ---> Using cache
 ---> c8aced358a72
Step 8/9 : CMD ["prod"]
 ---> Using cache
 ---> 4e8724ebdd5f
Step 9/9 : ENTRYPOINT ["bundle", "exec", "rake"]
 ---> Using cache
 ---> ba1b3775dce6
Successfully built ba1b3775dce6
Successfully tagged ci_web:latest
Starting ci_web_1 ... done
Attaching to ci_web_1
web_1  | bundle exec rackup -p 8080 --env production
web_1  | warning: parser/current is loading parser/ruby25, which recognizes
web_1  | warning: 2.5.0-compliant syntax, but you are running 2.5.1.
web_1  | warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
web_1  | RuntimeError: The web app not built. Build with the Angular CLI and Try Again, e.g. ng build --deploy-url="/.dist"
[...]

PR #839, opened 2018-05-10
Dockerfile updates

taquitos:
- Switch to testing prod mode for now.
- include node
- update ruby version to 2.3 instead of 2.5
- expose port 8080 by default

Fixes: https://github.com/fastlane/ci/issues/830
Although, note that it's running the rake task: `hosted_prod_test` not `prod`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,5 +5,6 @@ fastlane.log
 fastlane.verbose.log
 public/.dist/*
 **/node_modules/*
+node_modules
 package-lock.json
 npm-debug.log
--- a/Rakefile
+++ b/Rakefile
@@ -6,11 +6,15 @@ task :prod do
   sh "bundle exec rackup -p 8080 --env production"
 end
 
+task :hosted_prod_test do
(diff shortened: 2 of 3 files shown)


.dockerignore
```diff
@@ -5,5 +5,6 @@ fastlane.log
 fastlane.verbose.log
 public/.dist/*
 **/node_modules/*
+node_modules
```
KrauseFx: My favorite folder isn't included 😢

taquitos: don't worry, you can run `npm install` and that will still download 85gb of `pad-left` dependencies.

KrauseFx: Beautiful

Rakefile
```diff
@@ -6,11 +6,15 @@ task :prod do
   sh "bundle exec rackup -p 8080 --env production"
 end
 
+task :hosted_prod_test do
+  sh("FASTLANE_CI_ERB_CLIENT=1 DEBUG=1 FASTLANE_CI_DISABLE_REMOTE_STATUS_UPDATE=1 FASTLANE_CI_DISABLE_PUSHES=1 FASTLANE_CI_SKIP_RESTARTING_PENDING_WORK=1 bundle exec rackup --host 0.0.0.0 -p 8080 --env development")
```
KrauseFx: This is interesting, this fails `bundle exec rubocop` for me locally:

```
Rakefile:10:121: C: Metrics/LineLength: Line is too long. [215/120]
```

Anybody got an idea on why that is? Rubocop version is locked, and it should have failed this PR's build?

PR #840, opened 2018-05-11
Added specs for `JSONProjectDataSource`.

AndrewMcBurney:
### Changes
- Added specs for the `JSONProjectDataSource` configuration data source.
- `update_project!` and `delete_project!` in the `JSONProjectDataSource` now return `true` if they complete successfuly, like the `JSONUserDataSource`.

### DIFF
--- a/app/services/config_data_sources/json_project_data_source.rb
+++ b/app/services/config_data_sources/json_project_data_source.rb
@@ -226,6 +226,8 @@ module FastlaneCI
         projects = self.projects
         projects[project_index] = project
         self.projects = projects
+
+        return true
       end
     end
 
@@ -233,6 +235,7 @@ module FastlaneCI
       unless project.nil?
         raise "project must be configured with an instance of #{Project.name}" unless project.class <= Project
       end
+
       project_index = nil
       existing_project = nil
       projects.each.with_index do |old_project, index|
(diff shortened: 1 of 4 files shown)


app/services/config_data_sources/json_project_data_source.rb
```diff
@@ -226,13 +226,16 @@ def update_project!(project: nil)
         projects = self.projects
         projects[project_index] = project
         self.projects = projects
+
+        return true
```
taquitos: This is weird to me. I get that it can help with testing, but I think the better way to test that this method does as it's supposed to is by ensuring the `project` you updated is actually up-to-date by testing the resulting `update` is better than adding a return statement here. Since we're dealing with state here, I think explicitly checking that our state updated exactly how we expected is 🔑

app/services/config_data_sources/json_project_data_source.rb
```diff
@@ -253,6 +256,8 @@ def delete_project!(project: nil)
         projects = self.projects
         projects.delete_at(project_index)
         self.projects = projects
+
+        return true
```
taquitos: same comment here about return statements

app/services/config_data_sources/json_project_data_source.rb
taquitos: Yeah, I think removing this stuff is the best course of action. Returning `true` isn't really helpful. It is really just a signal that this method exited without raising an exception, and it doesn't provide useful info since you're not going to check for it during runtime.

If we do this in other places, we should probably fix that at some point, would you mind opening an issue that points this out?

PR #842, opened 2018-05-11
Remove `return true` in CRUD actions in data sources

### LINKED ISSUE #841: Remove `return true` from `JSONUserDataSource` `update` and `delete` methods.
As per [this comment](https://github.com/fastlane/ci/pull/840#discussion_r187680366), we want to remove instances where we return `true` in our CRUD actions for data source services.

AndrewMcBurney:
### Changes
- Closes #841

### DIFF
--- a/app/features/environment_variables/environment_variables_controller.rb
+++ b/app/features/environment_variables/environment_variables_controller.rb
@@ -42,10 +42,9 @@ module FastlaneCI
           value: params[:value]
         )
 
-        success = Services.environment_variable_service.update_environment_variable!(
+        Services.environment_variable_service.update_environment_variable!(
           environment_variable: environment_variable
         )
-        logger.error("Something went wrong") unless success
       end
 
       redirect(HOME)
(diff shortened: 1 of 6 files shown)


app/features/environment_variables/environment_variables_controller.rb
```diff
@@ -42,21 +42,19 @@ class EnvironmentVariablesController < AuthenticatedControllerBase
           value: params[:value]
         )
 
-        success = Services.environment_variable_service.update_environment_variable!(
+        Services.environment_variable_service.update_environment_variable!(
           environment_variable: environment_variable
         )
-        logger.error("Something went wrong") unless success
```
AndrewMcBurney: We log the error if this doesn't succeed [here](https://github.com/fastlane/ci/compare/master...AndrewMcBurney:remove_return_true_in_crud_actions?expand=1#diff-040760123e1c4871883a4e55898a005bR58).

PR #848, opened 2018-05-13
Support any hostname that points to localhost

### LINKED ISSUE #210: Support for 127.0.0.1 additionally to localhost
Since https://github.com/fastlane/ci/pull/200 switched to the `thin` web server, we now need to access `localhost` instead of `127.0.0.1`, ideally we support both. Probably just a setting in the web server somewhere

thii:
Fixes #210.

Enable this in order to run fastlane.ci inside a VM and allow accessing
to it from the host machine.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,5 @@
 task :dev do
-  sh "bundle exec rackup -p 8080 --env development"
+  sh "bundle exec rackup -o 0.0.0.0 -p 8080 --env development"
 end
 
 task :prod do


taquitos:
Ugh, it seems like this is required for VMs, but I had problems mapping a docker container port without `0.0.0.0`. Maybe we need a docker specific rake task and a VM specific rake task.

thii:
Same for VMs, you need to expose `0.0.0.0` in order map a VM's port to the host machine.

PR #850, opened 2018-05-13
Support deployment using Vagrant

thii:
Since Docker does not support macOS guests, this allows deploying an isolated instance that can be used to build iOS projects.

### DIFF
--- a/vagrant-provision.sh
+++ b/vagrant-provision.sh
@@ -1,0 +1,19 @@
+if [ ! -d /usr/local/Homebrew/.git ]; then
+  echo "==> Installing Homebrew ..."
+  /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+  brew doctor
+else
+  echo "==> Homebrew is already installed"
+fi
+
+echo "==> Installing node and npm ..."
+brew install node
+
+echo "==> Installing Bundler ..."
+sudo gem install bundler -NV
+
+echo "==> Installing all dependencies ..."
+cd /fastlane-ci
+bundle install
+npm install
+npm run build
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ fastlane/report.xml
 
 # unless supporting rvm < 1.11.0 or doing something fancy, ignore this:
 .rvmrc
+
+# Vagrant VM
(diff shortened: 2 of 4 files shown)


vagrant-provision.sh
```diff
@@ -0,0 +1,19 @@
...
+  brew doctor
+else
+  echo "==> Homebrew is already installed"
+fi
+
+echo "==> Installing node and npm ..."
+brew install node
+
+echo "==> Installing Bundler ..."
+sudo gem install bundler -NV
```
snatchev: Is this sudo required?

tkrajacic: That depends on if you are using `rbenv`?

thii: Yes, it is required on a fresh install of masOS where no Ruby version manager is installed.

snatchev: Okay as long as the vagrant user has password-less sudo, and I think the first time you sudo you get a prompt saying you agree to perform the action.

Alternatively, you can `gem install --user-install bundler` and make sure `~/.gem/ruby/2.3.0/bin` is in the PATH

snatchev: Also, please use the long-form switches (e.g. `--no-document`) in scripts. It's much easier to read :)

thii: Done. PTAL.

snatchev:
Thanks @thii! I am very much in favor of being able to use fastlane.ci via vagrant. I think this PR is great. I haven't been able to test it, because... I ran out of disk space downloading macOS! 😅

snatchev:
@thii I was not able to complete the provision phase successfully.

During brew install I got:
```
    default: Software Update Tool
    default: ==> /usr/bin/sudo /bin/rm -f /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress
    default: ==> /usr/bin/sudo /usr/bin/xcode-select --switch /Library/Developer/CommandLineTools
    default: xcode-select: error: invalid developer directory '/Library/Developer/CommandLineTools'
    default: Failed during: /usr/bin/sudo /usr/bin/xcode-select --switch /Library/Developer/CommandLineTools
    default: /tmp/vagrant-shell: line 7: brew: command not found
```

Any ideas?

thii:
Can we :ship: this?

PR #852, opened 2018-05-14
Don't strip value of ENV variable

KrauseFx:
This would cause errors if e.g. a password is passed via an ENV variable and it starts or end with a space


KrauseFx:
I don't think there was a reason

PR #857, opened 2018-05-14
Add Docker-specific rake tasks

taquitos:
- Prevent conflicts with non-Docker setups
- See https://github.com/fastlane/ci/pull/848 for an example

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,13 +1,9 @@
 task :dev do
-  sh "bundle exec rackup -o 0.0.0.0 -p 8080 --env development"
+  sh "bundle exec rackup -p 8080 --env development"
 end
 
 task :prod do
   sh "bundle exec rackup -p 8080 --env production"
-end
-
-task :hosted_prod_test do
-  sh("FASTLANE_CI_ERB_CLIENT=1 DEBUG=1 FASTLANE_CI_DISABLE_REMOTE_STATUS_UPDATE=1 FASTLANE_CI_DISABLE_PUSHES=1 FASTLANE_CI_SKIP_RESTARTING_PENDING_WORK=1 bundle exec rackup --host 0.0.0.0 -p 8080 --env development")
 end
 
 task :devbootstrap do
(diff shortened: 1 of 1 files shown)


taquitos:
Oh, I didn't realize that was a thing. I'm new to Rake. I'll do some reading and see if it makes sense to use.

KrauseFx:
While we talk about this, how much of the code do we want to have in a Rakefile, or should those be separate Ruby/shell files in a different structure? 

From what I see, Rake has 2 powerful main features we'd make use of

1) `task :name` to easily define tasks in combination with namespaces
2) Lots of helpers, especially around file system related tasks. I don't think we make use of them yet for `fastlane.ci`

Either way is fine, I know Rakefile can be structured really well, as you can extend it and provide your own helper methods, to keep the main Rakefile clean 👍

Rakefile
```diff
@@ -1,22 +1,41 @@
...
+  task :prod_test do
+    # rubocop:disable Metrics/LineLength
+    sh("FASTLANE_CI_ERB_CLIENT=1 DEBUG=1 FASTLANE_CI_DISABLE_REMOTE_STATUS_UPDATE=1 FASTLANE_CI_DISABLE_PUSHES=1 FASTLANE_CI_SKIP_RESTARTING_PENDING_WORK=1 bundle exec rackup --host 0.0.0.0 -p 8080 --env development")
+    # rubocop:enable Metrics/LineLength
+  end
+
+  desc "Bootstrap for running in a Docker container"
+  task :dev_bootstrap do
+    sh("bundle install")
+    sh("brew install node") unless sh("npm -v")
```
KrauseFx: This is great

Rakefile
```diff
@@ -1,22 +1,41 @@
...
   sh("npm install")
   sh("ln -sf ../../.pre-commit .git/hooks/pre-commit")
 end
 
+# It seems like mapping ports in Docker requires --host which can break hosting in VMs
+# So we'll just keep some docker-specific tasks down here
+namespace :docker do
+  desc "Run in a Docker container in production mode"
+  task :prod do
+    sh "bundle exec rackup --host 0.0.0.0 -p 8080 --env production"
```
KrauseFx: Maybe I'm just confused here, but didn't you want the 0.0.0.0 to not be in `docker`?

Rakefile
thii: I guess we can't use `brew` in Linux.

KrauseFx: nice catch, we could move it out into its own method then to make those checks and use different installation methods

Rakefile
taquitos: Docker needs 0.0.0.0 for the host, while VMs shouldn't have it. So, if we're running on local or in a VM, we can just use `prod`

taquitos:
I agree, we should make sure the rakefile is clean, and organized. I also think we should limit the amount of stuff that goes in there.

Rakefile
taquitos: You aren't confused, I was 🙃

Rakefile
```diff
@@ -3,20 +3,38 @@ task :dev do
 end
 
 task :prod do
-  sh "bundle exec rackup -p 8080 --env production"
+  sh "bundle exec rackup -o 0.0.0.0 -p 8080 --env production"
```
thii: Oh, and since we're using long-form flags in other places, let's use `--host` and `--port` instead of `-o` and `-p` 👍

taquitos: Great idea 💯

PR #862, opened 2018-05-14
Add endpoint which exposes a project's lanes

nakhbari:
Issue: #713 

Example Response Data: `[{"name":"test","platform":"ios"}]`


nakhbari:
Realized this needs to be independent of the project. Will work on it some more.

nakhbari:
Good to be reviewed again!

app/features-json/repos_json_controller.rb
```diff
@@ -16,5 +17,57 @@ class RepositoryJSONController < AuthenticatedControllerBase
...
+      )
+
+      github_service = FastlaneCI::GitHubService.new(provider_credential: provider_credential)
+
+      selected_repo = github_service.repos.detect do |repo|
+        repo_full_name == repo[:full_name]
+      end
+
+      if selected_repo.nil?
+        raise "Could not find repo, check that your github token has access to repo: #{repo_full_name}}"
```
KrauseFx: In this case, we're using the user's GitHub token, not the fastlane bot account token, right? If so, maybe we can change the messaging to mention which GH token was tried to be used?

app/features-json/repos_json_controller.rb
```diff
@@ -16,5 +17,57 @@ class RepositoryJSONController < AuthenticatedControllerBase
...
+      )
+
+      return fetch_available_lanes(fastfile_parser).to_json
+    end
+
+    def fetch_available_lanes(fastfile_parser)
+      # we don't want to show `_before_all_block_`, `_after_all_block_` and `_error_block_`
+      # or a private lane as an available lane
+      lanes = []
+      fastfile_parser.tree.each do |platform, value|
```
KrauseFx: We already have that code in `project_controller.rb` - the plan is to remove it from project controller as it's only for the `erb`, right?

app/features-json/view_models/lane_view_model.rb
```diff
@@ -0,0 +1,19 @@
+require_relative "../../shared/json_convertible"
+
+module FastlaneCI
+  # View model to expose the a project's lane.
+  class LaneViewModel
```
KrauseFx: I love those view models 😍

app/features-json/repos_json_controller.rb
nakhbari: yes. I'm imagining the ERB controllers as a scratch pad and the final code goes here.

PR #868, opened 2018-05-15
Create PULL_REQUEST_TEMPLATE.md

KrauseFx:
A first version, we'll add more things as we go. As right now `master` doesn't launch up the server due to a regression, this makes it really hard to work in the fastlane/ci code base.


KrauseFx:
That's a great idea 👍 Will do

PR #879, opened 2018-05-16
redo pending builds on launch

taquitos:
Passing in a list of branches is actually not what we wanted to do for PR checking. That only should be used when checking for commits.


snatchev:
Lets get rid of the bit that fetches `branches_to_check` as well.

PR #882, opened 2018-05-16
Close that dialog! (Add Project dialog)

nakhbari:
Issue: #713 


web/app/dashboard/add-project-dialog/add-project-dialog.component.html
```diff
@@ -46,9 +46,9 @@
   </div>
 </div>
 <div mat-dialog-actions>
-  <button mat-button>Close</button>
+  <button mat-button (click)="closeDialog()">Close</button>
```
taquitos: TIL: `(click)` is a thing

PR #885, opened 2018-05-16
Revert "Use JWTAuth Middleware to check correct authentication with t…

nakhbari:
Revert "Use JWTAuth Middleware to check correct authentication with the backend. (#872)"
This reverts commit 1a25394c1d65c8828559c75570ccf368994f7926.

The middleware auth is being applied to all routes, blocking most requests.


minuscorp (approved):
I guess it's ready to merge, although I don't know why is this blocking, it should block every viewmodel request except from logging in (as expected)

minuscorp (changes requested):
Will block this until #888 is reviewed and merged. Better that than reverting the whole auth.

taquitos:
@minuscorp / @nakhbari now that the referenced PR is merged (#888) do we still need to revert or does that fix the problem?

nakhbari:
@snatchev is working on robustification, let's close this for now

PR #889, opened 2018-05-17
Fix user enforcement [WIP]

nakhbari:
- [ ] I have tested this change locally and tried to launch the server as well as access a project, and with that at least one build
- [ ] If there is an existing issue, make sure to add `Fixes ...` as part of the PR body to refernece the issue you're solving

### DIFF
--- a/app/features-json/middleware/jwt_auth.rb
+++ b/app/features-json/middleware/jwt_auth.rb
@@ -5,7 +5,7 @@ require "jwt"
 module FastlaneCI
   # API Middleware responsible of authenticate all the requests that uses it.
   class JwtAuth
-    def initialize(app, encryption_key=nil, protect_endpoints_starting_with="/data")
+    def initialize(app, encryption_key = nil, protect_endpoints_starting_with = "/data")
       @app = app
       @encryption_key = encryption_key || FastlaneCI.dot_keys.encryption_key
       @protect_endpoints_starting_with = protect_endpoints_starting_with
@@ -18,7 +18,7 @@ module FastlaneCI
         begin
           options = { verify_iss: true, verify_iat: true, algorithm: "HS256", iss: "fastlane.ci" }
           bearer = env.fetch("HTTP_AUTHORIZATION", "").slice(7..-1)
-          payload = JWT.decode(bearer, @encryption_key, true, options)
+          payload = JWT.decode(bearer, @encryption_key, true, options)[0]
 
           raise "Missing user ID in payload." unless payload["user"]
           env[:user] = payload["user"]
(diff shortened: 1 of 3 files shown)


app/features-json/middleware/jwt_auth.rb
```diff
@@ -5,7 +5,7 @@
 module FastlaneCI
   # API Middleware responsible of authenticate all the requests that uses it.
   class JwtAuth
-    def initialize(app, encryption_key=nil, protect_endpoints_starting_with="/data")
+    def initialize(app, encryption_key = nil, protect_endpoints_starting_with = "/data")
```
taquitos: instead of `=` we generally default to `:` so they can be used as named params

app/features-json/middleware/jwt_auth.rb
```diff
@@ -18,7 +18,7 @@ def call(env)
         begin
           options = { verify_iss: true, verify_iat: true, algorithm: "HS256", iss: "fastlane.ci" }
           bearer = env.fetch("HTTP_AUTHORIZATION", "").slice(7..-1)
-          payload = JWT.decode(bearer, @encryption_key, true, options)
+          payload = JWT.decode(bearer, @encryption_key, true, options)[0]
```
taquitos: let's use named params

taquitos: What is `[0]`? Let's add a comment here

spec/features-json/jwt_auth_spec.rb
```diff
@@ -7,8 +7,20 @@
   let(:inner_app) { ->(env) { [200, env, "app"] } }
   let(:app) { described_class.new(inner_app, "fastlane-ci-test", "/buy") }
 
-  context "Client makes a request without authentication headers" do
+  context "Client makes a request with correct auth header" do
```
taquitos: ❤️  testsssssss

nakhbari:
@snatchev is working on a better version. Gonna close this one.

app/features-json/middleware/jwt_auth.rb
minuscorp: why is this being changed? 🤔  better take `payload,` if there multiple return statements from `.decode`

app/features-json/middleware/jwt_auth.rb
minuscorp: This was intentionally done, so the middleware had minimum configuration and preferred convention, but can be changed if you think so

taquitos: In general we prefer to have named parameters. If there's a reason to keep it this way, that's totally cool. I just don't understand the advantage in not using named parameters here.

minuscorp: There's no other reason that try to follow how other Rack middleware's parameter naming worked, so I used unnamed params. But ofc we can change it to named one 😄

PR #891, opened 2018-05-18
new authenticated base class for API controllers.

snatchev:

Context
---
fastlane.ci is developing an Angular SPA for it's UI. The app will talk to the backend via JSON APIs and will use JWT for it's authentication.

Why a new API Controller base class vs `ControllerBase` ?

API requests are very different from regular browser based requests. The `ControllerBase` assumes we are serving to regular browser traffic. For APIs we want:
* no flash
* no templates and views
* no session
* JSON requests and responses by default

For these reasons, it makes sense to just create a new base class from `Sinatra::Base` than to try to make `ControllerBase` more generic.

Why not check authentication in a Rack middleware?

Using a Rack middleware makes selective authentication on endpoints tricky. We want security by default for a given controller, but we also want to allow controllers that have some endpoints secure and some public. It feels brittle to couple the authentication to specific routes or route patterns. We also want to make sure it's easy for developers to do the right thing. For example, if a developer changes a route for an endpoint, they shouldn't have to remember to update an authentication route list in another file.

Check the specs for some example usage.

Remaining tasks
- [x] add specs for user authentication methods `current_user` and `user_id`

### DIFF
--- a/app/features-json/api_controller.rb
+++ b/app/features-json/api_controller.rb
@@ -1,0 +1,125 @@
+require "sinatra/base"
+require "sinatra/reloader"
+require "jwt"
+
+require_relative "../services/services"
+require_relative "json_params"
+
+module FastlaneCI
+  ##
+  # APIController is designed for state-less API requests servicing the Angular front-end
+  #
+  # Requests are expected to be application/json.
+  # The `params` method can be used to access the JSON body as if it were a hash
+  #
+  # Authentication is enabled by default. It can be disabled for the entire controller,
+  # and individual endpoints can be give an authentication conditions.
+  #
+  # Usage
+  # ===
+  # Subclass the APIController for any controller that you expect JSON requests and responses.
+  #
+  # Settings
+  # ===
+  #
+  # The following settings are provided to subclasses of APIController:
+  #
+  # * `authentiation`- boolean value that makes every request check authentication in a before filter.
+  #    Disable with: `disable :authentication`. Enabled by default.
+  # * `authenticate_via` - which authentication scheme to use. `:jwt` by default.
+  # * `jwt_secret` - The key to use in decoding JWT tokens.
+  #
+  # Per-route Authentication
+  # ===
+  # If you disable authentication for the whole controller, you can enable it on a per-route basis using
+  # a route condition like this:
+  #
+  #   get '/private', authenticate: :jwt do
+  #     json({message: 'secret'})
+  #   end
+  #
+  # User authentication
+  # ===
+  # This controller also provides a few helper methods that return information about the current user
+  # * `user_id` - the user id encoded in the jwt.
+  # * `current_user` - the user model fetched from the data service, or nil if it cannot be found.
+  # * `user_logged_in?` - boolean whether the user is found.
+  #
+  class APIController < Sinatra::Base
+    include JSONParams
+    include Logging
+
+    configure(:development) do
+      register Sinatra::Reloader
+    end
+
+    # to disabled authentication for the entire controller use:
+    # disable(:authentication)
+    set(:authentication, true)
+
+    # the default authentication scheme. We only support `:jwt` at this time.
+    set(:authenticate_via, :jwt)
+
+    # the condition can be added to any route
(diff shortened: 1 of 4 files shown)


app/features-json/api_controller.rb
```diff
@@ -0,0 +1,124 @@
...
+    set(:authenticate) do |auth_type|
+      condition { authenticate!(via: auth_type) }
+    end
+
+    # the JWT secret uses the fastlane encryption key
+    set(:jwt_secret, FastlaneCI.dot_keys.encryption_key)
+
+    before do
+      next unless settings.authentication?
+      authenticate!(via: settings.authenticate_via)
```
nakhbari: in case we add more logic here would it make sense to do authenticate!(..) if settings.authentication

app/features-json/api_controller.rb
```diff
@@ -0,0 +1,124 @@
...
+    end
+
+    helpers do
+      # Decode the JWT or halt.
+      def jwt
+        authorization = request.env["HTTP_AUTHORIZATION"]
```
nakhbari: is request made available through Sinatra:Reloader?

nakhbari:
Oh my lord this is beautiful 🥇

app/features-json/api_controller.rb
snatchev: Nah, Reloader is just a development hot-reloader so you dont need to restart the server.

The request comes in thru Rack. I am not entirely sure how it get exposed tho. I tried reading the sinatra source and it wasn't super clear.

app/features-json/api_controller.rb
```diff
@@ -0,0 +1,127 @@
...
+  # Usage
+  # ===
+  # Subclass the APIController for any controller that you expect JSON requests and responses.
+  #
+  # Settings
+  # ===
+  #
+  # The following settings are provided to subclasses of APIController:
+  #
+  # * `authentiation`- boolean value that makes every request check authentication in a before filter.
```
KrauseFx: `authentiation` spelling mistake

app/features-json/api_controller.rb
```diff
@@ -0,0 +1,127 @@
...
+  # * `authenticate_via` - which authentication scheme to use. `:jwt` by default.
+  # * `jwt_secret` - The key to use in decoding JWT tokens.
+  #
+  # Per-route Authentication
+  # ===
+  # If you disable authentication for the whole controller, you can enable it on a per-route basis using
+  # a route condition like this:
+  #
+  #   get '/private', authenticate: :jwt do
+  #     json({message: 'secret'})
```
KrauseFx: This sample wouldn't pass `rubocop`

app/features-json/api_controller.rb
```diff
@@ -0,0 +1,127 @@
+require "sinatra/base"
+require "sinatra/reloader"
+require "jwt"
+
+require_relative "../services/services"
+require_relative "json_params"
+
+module FastlaneCI
+  ##
+  # APIController is designed for state-less API requests servicing the Angular front-end
```
KrauseFx: I love those comments, great work!

app/features-json/api_controller.rb
```diff
@@ -0,0 +1,127 @@
...
+    helpers do
+      # Decode the JWT or halt.
+      def jwt
+        authorization = request.env["HTTP_AUTHORIZATION"]
+        bearer_token = authorization && authorization.slice(7..-1) # strip off the `Bearer `
+
+        payload, _header = JWT.decode(
+          bearer_token,
+          settings.jwt_secret,
+          true, # Validate Issuer?
```
KrauseFx: What does the comment mean?

snatchev: The comment was meant to indicate that the `true` parameter will make the `decode` method validate the issuer.

app/features-json/api_controller.rb
snatchev: Good catch.

app/features-json/api_controller.rb
KrauseFx: Ah got it, I thought it's a question, ha :)

Issue #892, opened 2018-05-18
Adding new project leads to git error


nakhbari:
```
[2018-05-18 11:17:38] ERROR GitRepo:  Error commiting changes to ci-config repo
[2018-05-18 11:17:38] ERROR GitRepo:  git '--git-dir=/Users/nakhbari/.fastlane/ci/fastlane-ci-config/.git' '--work-tree=/Users/nakhbari/.fastlane/ci/fastlane-ci-config' push 'origin' '{:branch=>"master", :force=>true}'  2>&1:error: src refspec {:branch=>"master",  does not match any.
error: failed to push some refs to 'https://github.com/nakhbari/HelloWorldCiConfig'
Thread terminated with exception: git '--git-dir=/Users/nakhbari/.fastlane/ci/fastlane-ci-config/.git' '--work-tree=/Users/nakhbari/.fastlane/ci/fastlane-ci-config' push 'origin' '{:branch=>"master", :force=>true}'  2>&1:error: src refspec {:branch=>"master",  does not match any.
error: failed to push some refs to 'https://github.com/nakhbari/HelloWorldCiConfig'
  /Users/nakhbari/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/ruby-git-166273eac44a/lib/git/lib.rb:939:in `command'
  /Users/nakhbari/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/ruby-git-166273eac44a/lib/git/lib.rb:744:in `push'
  /Users/nakhbari/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/ruby-git-166273eac44a/lib/git/base.rb:329:in `push'
  /Users/nakhbari/Documents/fastlane/ci/app/shared/models/git_repo.rb:274:in `setup_repo'
  /Users/nakhbari/Documents/fastlane/ci/app/shared/models/git_repo.rb:152:in `block in initialize'
  /Users/nakhbari/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/TaskQueue-5b3f459be5c4/lib/queue_worker.rb:82:in `block in process'
[...]

Issue #727, opened 2018-04-23
Unable add new project


HannesVDB:
I tried to add a new project as I wanted to investigate the Fastlane CI tool.

I got this error
```
`TypeError - no implicit conversion of Symbol into Integer:
	/Users/hannesvandenberghe/Developer/ci/app/shared/fastfile_peeker.rb:74:in `[]'
	/Users/hannesvandenberghe/Developer/ci/app/shared/fastfile_peeker.rb:74:in `block in fastfile_from_contents_map'
	/Users/hannesvandenberghe/Developer/ci/app/shared/github_handler.rb:78:in `github_action'
	/Users/hannesvandenberghe/Developer/ci/app/shared/fastfile_peeker.rb:70:in `fastfile_from_contents_map'
	/Users/hannesvandenberghe/Developer/ci/app/shared/fastfile_peeker.rb:115:in `fastfile_from_github'
	/Users/hannesvandenberghe/Developer/ci/app/shared/fastfile_peeker.rb:21:in `fastfile'
	/Users/hannesvandenberghe/Developer/ci/app/features/project/project_controller.rb:135:in `block in <class:ProjectController>'
	/Library/Ruby/Gems/2.3.0/gems/sinatra-2.0.1/lib/sinatra/base.rb:1634:in `call'
	/Library/Ruby/Gems/2.3.0/gems/sinatra-2.0.1/lib/sinatra/base.rb:1634:in `block in compile!'
	/Library/Ruby/Gems/2.3.0/gems/sinatra-2.0.1/lib/sinatra/base.rb:992:in `block (3 levels) in route!'
	/Library/Ruby/Gems/2.3.0/gems/sinatra-2.0.1/lib/sinatra/base.rb:1011:in `route_eval'
	/Library/Ruby/Gems/2.3.0/gems/sinatra-2.0.1/lib/sinatra/base.rb:992:in `block (2 levels) in route!'
	/Library/Ruby/Gems/2.3.0/gems/sinatra-2.0.1/lib/sinatra/base.rb:1039:in `block in process_route'
	/Library/Ruby/Gems/2.3.0/gems/sinatra-2.0.1/lib/sinatra/base.rb:1037:in `catch'
	/Library/Ruby/Gems/2.3.0/gems/sinatra-2.0.1/lib/sinatra/base.rb:1037:in `process_route'
	/Library/Ruby/Gems/2.3.0/gems/sinatra-2.0.1/lib/sinatra/base.rb:990:in `block in route!'
[...]

Issue #677, opened 2018-04-17
Issue when trying to create a new project.


minuscorp:
Also, I'm getting another issue when trying to create a new project.
```
Git::GitExecuteError at /projects_erb/minuscorp/fastlane-ci-demoapp/add
git '--git-dir=/Users/GabrielVillarrubia/.fastlane/ci/fastlane-ci-config/.git' '--work-tree=/Users/GabrielVillarrubia/.fastlane/ci/fastlane-ci-config' push 'origin' '{:branch=>"master", :force=>true}' 2>&1:error: src refspec {:branch=>"master", does not match any. error: failed to push some refs to 'https://github.com/minuscorp/ci-config-1'
```
### Backtrace
```
/Users/GabrielVillarrubia/Documents/ci/app/shared/models/git_repo.rb in setup_repo
                  git.push("origin", branch: "master", force: true) unless GitRepo.pushes_disabled?
/Users/GabrielVillarrubia/Documents/ci/app/shared/models/git_repo.rb in block in initialize
        setup_repo
```

taquitos:
I've disabled the credential cleanup code in master, could you pull and try again?

minuscorp:
As soon as I arrive home I try it🙌

minuscorp:
It seems that resetting the whole `ci-config` repo did the trick!

PR #893, opened 2018-05-18
Git repo pushes with correct args

nakhbari:
Fixes: #892, #727, #677

### DIFF
--- a/app/shared/models/git_repo.rb
+++ b/app/shared/models/git_repo.rb
@@ -271,7 +271,7 @@ module FastlaneCI
                 begin
                   repo.add(all: true)
                   repo.commit("Sync changes")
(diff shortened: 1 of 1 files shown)


app/shared/models/git_repo.rb
```diff
@@ -271,7 +271,7 @@ def setup_repo
                 begin
                   repo.add(all: true)
                   repo.commit("Sync changes")
-                  git.push(GitRepo::DEFAULT_REMOTE, branch: "master", force: true) unless GitRepo.pushes_disabled?
+                  git.push(GitRepo::DEFAULT_REMOTE, "master", force: true) unless GitRepo.pushes_disabled?
```
taquitos: Have you tried this with a different branch name? Like, just hardcoded "my_branch"?

nakhbari: No i haven't. Is there a scenario you're worried about?

PR #894, opened 2018-05-18
Update PR template to call out steps you need to check off

taquitos:


AndrewMcBurney:
nice 👌 

screen shot 2018-05-18 at 3 17 38 pm

PR #895, opened 2018-05-18
Convert folder path to string

nakhbari:


KrauseFx:
Don't forget to delete merged branches @nakhbari 👍

PR #900, opened 2018-05-21
Add coverage support for coveralls

minuscorp:
- [x] Run tests fine.
~Use PhantomJS as testing browser~
- [x] Test that actually pushes coverage reports to coveralls
```
=============================== Coverage summary ===============================
Statements   : 96.52% ( 527/546 )
Branches     : 63.11% ( 260/412 )
Functions    : 97.5% ( 117/120 )
Lines        : 98.93% ( 461/466 )
================================================================================
22 05 2018 20:59:33.358:INFO [coveralls.io]: uploading...
22 05 2018 20:59:33.359:INFO [coveralls.io]: 422 --- {"message":"Couldn't find a repository matching this job.","error":true}
```


KrauseFx:
Would that work together with our Ruby based tests that also use Coveralls? Is Coveralls going to show it as one coverage?

minuscorp:
Those are questions that I have made for myself and couldn't figure out how to solve them. There's some kind of token or something to upload the report to coveralls or how does the platform manage it?

minuscorp:
From `node-coveralls` documentation:

> This library currently supports travis-ci with no extra effort beyond piping the lcov output to coveralls. However, if you're using a different build system, there are a few environment variables that are necessary:

> COVERALLS_SERVICE_NAME (the name of your build system)
COVERALLS_REPO_TOKEN (the secret repo token from coveralls.io)
There are optional environment variables for other build systems as well:

> COVERALLS_SERVICE_JOB_ID (an id that uniquely identifies the build job)
COVERALLS_RUN_AT (a date string for the time that the job ran. RFC 3339 dates work. This defaults to your build system's date/time if you don't set it.)
COVERALLS_PARALLEL (more info here: https://coveralls.zendesk.com/hc/en-us/articles/203484329)

So I think we need to provide more information (that I cannot know at this point) to complete the deploy of the coverage results (I have confirmed that the reports are being created but it seems that are not being sent)

karma.conf.js
```diff
@@ -19,15 +27,18 @@ module.exports = function (config) {
...
     },
-    reporters: ['progress', 'kjhtml'],
+    reporters: ['progress', 'kjhtml', 'coverage-istanbul', 'coverage', 'coveralls' ],
     port: 9876,
     colors: true,
     logLevel: config.LOG_INFO,
     autoWatch: true,
     browsers: ['Chrome'],
-    singleRun: false
+    singleRun: !argv.watch,
```
nakhbari: I'm not sure i follow this line. `singleRun` is the default right? The `watch` arg is overwritting the default, why do we need to have it here as well?

minuscorp: Yeah, maybe I should delete autowatch in favor of singleRun by arguments. What do you think?

nakhbari: What's the worry here? Trying to understand why we need to change singleRun at all

nakhbari: isn't this what the `--watch` flag will do?

minuscorp: So why do they define two different properties for the same purpose? 😕

nakhbari: Yeah it's weird. I saw bugs where you can't have both `singleRun` and `autoWatch` set to true. 🤷‍♂️ 

I think everything else looks good from my point of view. Let's make this change and get some code coverage action!

minuscorp: - `singleRun`: Continuous Integration mode.
If true, Karma will start and capture all configured browsers, run tests and then exit with an exit code of 0 or 1 depending on whether all tests passed or any tests failed.
- `autoWatch`: Enable or disable watching files and executing the tests whenever one of these files changes.

minuscorp: Which flag do I delete?

nakhbari: I was thinking that we don't need the `!argv.watch`. Maybe i'm misunderstanding something. Let's chat today or Tuesday (long-weekend here)?

minuscorp: I'm abroad until next Thursday, I'll make tests with and without the flag and let you know the best result and then we can merge or chat if we need to sync😁🙌🙌 sounds good to you?

PR #905, opened 2018-05-21
Update dependencies to use proper versioning and use GC fork

KrauseFx:
Also really happy we have that written down now, so we can always reference it


snatchev:
@KrauseFx I added https://github.com/fastlane/ci/pull/909 in order to minimize this pattern in the future :)

KrauseFx:
Cool, thanks

Issue #886, opened 2018-05-17
Fix xcode-install tests


KrauseFx:
Via @taquitos via https://github.com/fastlane/ci/pull/878

I'm gonna fix this

> When you get a chance: https://github.com/fastlane/ci/pull/878
> The xcode-select specs don’t work, so stefan disabled them
> Also, I ran into crashes, I think because the initializer of `EnvironmentVariableService` needs a user or apple_id and that isn’t provided because in `services.rb` you use an ENV var of `ENV["FASTLANE_USER"]`  which we don’t have setup.

PR #907, opened 2018-05-21
Improve the way we pass AppleID credentials to xcode-install

KrauseFx:
This also fixes + updates our unit tests

Fixes https://github.com/fastlane/ci/issues/886

### DIFF
--- a/app/services/xcode_manager_service.rb
+++ b/app/services/xcode_manager_service.rb
@@ -68,7 +68,7 @@ module FastlaneCI
     # the resulting list is sorted with the most recent release as first element
     def available_xcode_versions
       apple_id_credentials_block do
-        return installer.seedlist.reverse
+        installer.seedlist.reverse
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/services/xcode_manager_service_spec.rb
```diff
@@ -107,8 +137,7 @@
       })
     end
 
-    # TODO: fix this
-    xit "raises an exception if installation is already in progress" do
+    it "raises an exception if installation is already in progress", nower: true do
```
snatchev: what's `nower` ?

app/services/xcode_manager_service.rb
```diff
@@ -67,7 +67,9 @@ def xcode_queue
     #
     # the resulting list is sorted with the most recent release as first element
     def available_xcode_versions
-      return installer.seedlist.reverse
+      apple_id_credentials_block do
+        return installer.seedlist.reverse
```
snatchev: I would like to discourage the use of `return` inside blocks. They can lead to surprising behavior. Correct me if I am wrong, but this will not reach the `ENV.delete` methods in the `ensure` part of the block.

KrauseFx: Ohhh right, nice catch!!

KrauseFx:
Updated

PR #912, opened 2018-05-22
 Inject actual build URL to ENV variables

KrauseFx:


app/features/build_runner/build_runner.rb
```diff
@@ -201,6 +201,7 @@ def setup_build_specific_environment_variables
       @environment_variables_set = []
 
       # Set the CI specific Environment variables first
+      build_url = File.join(Services.dot_keys_variable_service.ci_base_url, "projects", project.id, "builds", build.number)
```
snatchev: This is a URL right? We should use https://ruby-doc.org/stdlib-1.9.3/libdoc/uri/rdoc/URI.html#method-c-join instead of `File.join` since there are some escaping issues that might be handled better by the `URI` module.

snatchev: If you decide to use `URI` don't forget to `to_s` at the end, because it's not a string object.

KrauseFx: I just realized we have this already in 

https://github.com/fastlane/ci/blob/master/app/features/build/build_controller.rb#L23-L25

What part of the system is responsible for building URLs?

snatchev: As far as I know, there is no system for building URLs in Sinatra or in our own code.

snatchev: @KrauseFx I was wrong. Sinatra does help a helper method for constructing urls
https://github.com/sinatra/sinatra/blob/master/lib/sinatra/base.rb#L286-L303

KrauseFx: Perfect

KrauseFx: So should we apply any changes to this PR then?

PR #919, opened 2018-05-23
Fix crash in xcode manager if no user is provided

KrauseFx:


taquitos:
@snatchev that's a great point, setting things at initialization and not mutating is  🔑

KrauseFx:
Okay, we have the `XcodeManagerService` in our `Services` class in `fastlane.ci`. Do you propose we create a new service every time? 

Right now, we only set the value during init time, as we don't have a UI/API yet to set a default Apple ID, but in the future this will be possible. Either way is fine for me. I'd propose to go ahead and merge this PR assuming this is still a problem, and create an issue to discuss this further

snatchev:
I think I might not have explained my initial suggestion. Basically, I propose have a method that returns the `apple_id`, like this:

```ruby
def apple_id(user:)
  apple_id = Services.apple_id_service.apple_ids.find { |a| a.user == user } if user
  if apple_id.nil?
    # Think about the user flow on how this would be shown, we probably want to check
    # for the existance of the Apple ID earlier, and then not even show the button,
    # or make the button redirect to the Apple ID login instead
    raise "No registered Apple ID found with user #{user}, make sure to add your Apple account to fastlane.ci"
  end
  return apple_id
end
```

and we set `@apple_id = apple_id(user: user)` in the initializer. The benefit being that instance state is set in 1 place in the initializer which is a little more clear, IMO.

I also didn't see any calls to `use_apple_id` with an `apple_id:` parameter. If that's unused, we should remove it.

KrauseFx:
Alright, would hit the merge button now, but @taquitos would tap on shoulder and slap on finger, please merge as soon as CI is ready again

Issue #917, opened 2018-05-23
Build Details Endpoint


nakhbari:
Parent: #709 

Requirements:
- Status
- Trigger
- Branch
- Sha
- Start Time
- Duration time
- Build Tools
- Environment variables?

KrauseFx:
Cool, taking on this task now 👍

PR #923, opened 2018-05-23
Add initial BuildJSONController for build details

KrauseFx:
screen shot 2018-05-23 at 3 30 31 pm

Fixes https://github.com/fastlane/ci/issues/917

### DIFF
--- a/app/features-json/view_models/build_view_model.rb
+++ b/app/features-json/view_models/build_view_model.rb
@@ -1,0 +1,76 @@
+require_relative "../../shared/models/build"
+require_relative "../../shared/json_convertible"
+
+module FastlaneCI
+  # View model to expose the detailed info about a build.
+  class BuildViewModel
+    include FastlaneCI::JSONConvertible
+
+    # The project ID this build is associated with
+    attr_reader :project_id
+
+    # @return [Integer]
+    attr_reader :number
+
+    # @return [String]
+    attr_reader :status
+
+    # @return [DateTime] Start time
+    attr_reader :timestamp
+
+    # @return [Integer]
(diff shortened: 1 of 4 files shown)


app/features-json/view_models/project_view_model.rb
```diff
@@ -3,7 +3,7 @@
 require_relative "../../shared/json_convertible"
 
 module FastlaneCI
-  # View model to expose the basic info about a project.
+  # View model to expose the detailed info about a project.
```
KrauseFx: @nakhbari can you confirm this is accurate? This and another class had the same class level description, and I wasn't 100% certain when which one is used

nakhbari: yep this is good 👍

app/features-json/view_models/build_view_model.rb
```diff
@@ -0,0 +1,76 @@
...
+    attr_reader :build_tools
+
+    # @return [String] The git URL
+    attr_reader :clone_url
+
+    # @return [String] The git branch
+    attr_reader :branch
+
+    # @return [String] The git ref
+    attr_reader :ref
```
nakhbari: @KrauseFx what's the git ref?

KrauseFx: We have some more docs on this over here https://github.com/fastlane/ci/blob/master/app/shared/models/git_fork_config.rb#L19-L21

I don't know for sure when this gets filled out, as every `ref` can also be represented by a `sha`. Maybe @taquitos knows more, as he invented the class `GitForkConfig`

PR #927, opened 2018-05-24
Add Artifacts JSON API

KrauseFx:


app/features-json/artifact_json_controller.rb
```diff
@@ -0,0 +1,79 @@
+require_relative "./json_authenticated_controller_base"
+require_relative "./view_models/artifact_view_model"
+
+module FastlaneCI
+  # Controller for providing all data relating to artifacts
+  class ArtifactJSONController < JSONAuthenticatedControllerBase
+    HOME = "/data/project/:project_id/build/:build_number/artifact"
+
+    class ArtifactNotFoundError < FastlaneCIError
```
snatchev: Lets use `halt(404, header, body)` instead of raise and catch. halting uses throw/catch so that we don't have to worry about proceeding further in the method.

app/features-json/artifact_json_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+
+    # Fetch all the available metadata for a given artifact, the return data is JSON
+    get "#{HOME}/:artifact_id/info" do |project_id, build_number, artifact_id|
+      begin
+        artifact, = fetch_artifact_details(
```
snatchev: if destructuring and we have an un-used parameter, please use `_` instead of leaving it blank.

app/features-json/artifact_json_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    private
+
+    def fetch_artifact_details(project_id:, build_number:, artifact_id:)
+      project = user_project_with_id(project_id: project_id)
+      build = project.builds.find { |b| b.number == build_number.to_i }
+
+      artifact = build.artifacts.find { |find_artifact| find_artifact.id == artifact_id }
+
+      raise ArtifactNotFoundError if artifact.nil?
```
snatchev: This can be re-written as
```ruby
halt(404, {"Content-Type": "application/json"}, "Artifact not found") if artifact.nil?
```

app/features-json/artifact_json_controller.rb
```diff
@@ -0,0 +1,79 @@
+require_relative "./json_authenticated_controller_base"
+require_relative "./view_models/artifact_view_model"
+
+module FastlaneCI
+  # Controller for providing all data relating to artifacts
+  class ArtifactJSONController < JSONAuthenticatedControllerBase
```
snatchev: `JSONAuthenticatedControllerBase` => `APIController`

app/features-json/artifact_json_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+      end
+    end
+
+    private
+
+    def fetch_artifact_details(project_id:, build_number:, artifact_id:)
+      project = user_project_with_id(project_id: project_id)
+      build = project.builds.find { |b| b.number == build_number.to_i }
+
+      artifact = build.artifacts.find { |find_artifact| find_artifact.id == artifact_id }
```
minuscorp: We should be using always `ArtifactProvider`'s `retrieve!` method to get the `Artifact` reference

KrauseFx:
> 2 Questions:

> I'm not too familiar with how our artifact system works. Is it always hosted in some cloud storage or do we have a local destination on the server host that stores it.

Yeah so we haven't fully decided that yet. I think we'll allow local storage for simple self-hosted setups if the user just has a single Mac Mini, however I don't know how this will play along with @snatchev's works on separating the runner from the server. 
So the user can choose to store the artifacts locally, or provide a Google API token to upload them to Google Cloud. More info should be in our artifacts design doc https://github.com/fastlane/ci/blob/master/docs/ArtifactsAndBuilds.md

> What's the reasoning for splitting the download portion into a separate API instead of having the uri passed alongside the artifacts response?

Can you tell me a bit more what you mean? Like why do we have 2 different API endpoints? The thought was that eventually we might have additional metadata for each of the artifacts, so we'll have an `/info` API endpoint to get all the data available for a given artifact, and we have the `/download` API endpoint, which will provide the raw download for the given artifact.

The `/info` would also return an exception if the artifact can't be found any more (e.g. if an AWS S3 bucket got deleted)

Or did you mean something else?

nakhbari:
>we'll have an `/info` API endpoint to get all the data available for a given artifact, and we have the `/download` API endpoint, which will provide the raw download for the given artifact.

Ah i thought `/download` endpoint returns the download URL, since it could potentially be hosted elsewhere, but I see that it send the artifact back. Does this mean that if the file is hosted we download it locally then send that local file? If so, that would be twice the work right?

What I was imagining was a `../build/:build_number/artifacts` endpoint that would list the artifacts. The response would have the `download_url` whether it be hosted locally or in the cloud. 

Ex.
```json
[{
  "id": "123",
  "name": "best-file",
  "download_url": "https://cloud.storage.google.com/best-bucket/best-folder/best-file"
},
{
  "id": "1234",
  "name": "bestest-file",
  "download_url": "http://localhost:8080/api/artifact/1234/download"
}]
```

The difference here is that the web-app can be dumb and just reference whichever link the backend provides. And we wouldn't have to have the backend download the cloud hosted file and reserve it with it's own endpoint, reducing our traffic and the effective latency.

KrauseFx:
Yeah I like that. I'm happy to update the API to provide the URL like this 👍 `best-bucket` 🚀

KrauseFx:
Ready to be reviewed & merged from my side 👍

app/features-json/artifact_json_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+      end
+
+      uri = URI.parse(artifact_reference)
+
+      return artifact, artifact_reference, uri
+    end
+
+    def current_project
+      current_project = FastlaneCI::Services.project_service.project_by_id(params[:project_id])
+      halt(404) unless current_project
```
taquitos: Same here, do we want to update it for the new error handling?

app/shared/models/artifact.rb
```diff
@@ -12,7 +12,8 @@ class Artifact
     # @return [UUID] The id of the artifact.
     attr_reader :id
 
-    # type is just a String that refers to the kind of data that the Artifact represents (i.e., ipa, apk, log, etc.)
+    # type is just a String that refers to the kind of data that the Artifact represents
```
taquitos: can remove the extra words: `type is just a`

app/features-json/artifact_json_controller.rb
```diff
@@ -0,0 +1,86 @@
+require_relative "./view_models/artifact_view_model"
+
+module FastlaneCI
+  # Controller for providing all data relating to artifacts
+  class ArtifactJSONController < APIController
```
taquitos: Cool, auth in enabled on these end points ❤️

KrauseFx: What do you mean?

taquitos: I mean, these end points require JWT auth in order to allow you to access them, which I think is the right move here.

app/features-json/artifact_json_controller.rb
KrauseFx: Yeah, because I renamed the method on the `update-api-status-codes` branch, I was thinking of doing it in a separate after merging both to always be in a clean state

Issue #870, opened 2018-05-15
[Proposal] System to allow system-wide preferences


KrauseFx:
Without having a clear specification of this feature yet, I want to get the discussion started.

It seems like we'll have some system settings that are system wide, but don't deserve their own configuration file. Some examples I can think of:

- Default Xcode version
- Default Apple ID to use (e.g. for installing Xcode)
- Boolean options like:
  - Automatically install new versions of Xcode
  - Automatically maintain local simulators
  - Disable metric collections
- fastlane.ci web host URL (`FASTLANE_CI_BASE_URL`)
- Many more in the future probably

Should we introduce a `settings.json` in the root of the `ci-config` repo that contains all those attributes? If so, we should have a system built-in to override any value on a project/branch level (branch overrides can be post 1.0 feature)

We already have project-level overrides for ENV variables with https://github.com/fastlane/ci/pull/768, so we could something similar in a more generic way.

Ideally the `settings.json` is a hash that allows any key/value pair that we define in a single `.rb` file, as it should be easy for us to add new options without having to write duplicate code. 

A `settings.json` for those settings will also allow us to tell the user to modify the JSON file directly, for hidden options (options that we support, but don't have a full UI yet), so we can iterate faster and show what's most relevant to the user, while still putting the user in full control over everything.

---

### Proposal

Based on the description above it seems like we could do the following:

Simplified Ruby interface
```ruby
class Settings
  attr_accessor :metrics_disabled
  attr_accessor :install_new_versions_of_xcode
  attr_accessor :default_apple_id
  ...
end
```
those are automatically mapped to be stored in a `settings.json` file:
```json
{
  "metrics_disabled": false,
  "install_new_versions_of_xcode": true,
  "default_apple_id": "apple@krausefx.com"
}
```
we could even use JSON comments when generating the file during the setup to allow anyone to modify the values without even having to read the docs. There are multiple ways to provide comments in JSON, either by not following the JSON standard and using `/* ... */` or by wrapping the value in another hash:
```json
{
  "metrics_disabled": {
    "value": false,
    "comment": "Don't collect any metrics, more information on privacy at https://privacy.fastlane.ci"
  },
  "default_apple_id": {
    "value": "apple@krausefx.com",
    "comment": "What Apple ID should be used to install Xcode"
  }
}
```

[...]

PR #935, opened 2018-05-25
Add management of system settings

KrauseFx:
Fixes https://github.com/fastlane/ci/issues/870

This PR implements the global fastlane.ci settings. The user will have a `settings.json` in the `ci-config` repo that looks like this

```json
[
  {
    "key": "metrics_enabled",
    "description": "Set this to false to opt out of all metrics tracking, more information on https://github.com/fastlane/ci",
    "value": false
  },
  {
    "key": "default_apple_id",
    "description": "The default Apple ID to use for certain tasks, like installing new versions of Xcode",
    "value": "apple@krausefx.com"
  }
]
```

We define a list of available options, looking like this:
```ruby
    def self.available_settings
      [
        Setting.new(
          key: :metrics_enabled,
          default_value: true,
          description: "Set this to false to opt out of all metrics tracking, more information on https://github.com/fastlane/ci"
        ),
        Setting.new(
          key: :default_apple_id,
          default_value: nil,
          description: "The default Apple ID to use for certain tasks, like installing new versions of Xcode",
          verify_block: proc do |value|
            apple_id = Services.apple_id_service.apple_ids.find { |a| a.user == value }
            raise "Couldn't find Apple ID with email '#{value}'" if apple_id.nil?
          end
        )
      ]
    end
```

Having the `verify_block` enables us to catch user errors early and show a proper error message as a result.

Using this approach we can add more options to fastlane.ci, but don't necessarily have to show them in the user interface if we think an option is only valuable to a small subset of users.

To access a given setting, you can use
```ruby
Services.setting_service.find_setting(setting_key: :metrics_enabled).value
```

To modify a given setting, you can use
```ruby
metrics_setting = Services.setting_service.find_setting(setting_key: :metrics_enabled)
metrics_setting.value = false
Services.setting_service.update_setting!(setting: metrics_setting)
```

To reset a given setting to the default value, you can use
```ruby
Services.setting_service.reset_setting!(setting_key: params[:setting_key])
```

The above isn't written in stone yet, we can provide an easier interface to edit values if necessary.

Having the `default_value` allows us to provide the `reset_setting!` method. If a user explicitly set a value we won't change it for them, but if they never did, we can change default values via an update, with confidence that we're not breaking existing setups


app/features-json/setting_json_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  class SettingJSONController < APIController
+    HOME = "/data/settings"
+
+    get HOME do
+      return json(Services.setting_service.settings)
+    end
+
+    post HOME do
+      metrics_setting = Services.setting_service.find_setting(setting_key: :metrics_enabled)
+      metrics_setting.value = false
```
taquitos: always false, let's pull this from the post body

app/features-json/setting_json_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+      Services.setting_service.update_setting!(setting: metrics_setting)
+
+      return json({ status: :success })
```
taquitos: how would we determine if this failed? Would we need to catch an exception or something?

app/features-json/setting_json_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    delete "#{HOME}/:setting_key" do
+      begin
+        Services.setting_service.reset_setting!(setting_key: params[:setting_key])
+        return json({status: :ok})
```
taquitos: I think we should standardize either `status: :ok` or `status: :success`, right?

app/services/setting_service.rb
```diff
@@ -0,0 +1,70 @@
...
+    def initialize(setting_data_source: nil)
+      unless setting_data_source.nil?
+        unless setting_data_source.class <= SettingDataSource
+          raise "setting_data_source must be descendant of #{SettingDataSource.name}"
+        end
+      end
+
+      if setting_data_source.nil?
+        # Default to JSONSettingDataSource
+        # TODO: do we need `sample_data` here?
```
taquitos: No, I think we don't need a `sample_data` here. If there is a missing file, we should create it with default settings.

app/services/setting_service.rb
```diff
@@ -0,0 +1,70 @@
...
+        logger.debug(
+          "setting_data_source is new, using `ENV[\"data_store_folder\"]` " \
+          "if available, or `sample_data` folder"
+        )
+        data_store_folder = ENV["data_store_folder"] # you can set it at runtime!
```
taquitos: We should revisit this separately. `data_store_folder` is used in multiple places and might not be what we want.

app/services/setting_service.rb
```diff
@@ -0,0 +1,70 @@
...
+        data_store_folder ||= File.join(FastlaneCI::FastlaneApp.settings.root, "sample_data")
+        setting_data_source = JSONSettingDataSource.create(data_store_folder)
+      end
+
+      self.setting_data_source = setting_data_source
```
taquitos: I think this should be named `settings_data_source` instead of the singular: `setting_data_source`

app/shared/models/setting.rb
```diff
@@ -0,0 +1,34 @@
+module FastlaneCI
+  # Setting model representing a single entry of a fastlane.ci system setting
+  class Setting
+    attr_accessor :key
+    attr_accessor :value
+    attr_accessor :verify_block
```
taquitos: just #fastlane things ❤️

app/features-json/setting_json_controller.rb
KrauseFx: Hehe, yah, this is just code to try out setting a value using an HTTP request, this will be removed for sure 👍

app/features-json/setting_json_controller.rb
KrauseFx: Yeah, I'm all for being consistent, which way do you prefer?

app/services/setting_service.rb
KrauseFx: Cool, we actually have the same code in all the other `_service.rb` files, I'll submit a separate PR that removes it from them also 👍

KrauseFx: Created https://github.com/fastlane/ci/issues/982

app/services/setting_service.rb
KrauseFx: Nice catch, will update 👍

app/shared/models/setting.rb
KrauseFx: Haha yeah, we're slowly building `ConfigItem` again :trollface:

KrauseFx:
Working on this now

app/features-json/setting_json_controller.rb
KrauseFx: Yah, right now we have the `verify_block` which could potentially raise an exception, so the caller of `setting_name.value = something` has to handle this

app/services/setting_service.rb
KrauseFx: Actually no, seems like we're using the singular everywhere, including https://github.com/fastlane/ci/blob/master/app/services/build_service.rb#L16 and https://github.com/fastlane/ci/blob/master/app/services/environment_variable_service.rb#L30

app/features-json/setting_json_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+module FastlaneCI
+  # Controller for providing all data relating to system settings
+  class SettingJSONController < APIController
+    HOME = "/data/settings"
+
+    get HOME do
+      return json(Services.setting_service.settings)
+    end
+
+    post "#{HOME}/:setting_key" do
```
KrauseFx: I couldn't test this locally for now, as I'm at home and can't get auth to work. Also, I don't think we have a way to add tests for authenticated API endpoints yet

app/features-json/setting_json_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+      # TODO: security aspect, how do we make sure this can't be abused?
+      # We don't want to hash/encrypt all of them, as this would make it harder for
+      # people to manually update those files
+      begin
+        setting.value = value
+        Services.setting_service.update_setting!(setting: setting)
+
+        return json({ status: :success })
+      rescue StandardError => ex
+        return json({ status: :error, message: ex.to_s })
```
snatchev: This will return with a 200 status. Sinatra providers an error handler, but we don't use it yet. I think we should let the error bubble up and then we add an error handler in `APIController` that will return a JSON error, with the proper status codes.

app/features-json/setting_json_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+      end
+    end
+
+    delete "#{HOME}/:setting_key" do
+      begin
+        # TODO: security for parameters as above
+        Services.setting_service.reset_setting!(setting_key: params[:setting_key])
+        return json({ status: :success })
+      rescue SettingServiceKeyNotFoundError
+        return json({ status: :key_not_found })
```
snatchev: Same here, we should either ```halt(404, "`#{params[:setting_key]}` not found.")``` or let the error bubble up.

app/services/data_sources/json_setting_data_source.rb
```diff
@@ -0,0 +1,99 @@
...
+      reload_settings
+    end
+
+    def setting_file_path(path: "settings.json")
+      return File.join(json_folder_path, path)
+    end
+
+    # TODO: make synchronize work again
+    def settings
+      # JSONSettingDataSource.file_semaphore.synchronize do
```
snatchev: does this not work? It looks like it's used in line 49.

app/services/setting_service.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  # Provides access to system settings
+  class SettingService
+    include FastlaneCI::Logging
+    attr_accessor :setting_data_source
+
+    def initialize(setting_data_source: nil)
+      unless setting_data_source.nil?
+        unless setting_data_source.class <= SettingDataSource
```
snatchev: nit: I think we should use `kind_of?` instead of this comparison operator. No need to change, we should update rubocop instead.

KrauseFx: I'm all for kind_of? I use it usually. In this case the code is copied from all the other services we already use to be consistent.

app/services/data_sources/json_setting_data_source.rb
KrauseFx: Yes, worked when I tested it 👍 Added an open todo for when we work on the file sync and git commits

KrauseFx:
Let's 🚀is

Issue #709, opened 2018-04-20
Build Details Page


minuscorp:
##### UX
- [x] UI Wireframe #685

##### Backend
- [x] Build Details Endpoint #917 
- [x] Websocket endpoint for logs https://github.com/fastlane/ci/issues/890
- [ ] Websocket Logs are backfilled #1041
- [ ] Artifacts Endpoint #918 - blocked by #951 

##### Frontend
- [ ] Build logs formatting #1054
- [ ] Close websocket when getting `EOT` 
- [x] Connect to websocket from client #786
- [x] Display Build details
- [x] Display Logs in solarized dark color container
- [x] Display Artifacts
- [x] Display Build details
- [x] New Route for build #719 
- [x] Link the commit shas to their github page #708 
- [x] Relative Pipe for the latest build Date #666
- [x] Show error message on CI failure
- [x] Loading spinner while loading data #668

#### Mock
![image](https://user-images.githubusercontent.com/2754461/40148645-ee6f785e-5923-11e8-91d8-cfcc3a290258.png)

PR #937, opened 2018-05-25
Build page with details and placeholder for logs

nakhbari:
Issue: #709 

```
```

#### Screenshot
image

### DIFF
--- a/web/app/build/build.component.html
+++ b/web/app/build/build.component.html
@@ -1,4 +1,33 @@
 <fci-toolbar [breadcrumbs]="breadcrumbs"></fci-toolbar>
-
-<h4>Logs:</h4>
-<pre>{{logs}}</pre>
+<div class="fci-build-container">
+  <div class="fci-left-container">
+    <div class="fci-build-header">
+      <fci-status-icon *ngIf="build" [status]="build.status"></fci-status-icon>
+      <span class="fci-build-title">Build {{build ? build.number : ''}}</span>
+      <span *ngIf="build && build.description" class="fci-build-description">{{build.description}}</span>
+    </div>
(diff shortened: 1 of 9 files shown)


web/app/build/build.component.html
```diff
@@ -1,4 +1,33 @@
...
+      <mat-spinner *ngIf="!build" mode="indeterminate"></mat-spinner>
+      <ng-container *ngIf="build">
+        <h5>Trigger</h5>
+        <div>{{build.trigger}}</div>
+        <h5>Branch</h5>
+        <div>{{build.branch}}</div>
+        <h5>SHA</h5>
+        <div>{{build.shortSha}}</div>
+        <h5>Started</h5>
+        <div>{{build.date | amTimeAgo}}</div>
```
KrauseFx: 😍

web/app/models/build.ts
```diff
@@ -0,0 +1,59 @@
...
+  readonly date: Date;
+
+  constructor(buildSummary: BuildResponse) {
+    this.number = buildSummary.number;
+    this.projectId = buildSummary.project_id;
+    this.description = buildSummary.description;
+    this.trigger = buildSummary.trigger;
+    this.lane = buildSummary.lane;
+    this.platform = buildSummary.platform;
+    this.parameters = buildSummary.parameters;
```
KrauseFx: Looks like some changes overlap with https://github.com/fastlane/ci/pull/936, something went wrong with the branches?

nakhbari: yeah this should be targeting that branch instead of master. I changed it but doesn't look like it went through.

KrauseFx: 👍

PR #943, opened 2018-05-29
add more specs around api controller and add webmock

snatchev:
This PR adds some specs and api spec helpers.

Also, adds webmock and disables network requests in specs.


KrauseFx (approved):
This is great 🔢

Issue #784, opened 2018-05-01
Implement onboarding


KrauseFx:
This issue will basically include all the tasks needed to implement the fastlane.ci onboarding with the new front-end system 👍 

- [x] Endpoint to determine whether or not the server needs to be configured
- [x] Point to Onboarding page when server needs configuration on web app init
- [x] Onboarding Page Mocks

#### Frontend
- [x] Route to onboarding when server is not configured
- [x] Add stepper to onboarding page
- [x] Add encryption key input
- [x] Add bot account setup
- [x] Add ci config repo setup
- [ ] Add user account setup
- [ ] Send request to configure the server
- [ ] Add 5 char validation to encryption token
- [ ] Route to Dashboard
- [ ] Get proper screenshot of github permissions including email scope
- [ ] Update UI to new mocks

#### Backend
- [ ] JSON POST endpoint to create config repo #1012
- [ ] JSON POST endpoint to set Encryption key
- [ ] JSON POST endpoint to set OAuth info
- [ ] JSON GET to get the clientId
- [ ] JSON GET to get the GitHub auth token from a given code
- [ ] Change Auth in backend to verify that the passed token has access to all the repos required to complete any JSON Request
- [ ] (Lower priority) Endpoint that replies whether or not a config repo exists in the selected org
- [x] JSON GET to get the User's email associated with their account set up #1010

nakhbari:
We're gonna need a way to know when the CI hasn't been setup yet.

I think @taquitos suggested we make a `/onboard` url and indicate in our Repo that they must go there first for now.

taquitos:
I think we should have an endpoint that our frontend queries for an initial payload that contains things like localization info, and other constants that the server should be providing. As part of that payload, we should also indicate whether the server is totally setup and ready.

PR #949, opened 2018-05-31
Check if server is configured on web app init and forward to onboarding

nakhbari:
Issue: #784 

Point to the ERB onboarding for now while we're waiting on the webapp oboarding to be complete


app/features-json/setup_json_controller.rb
```diff
@@ -0,0 +1,12 @@
+require_relative "api_controller"
+
+module FastlaneCI
+  # Controller for providing all setup APIs
+  class SetupJSONController < APIController
+    HOME = "/data/setup"
+
+    get "#{HOME}/configured", authenticate: false do
+      return json(Services.onboarding_service.correct_setup?)
```
nakhbari: it's just a boolean, so i'm not sure we actually need the `json()` method. Please advise.

taquitos: Let's actually return a real object here, maybe a `ConfigurationData`. That way we can add more configuration properties as needed in the future with out having to refactor.

taquitos: Also, let's call this endpoint `"#{HOME}/configuration"`

KrauseFx:
@taquitos with @nakhbari on vacation, we should probably take this over. Looks like we caused some merge conflicts though 😢

app/features-json/setup_json_controller.rb
KrauseFx: Does that come with any privacy/security side effects, if a user has only e.g. very basic read access to some repos? Either way, this is not a blocker - let's merge this PR as Nima is OOO right now

KrauseFx:
Alright, I resolved the merge conflicts 👍

KrauseFx:
Cool, so to avoid having this PR open any longer, I'm gonna make a call here and merge this. As this is an unauthenticated request, I'm in favor of exposing as little information as we can for now. So having a boolean value is okay and does what we need it for. As we implement the onboarding, we can refine or add new API endpoints 👍

PR #951, opened 2018-05-31
[WIP] Distributed fastlane.ci builds

KrauseFx:
Fixes https://github.com/fastlane/ci/issues/457

**Open tasks:**

- [x] Get rid of the Rakefile in the worker and replace with a Ruby class #987 (@snatchev)
- [x] Storing of artifacts, which includes storing & loading previous build outputs #987 (@snatchev)
- [x] Pass on exceptions from the worker to the fastlane.ci #987
- [x] Fix `ws` being `nil` causing a crash (https://github.com/fastlane/ci/issues/826) (@taquitos)
- [x] Rename `Server < Service` class to something else #987 (@snatchev) 

**Moved to separate tasks**
- Seems like we need to flush the output on some part of the stack (https://github.com/fastlane/ci/issues/954)
- Additional new lines at the end (https://github.com/fastlane/ci/issues/955)
- A worker needs a way to tell master that they can't handle a given job (https://github.com/fastlane/ci/issues/958)
- Automatic Xcode installation (https://github.com/fastlane/ci/issues/446) depends on https://github.com/fastlane/ci/issues/958
- Enable certificates for HTTPs encryption for distributed builds
- Use our Git Helper as part of the worker (https://github.com/fastlane/ci/issues/966)
- Discovery or management of worker nodes inside the network (https://github.com/fastlane/ci/issues/967)

PR #957, opened 2018-06-01
Add information about how we unblock Agents during Xcode installations

KrauseFx:
Related to https://github.com/fastlane/ci/pull/951

### DIFF
--- a/docs/WorkerAgent.md
+++ b/docs/WorkerAgent.md
@@ -33,6 +33,17 @@ The Agent may have to fetch resources (such as source code) from other services 
 ## Permission model
 The Agent will be required to run arbitrary Ruby code on the Worker. It may also be required to perform operations that require root such as installing Xcode. While there will be other security mechanisms that will enforce strict security against malicious users, we can help ourselves by preventing accidental (or incidental) damage to the Worker by running the Agents under a special user that belongs to the sudoers group. This will allow us to whitelist the commands and permissions the user can perform on a Worker.
 
+## Rejecting a job
+An Agent must be able to reject a job from the server. This is needed for a common situation around Xcode
+1. Server wants to build a PR
+1. Agent gets job assigned, checks out the repo, and finds an updated `.xcode-version` file specifying an Xcode beta release
+1. Agent checks if the specific Xcode beta is installed on its machine
+    1. If available, use that version to run _fastlane_
+    1. If not available, Agent triggers installation of given Xcode beta release
+        1. Agent notifies Server that it can't handle this particular build right now
+        1. Server puts task back on the task queue and tries to distribute this task again later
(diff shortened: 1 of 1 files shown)


snatchev (approved):

docs/WorkerAgent.md
```diff
@@ -33,6 +33,17 @@ The Agent may have to fetch resources (such as source code) from other services
...
+## Rejecting a job
+An Agent must be able to reject a job from the server. This is needed for a common situation around Xcode
+1. Server wants to build a PR
+1. Agent gets job assigned, checks out the repo, and finds an updated `.xcode-version` file specifying an Xcode beta release
+1. Agent checks if the specific Xcode beta is installed on its machine
+    1. If available, use that version to run _fastlane_
+    1. If not available, Agent triggers installation of given Xcode beta release
+        1. Agent notifies Server that it can't handle this particular build right now
+        1. Server puts task back on the task queue and tries to distribute this task again later
+        1. While Agent installs new Xcode (which might take multiple hours), Agent is free to run other jobs
```
taquitos: Installing Xcode will also block you from using Xcode too, right? So maybe setting itself as unavailable is the way to go?

bogdanbrato: I also have a question related to this? Why notify the server that it can't handle the assigned job instead of streaming back an "Installing xcode..." and resuming the job once xcode install is done? Is it we hope for the server to find another available agent that already has the needed xcode version?

KrauseFx: > Installing Xcode will also block you from using Xcode too, right? So maybe setting itself as unavailable is the way to go?

I don't think that's true. If the worker has Xcode 9.2 installed, and one build requires Xcode 9.4 it would now go ahead and start the download & installation. In the mean-time the worker can use Xcode 9.2 to compile and test apps that require Xcode 9.2. This is important for when one branch only for example requires a new version of Xcode

> I also have a question related to this? Why notify the server that it can't handle the assigned job instead of streaming back an "Installing xcode..." and resuming the job once xcode install is done? Is it we hope for the server to find another available agent that already has the needed xcode version?

Yeah that's a great one. I feel like we can optimize this a lot in the future. For now, putting the responsibility back to the fastlane.ci server might be a good call for a few reasons:

1) If there are multiple workers, the `fastlane.ci` server can hand out the task to a different worker that might already have Xcode 9.4 installed
2) We tried to keep the workers as dumb as possible. This way if a worker dies, the master can make sure to continue working through the queue of jobs without missing something. Our assumptions is that a worker might die at any time, and we still want to be able to properly recover

Let me know if that makes sense

bogdanbrato: Makes sense - thanks.

PR #963, opened 2018-06-01
Verify bot has access to user repo on project creation

### LINKED ISSUE #881: Verify bot has access to user repo on project creation
From: #874 

We're using the user credentials to list repos and to get the lanes from the Fastfile. We should validate that the bot account has access to the selected repo.

armcburney:
Fixes #881.

### Changes

- Split responsibilities of `ConfigurationRepositoryService` into new `CollaboratorService` with a method to validate if the bot user is a collaborator on a given project ([`bot_user_collaborator_on_project?`](https://github.com/fastlane/ci/pull/963/files#diff-6a87f440f31f9198c999e7693c7d563bR56))

### TODO
- [x] Fix broken tests for the `ConfigurationRepositoryServiceSpec`
- [x] Add tests for the `CollaboratorService`
- [x] Do the verification on project creation


AndrewMcBurney:
@nakhbari, How were you envisioning validation for this? Were you envisioning a popup in the frontend notifying the user that the `bot_user` does not have access to the repository they're trying to create a project for?

Should I try and add the `bot_user` to the repository if they can't access it via first [`add_bot_user_as_collaborator!(repo_shortform:)`](https://github.com/fastlane/ci/pull/963/files#diff-6a87f440f31f9198c999e7693c7d563bR30), and then notify the user if that was unsuccessful in the frontend?

nakhbari:
I think there are two ways to do this.
1) we can show an intersection of repos that the user has access to and the bot has access to
2) we can show all the repos the user has access to and error out if the bot doesn't
I think showing an error message on the repo dropdown or a little red banner might be the way to go.

As for inviting the bot user to the repo, i'm wondering if we'll need an extra scope on the API token to do this.

Also, we just thought of moving to OAuth instead of API tokens. So having the `bot_user_collaborator_on_project?` method might make more sense in the GitHubService, which will eventually be stateless. Since we're removing the provisioning profiles in favor of passing along the GitHub OAuth token every time.

PR #973, opened 2018-06-07
Fix startup issues

taquitos:
- dot key service now has `keys` method that returns a new `DotKeysVariables` object with each call
- some variables were mis-named.

### DIFF
--- a/app/features/build_runner/build_runner.rb
+++ b/app/features/build_runner/build_runner.rb
@@ -201,7 +201,13 @@ module FastlaneCI
       @environment_variables_set = []
 
       # Set the CI specific Environment variables first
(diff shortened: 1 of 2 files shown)


app/features/build_runner/build_runner.rb
```diff
@@ -201,7 +201,13 @@ def setup_build_specific_environment_variables
       @environment_variables_set = []
 
       # Set the CI specific Environment variables first
-      build_url = File.join(Services.dot_keys_variable_service.ci_base_url, "projects", project.id, "builds", build.number)
+      build_url = File.join(
+        Services.dot_keys_variable_service.keys.ci_base_url,
```
KrauseFx: Whops, sorry, my bad 😢

Issue #985, opened 2018-06-12
Step 5 fails (bundle install from Using fastlane.ci)


androuis:
I got at step 5 from Using fastlane.ci: bundle install and it fails with this error:
Revision f71ec2e7a1eaf43c15312d611dd1cc4de530f823 does not exist in the
repository https://github.com/krausefx/google-cloud-ruby. Maybe you misspelled
it?

KrauseFx:
Ahhh, nice catch, thanks, working on a fix

KrauseFx:
Just merged https://github.com/fastlane/ci/pull/986, could you give it another try?

androuis:
Yes. It works. Thanks :)

KrauseFx:
Great to hear

Issue #984, opened 2018-06-12
KrauseFx:patch-1 branch for google-cloud-ruby deleted


kerry:
In Gemfile on the master, there is a dependency

Access to Google Cloud Platform Storage API.
`gem "google-cloud-storage", git: "https://github.com/krausefx/google-cloud-ruby", branch: "patch-1"`
TODO: Should be ">= 1.12.0", "< 2.0" once https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2099 is merged

But the patch-1 branch has already been deleted so this dependency fails to resolve.
Also, I saw that https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2099 will not merge and https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2129 is still open

Any resolution for this?

KrauseFx:
Ahhh, nice catch, thanks, working on a fix

KrauseFx:
Just merged https://github.com/fastlane/ci/pull/986, could you give it another try?

PR #986, opened 2018-06-12
Update dependencies

KrauseFx:
Fixes https://github.com/fastlane/ci/issues/985
Fixes https://github.com/fastlane/ci/issues/984


snatchev:
LGTM. Interesting to note we now pull in `activesupport`. Since this seems required, I don't think we should feel bad about using a lot of the utilities it provides. I know we try to keep dependencies down, but if we have it anyway, it makes sense to use.

PR #987, opened 2018-06-12
Worker agent updates

snatchev:
This PR attempts to break up #951 into 2 parts.
This part of the work is an update the `agent/` directory with the following high-level changes

- Get rid of the Rakefile in the worker and replace with a Ruby class (invocation.rb)
- Having a plan around Artifact storage. Currently this is implemented by sending back the artifact as part of the invocation response. (https://github.com/fastlane/ci/pull/987/files#diff-eac51902c47ef07f98b30b75b1f64a11R96)
- Passing exceptions back to the client making the request (https://github.com/fastlane/ci/pull/987/files#diff-eac51902c47ef07f98b30b75b1f64a11R96)

The agent will now use an `Invocation` to kick off a fastlane job. The invocation uses a state machine to transition through the different phases of a job.

The Invocation sends `InvocationResponse` back to the GRPC client which may contain:
* a state change
* a log line
* an Exception
* or a chunk of an Artifact.


agent/client.rb
```diff
@@ -3,16 +3,45 @@
...
       end
+
+      def request_run_fastlane(bin, *params, env: {})
+        command = Proto::Command.new(bin: bin, parameters: params, env: env)
+        @stub.run_fastlane(Proto::InvocationRequest.new(command: command))
+      end
     end
   end
 end
+
```
snatchev: Ignore this code from review. This will get removed with a subsequent PR.

agent/invocation.rb
```diff
@@ -0,0 +1,132 @@
...
+      error.stacktrace = exception.backtrace.join("\n")
+      error.error_description = exception.message
+
+      @yielder << FastlaneCI::Proto::InvocationResponse.new(error: error)
+    end
+
+    ## state machine transition guards
+
+    def has_required_xcode_version?
+      # TODO: bring in from build_runner
```
snatchev: This will get pulled in a subsequent PR

KrauseFx (approved):
Let's do this 🚀

agent/invocation/state_machine.rb
```diff
@@ -0,0 +1,105 @@
...
+  module StateMachine
+    include Logging
+    extend Forwardable
+
+    # forward methods to the state machine
+    def_delegators :state_machine, :state, :events, :states
+
+    def state_machine
+      @state_machine ||= MicroMachine.new("pending").tap do |fsm|
+        fsm.when(:run,     "pending"   => "running")
```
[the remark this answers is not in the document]
snatchev: 👍 good idea

agent/invocation.rb
```diff
@@ -0,0 +1,132 @@
...
+      Recipes.setup_repo(git_url)
+
+      unless has_required_xcode_version?
+        reject("Does not have required xcode version!. This is hardcode to be random.")
+        return
+      end
+
+      if Recipes.run_fastlane(@invocation_request.command.env.to_h)
+        finish
+      else
```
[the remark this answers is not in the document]
snatchev: You're right. It should return false in the case of a failed assertion during test. Will add this to `run_fastlane` which will eventually not shell out.

The rescue is in the `run_fastlane` method in `service.rb`

protos/agent.proto
```diff
@@ -43,3 +45,57 @@ message Command {
...
+      RUNNING = 1;
+      FINISHING = 2;  // job was successful, but we are still transmitting artifact.
+      BROKEN = 3;
+      SUCCEEDED = 4;
+      FAILED = 5;     // failed is when a test failed, error is when it could not complete the run.
+      REJECTED = 6;
+    }
+
+    State state = 1;
+    string description = 2; // provide more info for rejection.
```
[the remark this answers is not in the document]
snatchev: I like this suggestion quite a bit.

agent/invocation/state_machine.rb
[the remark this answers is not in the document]
snatchev: This doesn't look like it's going to be very easy, since the enum values are integers and the state machine accepts strings as fields. It may be possible to use some lookup methods to get the values you need, but I think it reduces readability substantially. I think it's fine to keep the hardcoded strings as long as we have a test to ensure these strings match the enums:

https://github.com/fastlane/ci/pull/987/files#diff-cfa28a249b00816d6bd72b264ead3a71R17

agent/invocation/state_machine.rb
```diff
@@ -0,0 +1,105 @@
...
+        # send update whenever we transition states.
+        fsm.on(:any) do |event, payload|
+          send_state(event, payload)
+        end
+      end
+    end
+
+    def send_state(event, payload)
+      logger.debug("Event `#{event}` causing state change to #{state}. #{payload}")
+      super if defined?(super)
```
AndrewMcBurney: This seems like a code smell calling `super` only if it's defined. What cases would `super` be defined and what cases would it not be defined?

snatchev: For `send_state`, super should always be defined, so I will get rid of this line. Thanks for calling that out.

For the other state transition methods (`run`, `fail`, etc.) , the module prepending `StateMachine` may or may not define any steps to take during the transition. Maybe we should just implement empty methods such as `succeed` and remove that `defined?` check. What do you think?

AndrewMcBurney: Ahh, okay - that makes sense. I have a mild preference towards the empty methods, but either way sounds good. If we decide on going with the current implementation, could we add a comment for future viewers (and future me because I'll probably forget 😂bahaha)

snatchev: @AndrewMcBurney updated with your suggestion of having empty methods

armcburney (approved):
Looks amazing! 💯

agent/invocation/state_machine.rb
AndrewMcBurney: Thank you very much @snatchev! Looks great!

bogdanbrato (approved):
👍

vicdes77:
I'm not doing this to my phone please reverse or stop this

PR #989, opened 2018-06-13
Fixed edge-case with server launch.

AndrewMcBurney:

### Changes

Fixed edge-case with server launch when `Services.ci_config_git_repo_path` exists, but the directory doesn't have the necessary configuration files.

### Rationale

When setting up `fastlane.ci` again, I ran into an issue where the configuration repository was not cloned properly. The directory for `Services.ci_config_git_repo_path` existed, but it was missing the necessary configuration files: `users.json`, `projects.json`, and `environment_variables.json`.

### DIFF
--- a/app/services/onboarding_service.rb
+++ b/app/services/onboarding_service.rb
@@ -8,6 +8,11 @@ module FastlaneCI
   class OnboardingService
     include FastlaneCI::Logging
 
+    # File names that should be present in configuration repository.
+    #
+    # @return [Array[String]]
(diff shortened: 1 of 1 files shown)


vicdes77:
I have never subscribed to this thread, someone is doing this to my phone And I want reversed, or stopped all off these things. Please. Someone is stealing all my info and getting g into my accounts, and pretending they are me. I mean I k ow who it is, just don't k ow how to stop it

PR #997, opened 2018-06-15
Remove Agent.spawn

bogdanbrato:
spawn was only a playground for experimenting with gRpc so no longer needed. 


KrauseFx:
Let's merge this 🚀

Issue #996, opened 2018-06-15
If missing `environment_variables.json` ci wants to onboard


taquitos:
I think this is a recent addition, but it breaks CI in that it won't startup due to:
```
[2018-06-15 10:44:38] DEBUG OnboardingService:  local configuration repo doesn't contain required configuration files
```
because in OnboardingService we have this:
`CONFIGURATION_FILES = ["users.json", "projects.json", "environment_variables.json"].freeze`
which is checked to make sure the repo is setup properly.
In this case, I think we should probably create defaults instead of going through onboarding.

KrauseFx:
```
def create_remote_configuration_files
      logger.debug("Creating remote configuration files `users.json` and `projects.json`")
      create_remote_json_file("users.json", json_string: serialized_users)
      create_remote_json_file("projects.json")
      create_remote_json_file("environment_variables.json")
    end
```

is created during onboarding. for now please create the file manually until we automate this

KrauseFx:
Looking at this, it does sound like it will cause some extra work to automatically migrate the user. New users will automatically have the file
 
The error message
```
local configuration repo doesn't contain required configuration files
```
is already very clear, I'll update it to explain how to easily create the file.

KrauseFx:
I'm gonna update it to call out the exact file name

KrauseFx:
It was my bad. Those files are generated on the fly anyway, no need to check if they exist. I'll update the PR, as well as added a comment mentioning this

AndrewMcBurney:
This was due to my oversight. I added `environment_variables.json` to the list , but the edge case described in #989  only exists when the `users.json` and `projects.json` files were not present in the `fastlane-ci-config` folder. I'll create a PR.

The problem is that if the `.keys` file exists with all the required variables set, but the `fastlane-ci-config` folder is empty (i.e., the git clone of the user's configuration repo was unsuccessful), it will not direct the user to onboarding, and the server will not be able to start due to an unrelated error.

AndrewMcBurney:
If you have all the required environment variables in `.keys`, have the `.fastlane/ci/fastlane-ci-config` folder locally, but don't have a `users.json` present in the folder, you will not be onboarded, and it will fail due to not being able to find the bot user account:

```
bundle exec rake dev DEBUG=1
bundle exec rackup --host 0.0.0.0 --port 8080 --env development
warning: parser/current is loading parser/ruby23, which recognizes
warning: 2.3.5-compliant syntax, but you are running 2.3.3.
[...]

PR #998, opened 2018-06-15
Update error message to indicate exact file missing

KrauseFx:
Fixes https://github.com/fastlane/ci/issues/996

### DIFF
--- a/app/services/onboarding_service.rb
+++ b/app/services/onboarding_service.rb
@@ -80,8 +80,9 @@ module FastlaneCI
       configuration_repo_contents = Dir[File.join(Services.ci_config_git_repo_path, "*")]
       configuration_files = CONFIGURATION_FILES.map { |f| File.join(Services.ci_config_git_repo_path, f) }
 
-      unless configuration_files.all? { |f| configuration_repo_contents.include?(f) }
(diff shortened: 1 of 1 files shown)


KrauseFx:
Actually no, gonna update this

KrauseFx:
Ready to be reviewed & merged from my side 👍

KrauseFx:
@AndrewMcBurney why? I thought that not having this file will allow the CI to run just fine, right? It's not actually required to get the server running and use fastlane.ci. From what I understand, the list should contain the files that are absolutely necessary to get the CI running. Once a user edits one ENV variable through the UI the file will automatically be created

taquitos:
@AndrewMcBurney I totally misunderstood how we used that list of files. I just synced with @KrauseFx and it makes sense to keep this file out of that list. 
I thought that this list was also used to generate the files, I didn't realize we generate this file only when needed.
So, in this case, we can leave it out of the `CONFIGURATION_FILES` list. Sorry for any confusion I caused here!

KrauseFx:
And to complete the context, here's where we create the file for the user: https://github.com/fastlane/ci/blob/be64b5a12ac22f73c4a6ccf855aca2f3fbb19e23/app/services/configuration_repository_service.rb#L127

PR #999, opened 2018-06-15
Add new json_error method for API layer

KrauseFx:
I really like the way the new AppStore Connect API works when it comes to error_keys, they do something like

```
InvalidParameter.KeyNotFound
```

where as the client can then decide how generic they want to parse the error message.

We could build up Ruby classes to represent those, but it might be overkill, and maybe we want just a markdown file maintained by us for now, listing all the available error codes, grouped by the first layer (e.g. InvalidParameter)

Long term, we probably do want those represented in Ruby classes/constants somehow and auto-generate the markdown file based on that.

### DIFF
--- a/app/features-json/api_controller.rb
+++ b/app/features-json/api_controller.rb
@@ -148,6 +148,15 @@ module FastlaneCI
         current_user != nil
       end
 
+      def json_error(error_message:, error_key:, error_code: 400)
+        status(error_code)
+
+        return json({
+          error: error_message,
+          error_code: error_key
+        })
+      end
+
       def current_user_provider_credential
         provider_credential = current_user.provider_credential
         halt(404, "Provider Credential not found.") unless provider_credential
--- a/app/features-json/api_controller.rb
+++ b/app/features-json/api_controller.rb
@@ -151,10 +151,10 @@ module FastlaneCI
       def json_error(error_message:, error_key:, error_code: 400)
         status(error_code)
 
-        return json({
+        halt(error_code, json({
           error: error_message,
           error_code: error_key
-        })
+        }))
       end
 
       def current_user_provider_credential
--- a/app/features-json/setting_json_controller.rb
+++ b/app/features-json/setting_json_controller.rb
@@ -16,7 +16,7 @@ module FastlaneCI
       setting = Services.setting_service.find_setting(setting_key: key.to_sym)
 
       if setting.nil?
-        return json_error(
+        json_error(
           error_message: "`#{params[:setting_key]}` not found.",
           error_key: "InvalidParameter.KeyNotFound"
         )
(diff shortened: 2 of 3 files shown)


taquitos:
Love the idea ❤️ ❤️ ❤️

KrauseFx:
For now, I'd recommend merging this. I'll then submit a separate PR migrating our existing APIs to use this new method. This way we can easily change the format to something else if that's better to use for the frontend

app/features-json/api_controller.rb
```diff
@@ -148,6 +148,15 @@ def user_logged_in?
         current_user != nil
       end
 
+      def json_error(error_message:, error_key:, error_code: 400)
+        status(error_code)
+
+        halt(error_code, json({
```
taquitos: neat, TIL `halt`

app/features-json/api_controller.rb
```diff
@@ -148,6 +148,15 @@ def user_logged_in?
         current_user != nil
       end
 
+      def json_error(error_message:, error_key:, error_code: 400)
```
taquitos: `400` code seem like a good default! ❤️

app/features-json/setting_json_controller.rb
```diff
@@ -36,7 +41,10 @@ class SettingJSONController < APIController
         Services.setting_service.reset_setting!(setting_key: params[:setting_key])
         return json({ status: :success })
       rescue SettingServiceKeyNotFoundError
-        halt(404, "`#{params[:setting_key]}` not found.")
+        json_error(
```
taquitos: changing from `404` to `400` is intentional, right?

KrauseFx:
Aaaand, we already abstracted it out, so we'll never have to use `halt` again 😂

app/features-json/setting_json_controller.rb
KrauseFx: Ohh, it's not actually, good find. Do you think it should be 404?

spec/features-json/setting_json_controller_spec.rb
```diff
@@ -66,5 +67,14 @@
 
       expect(FastlaneCI::Services.setting_service.find_setting(setting_key: metrics_key.to_sym).value).to eq(nil)
     end
+
+    it "returns an error if key can't be found" do
```
taquitos: I think we also test http status code (`400` in this case?) because client-side might have some status code logic.

KrauseFx: Yeah good point, will update 👍

spec/features-json/setting_json_controller_spec.rb
```diff
@@ -72,7 +72,7 @@
       non_existent_key = "non_existent_key"
       delete("/data/settings/#{non_existent_key}")
 
-      expect(last_response).to_not(be_ok)
+      expect(last_response.status).to eq(400)
```
taquitos: ❤️ ❤️

PR #1002, opened 2018-06-18
Update JSONSettingDataSource to return default values if no config is available

KrauseFx:
This fixes the tests on machines without a ci-config repo


KrauseFx:
yessss 💯

PR #1003, opened 2018-06-18
Remove `environment_variables.json` from list of required configuration files in `OnboardingService`.

### LINKED ISSUE #996: If missing `environment_variables.json` ci wants to onboard

AndrewMcBurney:

Fixes #996

### DIFF
--- a/app/services/onboarding_service.rb
+++ b/app/services/onboarding_service.rb
@@ -11,7 +11,7 @@ module FastlaneCI
     # File names that should be present in configuration repository.
     #
     # @return [Array[String]]
-    CONFIGURATION_FILES = ["users.json", "projects.json", "environment_variables.json"].freeze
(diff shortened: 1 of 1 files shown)


AndrewMcBurney:
After a discussion with @taquitos, we've decided to close this issue. This problem should only affect alpha users who had an outdated version of the `fastlane-ci-config` repo. The solution to fixing this problem is to create the file `environment_variables.json` in the `~/.fastlane/ci/fastlane-ci-config` folder.

PR #1004, opened 2018-06-18
Migrate all API endpoints to return error code and error description

KrauseFx:


app/features-json/api_controller.rb
```diff
@@ -107,13 +107,29 @@ def jwt
...
+        json_error!(
+          error_message: "The token does not have a valid issuer",
+          error_key: "Authentication.Token.InvalidIssuer",
+          error_code: 403
+        )
       rescue JWT::InvalidIatError
-        halt(403, { "Content-Type" => "text/plain" }, 'The token does not have a valid "issued at" time.')
+        json_error!(
+          error_message: "The token does not have a valid 'issued at' time",
+          error_key: "Authentication.Token.MissingIssuedTime",
```
taquitos: Since we use the string `Authentication.Token` a lot, it would be great to have some way of building these error keys that is less likely to be prone to typos.

docs/API.md
```diff
@@ -0,0 +1,60 @@
+# API
+
+This document covers the documentation for the API the Sinatra service provides.
+
+## Used status codes
+
+Status Code | Description
+------------|--------------
+200			| All good
+400			| Something went wrong
```
taquitos: for 400, something is wrong with the client's request
whereas, 5xx is something went wrong on the server.

KrauseFx:
That's great feedback, thanks, will do that first thing in the morning

app/features-json/api_controller.rb
```diff
@@ -148,9 +170,21 @@ def user_logged_in?
...
+        error_key = "Unknown" if error_key.to_s.length == 0
+
+        # Set the HTTP status for Sinatra
         status(error_code)
 
+        # Use `halt` to immediately return the error
+        # to the client
         halt(error_code, json({
           error: error_message,
           error_code: error_key
```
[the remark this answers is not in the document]
KrauseFx: Sure, updating now 👍

KrauseFx:
Yes, the long term plan is to have an enum that defines all the available error codes and auto generate the docs based on that. I feel like this could be a separate PR as it's not critical for a successful beta launch 👍

KrauseFx:
> I'd prefer some way to prevent typos when we reuse error code contexts, e.g.: using Authentication.Token.X we should consider using a const for Authentication.Token

@taquitos can you tell me a bit more about this? I'm not sure I'm following

spec/features-json/api_controller_spec.rb
```diff
@@ -125,17 +130,21 @@ class MyMixedAuthApiController < FastlaneCI::APIController
...
-        expect do
-          app.helpers.authenticate!(via: :jwt)
-        end.to throw_symbol(:halt)
-
-        app.helpers.request = Sinatra::Request.new(Rack::MockRequest.env_for("/", {}))
-        expect do
-          app.helpers.authenticate!(via: :jwt)
-        end.to throw_symbol(:halt)
-      end
+      # TODO: These tests need to be fixed, as they broke with https://github.com/fastlane/ci/pull/1004
```
snatchev: 🙋‍♂️  I can probably help with this

PR #1005, opened 2018-06-19
update for sinatra security vuln

taquitos:

Updates for security, Sinatra 2.0.3 finally plays nicely with the other gems 🎉

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,7 +6,7 @@ GIT
 
 GIT
   remote: https://github.com/fastlane/fastfile-parser
-  revision: a9707d87eb6bfdc74c9fc3125f7eec917c8c1e49
+  revision: 231cb18ea44974cffe9502fbfb7b42ebf7d25bc5
   specs:
     fastfile-parser (0.0.0)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -79,6 +79,11 @@ GEM
   specs:
     CFPropertyList (3.0.0)
     abstract_type (0.0.7)
+    activesupport (5.2.0)
```
KrauseFx: As pointed out by @snatchev, this is not perfect, but not much we can do for now. I'll investigate on why this is added, and submit an issue if there isn't one already

KrauseFx (approved):
💯

KrauseFx:
Alright it was added with https://github.com/sinatra/sinatra/pull/1442, but no issue where people discuss it, will create one now

KrauseFx:
Created an issue https://github.com/sinatra/sinatra/issues/1451

KrauseFx:
Either way, let's merge and go ahead

taquitos:
Great, thanks for creating the issue.

Issue #1010, opened 2018-06-19
JSON GET to get the User's email from API token


nakhbari:
Parent: #784
Need an API endpoint (probably in the `repo_json_controller`) to get the User email of the given API Token. This also verifies a token is good. This can be used in onboarding and re-used in the account creation.

PR #1016, opened 2018-06-19
Add new API endpoint to fetch user metadata

KrauseFx:
Fixes https://github.com/fastlane/ci/issues/1010

### DIFF
--- a/app/features-json/build_json_controller.rb
+++ b/app/features-json/build_json_controller.rb
@@ -10,15 +10,7 @@ module FastlaneCI
     use(FastlaneCI::BuildWebsocketBackend)
 
     get "#{HOME}/:build_number" do |project_id, build_number|
-      build = current_project.builds.find { |b| b.number == build_number.to_i }
-      if build.nil?
-        json_error!(
-          error_message: "Can't find build with ID #{build_number} for project #{project_id}",
-          error_key: "Build.Missing",
-          error_code: 404
-        )
-      end
-      build_view_model = BuildViewModel.new(build: build)
+      build_view_model = BuildViewModel.new(build: current_build)
 
       json(build_view_model)
     end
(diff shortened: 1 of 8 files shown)


KrauseFx:
This is ready for review, I started writing tests, but haven't finished yet - running into meetings now for the next few hours.

app/services/code_hosting/git_hub_service.rb
```diff
@@ -69,6 +69,11 @@ def username
       client.login
     end
 
+    # Primary email address of the current GitHub user
+    def email
+      client.emails.find { |e| e[:primary] == true }[:email]
```
taquitos: we use this shorter form in other places, any reason why we shouldn't use it here?: `client.emails.find(&:primary).email`

app/features-json/setup_json_controller.rb
```diff
@@ -8,5 +8,19 @@ class SetupJSONController < APIController
     get "#{HOME}/configured", authenticate: false do
       return json(Services.onboarding_service.correct_setup?)
     end
+
+    get "#{HOME}/userDetails" do
```
taquitos: I think for the most part we've been using `snake_case` instead of `camelCase` for our routes. Since `snake_case` seems to be the dominant format, I suggest we stay consistent with it.

app/features-json/setup_json_controller.rb
```diff
@@ -8,5 +8,19 @@ class SetupJSONController < APIController
...
+      code_hosting_service = current_user_config_service.code_hosting_service(
+        provider_credential: Services.provider_credential
```
taquitos: `Services.provider_credential` is the bot provider credential. I think you want the current user instead, right?

app/services/code_hosting/git_hub_service.rb
KrauseFx: So fancy, so much magic, will update 👍 I just never think about it

app/features-json/setup_json_controller.rb
KrauseFx: Ah yeah, totally

app/features-json/setup_json_controller.rb
KrauseFx: Ah yeah, seems like I had the same user for both the bot and the user. Do you know how I can access the provider credentials of the user in a JSON controller?

KrauseFx:
I'm having a hard time stubbing the `current_user_config_service` method as part of the `SetupJSONController`, does anyone know more on how to best do this? 
I tried

```ruby
expect(app).to receive(:current_user_config_service).and_return("yolo")
```
as well as 
```ruby
allow_any_instance_of(described_class).to receive(:current_user_config_service).and_return("yolo")
```
however it doesn't return `"yolo"`

taquitos:
No idea on that @KrauseFx, I've had similar issues in the past, but don't know what the resolution was. I think I just never trust `allow_any_instance_of`

KrauseFx:
For now, we can merge this without a test to unblock Nima, and we can ask Stefan once he's back as he might know more about how to stub those Sinatra controllers

app/features-json/setup_json_controller.rb
```diff
@@ -8,5 +8,19 @@ class SetupJSONController < APIController
     get "#{HOME}/configured", authenticate: false do
       return json(Services.onboarding_service.correct_setup?)
     end
+
+    get "#{HOME}/user_details" do
+      code_hosting_service = current_user_config_service.code_hosting_service(
+        provider_credential: Services.provider_credential
```
taquitos: `Services.provider_credential` is the wrong credential still, gotta get the current user.

taquitos: The API here is weird. if a `code_hosting_service` is coming back with credentials of the current user, even though you passed in a `Services.provider_credential`, that's broken. I think what could be happening is that your current user is the same user as your bot.

app/features-json/setup_json_controller.rb
```diff
@@ -8,5 +8,19 @@ class SetupJSONController < APIController
     get "#{HOME}/configured", authenticate: false do
       return json(Services.onboarding_service.correct_setup?)
     end
+
+    get "#{HOME}/user_details" do
```
nakhbari: this is a repo specific task shouldn't it live in `repo_json_controller`?

app/features-json/setup_json_controller.rb
```diff
@@ -8,5 +8,19 @@ class SetupJSONController < APIController
...
+      code_hosting_service = current_user_config_service.code_hosting_service(
```
nakhbari: this API should be able to be called without any configuration being done yet to the server. I haven't look deeply into this, but does this method require initial configuration?

Issue #934, opened 2018-05-25
Accessing /login in a logged in state should redirect to the dashboard


KrauseFx:
Steps to reproduce

1. Login
2. Access `http://localhost:9292/login`

Expected result: Redirect to dashboard
Actual result: Login is shown

kevinsamoei:
How can one reproduce this, I don't think there is a way to register as a new user via the UI or is there an email and password that can be used for testing?

KrauseFx:
Hey @kevinsamoei, see our comments on https://github.com/fastlane/ci/issues/924#issuecomment-393209605, right now the onboarding isn't implemented on the new frontend yet :)

Issue #950, opened 2018-05-31
Cleanup build data after a certain limit


nakhbari:
To save disk space we should cleanup the repo clone used to run the build. @taquitos suggested a default limit of 20 builds to be kept in the history, any older build would be cleaned up in a separate thread/task since it could take time.

nakhbari:
Maybe this could be apart of the new hotness `settings.json` file

KrauseFx:
> To save disk space we should cleanup the repo clone used to run the build

What do you mean by that? Sorry I don't understand.

Like limit the number of `builds/[number].json` files?

nakhbari:
Each build has a copy of the entire repo, including any build data that was created in that workspace during the fastlane command. As seen when doing something like `npm install` this can be quite costly when it comes to disk space. Those lingering files are what should be deleted. The actual build data we display (outcome, logs, artifacts) should remain intact.

KrauseFx:
Ahh got it. Yeah I believe the workspaces can safely be cleaned after the build is finished and its artifacts were uploaded, right? This actually ties in with the new master/worker code

PR #1022, opened 2018-06-20
Redirect /login to /dashboard if logged in

powerivq:
- [X] I have run `rspec` and corrected all errors
- [X] I have run `rubocop` and corrected all errors
- [X] I have tested this change locally and tried to launch the server as well as access a project, and with that at least one build
- [X] If there is an existing issue, make sure to add `Fixes ...` as part of the PR body to reference the issue you're solving

Fixes https://github.com/fastlane/ci/issues/934

### DIFF
--- a/web/app/login/login.component.ts
+++ b/web/app/login/login.component.ts
@@ -8,7 +8,7 @@ import {AuthService} from '../services/auth.service';
   templateUrl: './login.component.html',
   styleUrls: ['./login.component.scss']
 })
-export class LoginComponent {
+export class LoginComponent implements OnInit {
   email: string;
   password: string;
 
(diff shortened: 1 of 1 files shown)


powerivq:
@nakhbari oops sry. any chance we run this on ci as well?

nakhbari:
@powerivq we tried, but because we currently do not cleanup the files created during a build the disk space ended up being completely used up. Maybe we can look into it again when we get that feature fixed. https://github.com/fastlane/ci/issues/950

powerivq:
@nakhbari Oh gotcha. Thanks for the fix btw.
@nakhbari @taquitos Do you think it's a good idea to run tests on circle as well? I feel that extra coverage won't hurt at the very least.

PR #1025, opened 2018-06-21
Tests for when user is already logged in

nakhbari:
```
- [ ] I have run `rubocop` and corrected all errors
- [ ] I have tested this change locally and tried to launch the server as well as access a project, and with that at least one build
- [ ] If there is an existing issue, make sure to add `Fixes ...` as part of the PR body to reference the issue you're solving
```


powerivq:
Thanks for fixing this! @nakhbari

powerivq:
Thanks for fixing this! @nakhbari

PR #1045, opened 2018-06-27
Get Build logs when build is complete

nakhbari:
Issue: #709 

Probably want to make this a part of the websocket in the future. This PR should give us basic functionality to switch over to the web app in all situations 😄 

#### Screenshot
image

```
```

### DIFF
--- a/app/features-json/build_json_controller.rb
+++ b/app/features-json/build_json_controller.rb
@@ -10,15 +10,7 @@ module FastlaneCI
     use(FastlaneCI::BuildWebsocketBackend)
 
     get "#{HOME}/:build_number" do |project_id, build_number|
-      build = current_project.builds.find { |b| b.number == build_number.to_i }
-      if build.nil?
-        json_error!(
-          error_message: "Can't find build with ID #{build_number} for project #{project_id}",
-          error_key: "Build.Missing",
-          error_code: 404
-        )
-      end
-      build_view_model = BuildViewModel.new(build: build)
+      build_view_model = BuildViewModel.new(build: current_build)
 
       json(build_view_model)
     end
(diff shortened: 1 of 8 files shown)


app/features-json/build_json_controller.rb
```diff
@@ -92,6 +84,42 @@ class BuildJSONController < APIController
...
+      # if the server was restarted, we're gonna end here in this code block
+      build_log_artifact = current_build.artifacts.find do |current_artifact|
+        # We can improve the detection in the future, to actually mark an artifact as "default output"
+        current_artifact.type.include?("log") && current_artifact.reference.end_with?("fastlane.log")
+      end
+
+      if build_log_artifact
+        artifact_file_content = File.read(build_log_artifact.provider.retrieve!(artifact: build_log_artifact))
+      else
+        raise "Couldn't load previous output for build #{build_number}"
```
nakhbari: should probably be one of our new errors. Will have to change that.

powerivq:
Just a comment: can we force logs to be displayed using monospace font? That would make it more readable because many command line apps assume that.

nakhbari:
@powerivq Yeah i'm sure we'll use something like that. I think there is a discussion to be had on how we styling the logs in general.

web/app/build/build.component.scss
```diff
@@ -11,14 +11,16 @@ $TOOLBAR_HEIGHT:80px;
         .fci-build-logs {
             flex: 1;
             padding: 8px;
-            color: #859900;
+            color: #93a1a1;
             background-color: #073642;
+            font-family: 'Google Sans Monospace', sans-serif;
```
powerivq: Thanks! Let's make it 
```
'Google Sans Monospace', monospace;
```
for now. We likely want a default monospace font at all times.

app/features-json/build_json_controller.rb
```diff
@@ -104,5 +137,10 @@ def current_project
 
       return current_project
     end
+
+    # convert .log files that include the color information as ANSI code to plain text
+    def convert_ansi_to_plain_text(data)
```
KrauseFx: Yeah that's good for now, we'll address that code again later to improve it

app/features-json/build_json_controller.rb
```diff
@@ -92,6 +84,47 @@ class BuildJSONController < APIController
       json(build_summary_view_model)
     end
 
+    get "#{HOME}/:build_number/logs" do |project_id, build_number|
+      # `current_build_runner` is only defined if the build was just run a while back
+      # if the server was restarted, we're gonna end here in this code block
+      build_log_artifact = current_build.artifacts.find do |current_artifact|
+        # We can improve the detection in the future, to actually mark an artifact as "default output"
```
KrauseFx: Yeah 👍

app/features-json/build_json_controller.rb
```diff
@@ -92,6 +84,47 @@ class BuildJSONController < APIController
...
+          error_message: "Logs file missing for build #{build_number}",
+          error_key: "Build.LogsMissing",
+          error_code: 404
+        )
+      end
+
+      plain_artifact_file_content = convert_ansi_to_plain_text(artifact_file_content)
+      log_array = plain_artifact_file_content.split("\n").collect do |log_line|
+        {
+          message: log_line
```
KrauseFx: This notation is here because in the future we'll also transfer the "metadata" of each row right? (e.g. warning, error). And we just don't have this information for now as we just store the plaintext log output for the time being

app/features-json/build_json_controller.rb
```diff
@@ -92,6 +84,47 @@ class BuildJSONController < APIController
...
+    get "#{HOME}/:build_number/logs" do |project_id, build_number|
+      # `current_build_runner` is only defined if the build was just run a while back
+      # if the server was restarted, we're gonna end here in this code block
+      build_log_artifact = current_build.artifacts.find do |current_artifact|
+        # We can improve the detection in the future, to actually mark an artifact as "default output"
+        current_artifact.type.include?("log") && current_artifact.reference.end_with?("fastlane.log")
+      end
+
+      if build_log_artifact
+        artifact_file_content = File.read(build_log_artifact.provider.retrieve!(artifact: build_log_artifact))
```
KrauseFx: This will work for local file storage only. As with `gcp` artifact storage provider we'll return a URL that we'll need to fetch. This doesn't have to be addressed in this PR, however let's either add a `# TODO` here or create a GH issue to follow up 👍 I'm sure we have other work to do for the GCP integration

app/features-json/build_json_controller.rb
nakhbari: Yeah, we'll have to talk about how we want to structure logs are some point. I think it would be OK to output plain text until the worker stuff is done. Don't want to do the log work twice.

app/features-json/build_json_controller.rb
nakhbari: added TODO

PR #1053, opened 2018-07-02
Add encryption key and form to onboarding

nakhbari:
Issue: #784 

#### Screenshot
image

```
```

### DIFF
--- a/web/global/input.scss
+++ b/web/global/input.scss
@@ -1,5 +1,12 @@
 .fci-input-container {
     padding-bottom: 16px;
+    label {
+        font-family: 'Roboto';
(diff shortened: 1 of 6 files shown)


web/global/input.scss
```diff
@@ -1,5 +1,12 @@
 .fci-input-container {
     padding-bottom: 16px;
+    label {
+        font-family: 'Roboto';
```
KrauseFx: I feel like we had that quite a lot in the recent PRs, should we unify? It's `scss`, so should be easy

KrauseFx: Yeah sure thing 👍

PR #1057, opened 2018-07-03
Add fastlane bot configuration to onboarding

nakhbari:
Issue: #784 

#### Screenshots
image

image

```
```


AndrewMcBurney:
I think we need the `user:email` scope on this API token (and the clone user API token) in addition to all `repo` permissions. Was something changed so that we don't need that scope anymore? I think there was a case where the primary email of a user on GitHub was `private`, yet we still need to find the email to match it with a fastlane.ci account.

nakhbari:
@AndrewMcBurney yes we still need this scope. I will have to update the screenshot at some point. I added a bullet in the top-level issue #784

Issue #1013, opened 2018-06-19
Endpoint to create an Account given an API Token


nakhbari:
Parent: #1011 

POST endpoint to create a user. Input should be an API Token and password. The email should be pulled from the API token and be used as the user ID.

This endpoint should login the new user by returning a JWT token.

```json
{
  "api_token": "234",
  "password": "456"
}
```

nakhbari:
@KrauseFx Could you help me out with this?

KrauseFx:
Yes, I can take that on today 👍

KrauseFx:
> This endpoint should login the new user by returning a JWT token.

I don't think that's a good idea. We probably will have some kind of user management at some point, which ideally uses the same API endpoint I guess? I believe we should let the frontend do 2 requests:

1. Create the user
1. Login

PR #1059, opened 2018-07-06
Add new API endpoint to register user

KrauseFx:
Fixes https://github.com/fastlane/ci/issues/1013

### DIFF
--- a/app/features-json/user_json_controller.rb
+++ b/app/features-json/user_json_controller.rb
@@ -1,0 +1,58 @@
+require_relative "../services/user_service"
+require_relative "api_controller"
+require_relative "json_params"
+
+module FastlaneCI
+  # Controller responsible for handling users
+  class UserJSONController < APIController
+    disable(:authentication)
+
+    HOME = "/api/user"
+
+    post HOME.to_s do
+      email = params[:email]
+
+      if email.nil?
+        # User didn't provide a custom email address
+        # fetch it based on the API token instead
+        github_client = Octokit::Client.new(access_token: params[:github_token])
+
+        begin
+          # Note: This fails if the user.email scope is missing from token
+          email = github_client.emails.find(&:primary).email
+        rescue Octokit::NotFound
+          json_error!(
+            error_message: "Provided API token needs user email scope",
+            error_key: "User.Token.MissingEmailScope",
+            error_code: 400
+          )
+        rescue Octokit::Unauthorized
+          json_error!(
+            error_message: "Provided API token is invalid",
+            error_key: "User.Token.Invalid",
+            error_code: 403
+          )
+        end
+      end
+
+      user = Services.user_service.create_user!(
+        email: email,
+        password: params[:password]
+      )
+
+      if user
+        Services.user_service.create_provider_credential!(
+          user_id: user.id,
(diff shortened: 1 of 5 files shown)


app/features-json/user_json_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+
+module FastlaneCI
+  # Controller responsible for handling users
+  class UserJSONController < APIController
+    disable(:authentication)
+
+    HOME = "/api/user"
+
+    post HOME.to_s do
+      email = params[:email]
```
nakhbari: I was wondering about this. Should we allow email as a param? Someone could contact our endpoint directly and set a random email. But pulling it from the token makes sure that it is always 1:1 with github or other providers.

app/features-json/user_json_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      if user
+        Services.user_service.create_provider_credential!(
+          user_id: user.id,
+          email: user.email,
+          api_token: params[:github_token]
+        )
+        return json({ status: :success })
+      else
+        json_error!(
```
nakhbari: could we single out the error where the user already exists? It would provide better messaging in the client.

app/features-json/user_json_controller.rb
KrauseFx: Ah, I assume it should be possible, but you're right, let's remove it for now until we need it

app/features-json/user_json_controller.rb
KrauseFx: Yeah, in this case we don't get any error information - as the `create_user!` method doesn't expose the error type as of now. There are probably gonna be other error types in the future. I'd be in favor of keeping the generic message for now, until we spend more time refining what kind of error messages we want, and how we want to handle them (e.g. missing config file, no internet connection, etc.) Right now we catch some of them, but most exceptions are just surfaced all the way up, and therefore not returning an actual JSON object. 

We can't just expose the stack trace and exception AFAIK as it potentially contains sensitive information.

KrauseFx:
Alright, waiting for @snatchev to confirm 👍

spec/features-json/user_json_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  def app
+    described_class
+  end
+  let(:json) { JSON.parse(last_response.body) }
+
+  before do
+    allow(FastlaneCI.dot_keys).to receive(:encryption_key).and_return("test")
+  end
+
+  describe "/api/user" do
```
snatchev: How does this test work, given that we disable network requests via Octokit?

KrauseFx: Investigating this now

KrauseFx: Ha, it actually doesn't work on my machine, fixing now

KrauseFx: Ah because I removed the `email` parameter last minute, we manually passed one before

KrauseFx:
Alright, pushed some changes, should be up in a few minutes 👍

KrauseFx:
cc @snatchev

Issue #1060, opened 2018-07-06
Project Settings Dialog


nakhbari:
Parent: #682 

- [x] Create a Master Detail Card Component
- [x] Create Settings Dialog
- [x] Add General Settings placeholder
- [ ] Create Environment Key Editor Component
- [ ] Pipe full Repo name and full lane name to the frontend
- [ ] Endpoint to update the project
- [ ] Endpoint to change environment Keys

PR #1062, opened 2018-07-06
Create Settings Dialog component

nakhbari:
Issue: #1060 

This is the component implementation. It will be added to the Project details page in a separate PR. We need to change the project details data to include the full repo name and the full lane name (including the platform) for us to get the lanes. We also need an endpoint to update the project. So all the inputs are disabled for now.

Sneak Peek:
image

```
```

### DIFF
--- a/web/app/project/settings-dialog/settings-dialog.component.scss
+++ b/web/app/project/settings-dialog/settings-dialog.component.scss
@@ -1,0 +1,21 @@
+@import '../../common/components/master-detail-card/master-detail-card-shared-styles';
+.fci-detail-container {
+    h3 {
+        font-size: 20px;
+        color: rgba(0, 0, 0, 0.87);
+        font-weight: 500;
+        margin: 0 0 24px 0;
+    }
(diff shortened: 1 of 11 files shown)


web/app/project/settings-dialog/settings-dialog.component.scss
```diff
@@ -0,0 +1,21 @@
...
+    h3 {
+        font-size: 20px;
+        color: rgba(0, 0, 0, 0.87);
+        font-weight: 500;
+        margin: 0 0 24px 0;
+    }
+    form {
+        display: flex;
+        flex-direction: column;
+        mat-form-field {
```
KrauseFx: Why is this weirdly highlighted here? GH bug or missing a symbol?

nakhbari: it's an unknown element tag. An known element tag would be `form`. But the css rules still work since i believe angular uses web components to represent their components in the DOM.

Issue #1012, opened 2018-06-19
JSON POST endpoint to create config repo


nakhbari:
Parent: #784 

Endpoint should take in an org and be able to create a config repo unless one already exists.

nakhbari:
Expecting the post body to look like this

```json
{
  "encryption_key": "key",
  "robot_account": {
    "token": "123",
    "password": "234"
  },
  "config_repo": "org/ci-repo"
}
```

There are a lot of errors that can happen here. 

@KrauseFx do you have cycles to work on this?

nakhbari:
Should break up the requests or just make a super-request at the end of the form?

Offline discussion between @anshulagl and I:
>Anshul Agrawal:
A super request at the end sounds a better approach to me, it really depends on if we want to checkpoint user's onboarding if she wants to return back and complete it at a later point of time, if that's the desired experience, independent requests might make more sense.
>
>Nima Akhbari:
I think incremental request would also give more feedback on incorrect params or other errors at the step that the user is on, instead of at the end of the flow. But this would also mean latency at each step as the user waits for the request to go through
>
>Anshul Agrawal:
That makes total sense if we need to validate the info at each step, showing all errors at the end wouldn't make much sense. Ideally the server latency would be <500ms. If the user is clicking the "Next" button, wouldn;t feel that big a difference.
>
>Nima Akhbari:
server latency is one thing, but we'll also have to do some work on the backend which could take time. For example, getting the email from Github for a specific access token.

KrauseFx:
Taking this on now

KrauseFx:
I'm gonna rename `robot` to `bot`
```json
{
  "encryption_key": "key",
  "bot_account": {
    "token": "123",
    "password": "234"
  },
  "config_repo": "org/ci-repo"
}
```

nakhbari:
The requirements for this endpoint are now changing due to the usage of GitHub OAuth Apps. I will update this issue with the draft proposal

nakhbari:
@KrauseFx With the proposal approved and some of the UI work being done. I think we can safely say that this is the direction to head with this endpoint. Of course now we have a couple more endpoints to add with the new OAuth design. Those end points are outlined in the parent issue: #784. **Are you still free to take some of these on?**

PR #1066, opened 2018-07-09
Add onboarding API endpoint

KrauseFx:
- [x] We need to protect the API endpoint

Fixes https://github.com/fastlane/ci/issues/1012


app/features-json/setup_json_controller.rb
```diff
@@ -8,5 +8,109 @@ class SetupJSONController < APIController
...
       return json(Services.onboarding_service.correct_setup?)
     end
+
+    # TODO: we need to protect this API endpoint
+    # one the onboarding user flow, we don't have any accounts yet
+    # however, we need to prohibit sending a POST request to this API endpoint
+    # after CI was already set up
+    post HOME.to_s, authenticate: false do
+      # we set this as default so that we don't need duplicate code for error detection
+      params["bot_account"] ||= {}
```
nakhbari: nit: maybe a new var so later on we don't have to do `params[..]`?

app/features-json/setup_json_controller.rb
```diff
@@ -8,5 +8,109 @@ class SetupJSONController < APIController
...
+
+      encryption_key = params["encryption_key"]
+      bot_token = params["bot_account"]["token"]
+      bot_password = params["bot_account"]["password"] # TODO: What do we need this for
+      config_repo = params["config_repo"]
+      initial_onboarding_token = params["initial_onboarding_token"]
```
nakhbari: I never understood what the initial onboarding token was. Why don't we just use the bot token?

app/features-json/setup_json_controller.rb
```diff
@@ -8,5 +8,109 @@ class SetupJSONController < APIController
...
+      validate_api_token_correct!(initial_onboarding_token)
+      dot_keys_values[:initial_onboarding_user_api_token] = initial_onboarding_token
+
+      # Step: Validate encryption key
+      validate_encryption_key!(encryption_key)
+      dot_keys_values[:encryption_key] = encryption_key
+
+      # Step: Validate repo URL
+      # TODO: What do we want to validate here?
+      dot_keys_values[:repo_url] = config_repo
```
nakhbari: should we validate that the bot has access to the repo?

app/features-json/setup_json_controller.rb
```diff
@@ -8,5 +8,109 @@ class SetupJSONController < APIController
...
+          error_message: error_message,
+          error_key: "Onboarding.Token.MissingScope"
+        )
+      end
+    end
+
+    def validate_encryption_key!(encryption_key)
+      if encryption_key.to_s.length < 5
+        json_error!(
+          error_message: "Encryption key not long enough, it must be at least 5 characters",
```
nakhbari: oh i didn't know this was a requirement, we should add it to the UI to make sure it's at least 5 chars

app/features-json/setup_json_controller.rb
KrauseFx: I think you're right, we could do that 👍 Will update the existing code base also to reflect this. Anything we're missing @taquitos

app/features-json/setup_json_controller.rb
KrauseFx: I only default to `{}` to easily access nested hashes later in the code. We can also update the code base to use `.fetch`

app/features-json/setup_json_controller.rb
```diff
@@ -8,5 +8,150 @@ class SetupJSONController < APIController
...
+      bot_password = params["bot_account"]["password"] # TODO: What do we need this for
+      config_repo = params["config_repo"].to_s
+      initial_onboarding_token = params["initial_onboarding_token"]
+
+      missing_parameters = {
+        encryption_key: encryption_key,
+        bot_token: bot_token,
+        bot_password: bot_password,
+        config_repo: config_repo,
+        initial_onboarding_token: initial_onboarding_token
```
nakhbari: so can we remove the requirement of this param?

>nakhbari:
I never understood what the initial onboarding token was. Why don't we just use the bot token?
>KrauseFx:
I think you're right, we could do that 👍 Will update the existing code base also to reflect this. Anything we're missing @taquitos

KrauseFx: So seems like for now we decided to leave it as is, and do the refactor in a separate PR, right?

app/features-json/setup_json_controller.rb
```diff
@@ -8,5 +8,150 @@ class SetupJSONController < APIController
...
+    # one the onboarding user flow, we don't have any accounts yet
+    # however, we need to prohibit sending a POST request to this API endpoint
+    # after CI was already set up
+    post HOME.to_s, authenticate: false do
+      # we set this as default so that we don't need duplicate code for error detection
+      params["bot_account"] ||= {}
+
+      encryption_key = params["encryption_key"]
+      bot_token = params["bot_account"]["token"]
+      bot_password = params["bot_account"]["password"] # TODO: What do we need this for
```
KrauseFx: @nakhbari what is this value being used for?

nakhbari:
I marked this as blocked since we're ironing out the details of Onboarding with the new Github OAuth App

KrauseFx:
I think it's safe to say that this change will not get merged, as we have a new approach for onboarding + auth. I'll close this PR for now 👍

PR #1071, opened 2018-07-12
RemoteRunner refactor with Notifications, add specs.

### LINKED ISSUE #457: Support for distributed builds

snatchev:
fixes #457
This PR is a refactor to the RemoteRunner (#951) for increased testability, and decomposition.
The biggest change is that WebSocket clients are now updated of build Agent events by the remote runner via Notifications.

`FastlaneCI::Notifications` are pulled in from ActiveSupport, with a minor change that allows notifying a subscriber directly (instead of having to go via topic) https://github.com/fastlane/ci/pull/1071/files#diff-4ef29c8b1610800f883c5eb428e5634aR83.
This is used when a client subscribers to a topic and all previous messages are delivered immediately.


app/features-json/build_json_controller.rb
```diff
@@ -85,7 +87,7 @@ class BuildJSONController < APIController
       # if the server was restarted, we're gonna end here in this code block
       build_log_artifact = current_build.artifacts.find do |current_artifact|
         # We can improve the detection in the future, to actually mark an artifact as "default output"
-        current_artifact.type.include?("log") && current_artifact.reference.end_with?("fastlane.log")
+        current_artifact.type.include?("log") && current_artifact.reference.end_with?("runner.log")
```
snatchev: this was changed to `runner.log` because I wanted to make sure we don't confuse this log with the `fastlane.log` which is part of the archive bundle. That log does not include things like the git checkout, and other setup. Also it is raw ANSI text, and not rows of JSON records, which is what `runner.log` is.

app/features-json/view_models/build_view_model.rb
```diff
@@ -57,8 +57,6 @@ class BuildViewModel
     attr_reader :artifacts
 
     def initialize(build:)
-      raise "Incorrect object type. Expected Build, got #{build.class}" unless build.kind_of?(Build)
```
snatchev: Removed because this prevents us from passing in test doubles.

app/features-json/build_json_controller.rb
```diff
@@ -1,14 +1,16 @@
...
 
+require "faye/websocket"
+Faye::WebSocket.load_adapter("thin")
+
 module FastlaneCI
   # Controller for providing all data relating to builds
   class BuildJSONController < APIController
     HOME = "/data/projects/:project_id/build"
 
-    use(FastlaneCI::BuildWebsocketBackend)
```
snatchev: We add a controller method in line 114 for the websocket connections instead of using this middleware. The implementation was significantly different, so it made sense to re-implement.

app/features/build_runner/remote_runner.rb
```diff
@@ -20,187 +28,193 @@ class RemoteRunner
...
-      fastlane_log_path = File.join(@artifacts_path, "fastlane.log")
-      @fastlane_log = Logger.new(fastlane_log_path)
-      @current_build.artifacts << Artifact.new(
-        type: "fastlane.log",
-        reference: fastlane_log_path,
-        provider: @project.artifact_provider
-      )
+      @grpc = grpc
 
-      @build_state = MicroMachine.new(:PENDING).tap do |fsm|
```
snatchev: Removing the state machine here. I feel pretty strongly that the only state should be defined in the Agent and it should be the source of truth about the state of the build.

The RemoteRunner will accept any state transition and pass it along to the subscribers.

app/features/build_runner/remote_runner.rb
```diff
@@ -20,187 +28,193 @@ class RemoteRunner
...
-        end
+    # subscribe will all send all historic data to the subscriber.
+    # the yielded object will be an InvocationResponse object
+    def subscribe(&block)
+      logger.debug("subscribing listener to topic `#{topic_name}`")
+      subscriber = FastlaneCI::Notifications.subscribe(topic_name, &block)
+
+      replay_history_to(subscriber)
+
+      # make sure we call completion blocks if a subscriber joins after we have completed.
```
bogdanbrato: Just wondering if we shouldn't block the subscribe once we are done. What are the cases when one would subscribe after completion and also, what is the lifetime of the runner - do we keep it around after is done?

bogdanbrato (commented):
Good job Stefan refactoring this - just have a couple of comments and questions.

app/features-json/build_json_controller.rb
```diff
@@ -109,6 +111,50 @@ class BuildJSONController < APIController
...
+        )
+
+        if current_build_runner.nil?
+          ws.close(1000, "no runner found for project #{project_id} and build #{build_number}.")
+          next
+        end
+
+        # subscribe the current socket to events from the remote_runner
+        # as soon as a subscriber is returned, they will receive all historical items as well.
+        @subscriber = current_build_runner.subscribe do |_topic, payload|
```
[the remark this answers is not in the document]
snatchev: That's a good point. It may be possible that something else enqueues a `FastlaneBuildRunner` for the particular project and build number.

app/features/build_runner/remote_runner.rb
```diff
@@ -20,187 +28,193 @@ class RemoteRunner
...
 
+      env = environment_variables_for_worker
       start_time = Time.now.utc
+
       begin
-        responses = Agent::Client.new("localhost").request_run_fastlane(
-          "bundle", "exec", "fastlane", project.platform, project.lane, env: env
-        )
+        responses = @grpc.request_run_fastlane("bundle", "exec", "fastlane", project.platform, project.lane, env: env)
+        return if responses.nil?
```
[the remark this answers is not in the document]
snatchev: Good catch 👍
I added this to support a test-case (when stubbing `request_run_fastlane` and not providing a return value) but I don't think we should have code that's designed for handling stubbed responses. Will remove.

app/features/build_runner/remote_runner.rb
```diff
@@ -20,187 +28,193 @@ class RemoteRunner
...
+        handle_error(FastlaneCI::Proto::InvocationResponse::Error.new(description: "The Agent is not available"))
+      ensure
+        @complete = true
+        @completion_blocks.each(&:call)
       end
-      emit_new_row(type: :last_message, message: nil, time: Time.now)
 
-      current_build.duration = Time.now.utc - start_time
-      if @build_state.state == :SUCCEEDED
+      persist_history!
```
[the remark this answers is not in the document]
snatchev: > This doesn't happen if the agent dies - and I believe we loose the logs we got that far in that case.

We can move it to the `ensure` block, and I believe that will get called if the agent dies during the `begin/rescue/ensure` block.

> Side note - I feel like we should save log lines to runner.log everytime we get one, rather than at the end.

That's how it was originally, but then noticed that in those cases, we would only persist logs, and would not get state or errors.

app/features/build_runner/remote_runner.rb
snatchev: That makes _a lot_ of sense 😄

The lifetime is not defined. We don't explicitly remove the runner, so it lives until the process is restarted. Perhaps we should explicitly removed it.

bogdanbrato (approved):
ltgm

snatchev:
blocking this from merging, as to not break the angular front-end. Will need to update that to accept the new log line formats.

snatchev:
Closing, this PR is superseded by: #1093

PR #1081, opened 2018-07-13
Fix not building the right sha

KrauseFx:
Seems like this was debug code, instead we want to trigger the specific sha we got as a parameter of the request

### DIFF
--- a/app/features-json/build_json_controller.rb
+++ b/app/features-json/build_json_controller.rb
@@ -59,8 +59,6 @@ module FastlaneCI
         return
       end
 
-      branch_to_trigger = "master" # TODO: how/where do we get the default branch
-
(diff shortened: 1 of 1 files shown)


app/features-json/build_json_controller.rb
```diff
@@ -59,8 +59,6 @@ class BuildJSONController < APIController
         return
       end
 
-      branch_to_trigger = "master" # TODO: how/where do we get the default branch
-
       git_fork_config = GitForkConfig.new(
         sha: current_sha,
         branch: branch_to_trigger,
```
AndrewMcBurney: It looks like we're not defining `branch_to_trigger` anywhere else, so this `GitForkConfig` would be initialized with a `nil` `branch:` parameter.

PR #1083, opened 2018-07-13
Fix not building the right sha

KrauseFx:
Seems like this was debug code, instead we want to trigger the specific sha we got as a parameter of the request.

From what I understand, we don't need to pass the branch anyway, as we have a sha, which basically is more precisely than a branch anyway

Taken from https://github.com/fastlane/ci/pull/1081

### DIFF
--- a/app/features-json/build_json_controller.rb
+++ b/app/features-json/build_json_controller.rb
@@ -59,11 +59,8 @@ module FastlaneCI
         return
       end
 
-      branch_to_trigger = "master" # TODO: how/where do we get the default branch
-
       git_fork_config = GitForkConfig.new(
         sha: current_sha,
-        branch: branch_to_trigger,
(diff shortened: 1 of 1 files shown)


app/features-json/build_json_controller.rb
```diff
@@ -59,11 +59,8 @@ class BuildJSONController < APIController
         return
       end
 
-      branch_to_trigger = "master" # TODO: how/where do we get the default branch
-
       git_fork_config = GitForkConfig.new(
         sha: current_sha,
-        branch: branch_to_trigger,
```
taquitos: deleting this might be the right choice, but I haven't checked to see if we use this `.branch` attribute from the `git_fork_config` object anywhere.
Did you double check that?

KrauseFx:
Yeah, this will need more re-evaluating, closing this now 👍

PR #1088, opened 2018-07-13
add Procfile for foreman

snatchev:
Now that fastlane.ci requires multiple processes to run, we can take advantage of tools like https://github.com/ddollar/foreman to start or manage the services for us.

Not putting `foreman` in the Gemfile because @ddollar says so.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,2 @@
+web: bundle exec rake prod
+worker: bundle exec ruby agent/service.rb


taquitos:
@snatchev so is the idea that we need to install foreman outside of the gemfile?

snatchev:
Yes, that's the author's recommendation.

Procfile
```diff
@@ -0,0 +1,2 @@
+web: bundle exec rake prod
```
nakhbari: is web the right term? Isn't this more of a `host`? I know we're avoiding the term `master`

Issue #1089, opened 2018-07-16
update websocket client for builds.


snatchev:
Unblocks: #1071

Update the frontend websocket request and response handlers. Because we don't want to support two code paths, based on what kind of BuildRunner is performing the action, lets just migrate all BuildRunner invocations to `RemoteRunner`

- [x] websocket request must use authentication to establish the connection
- [x] websocket request should use the new URL, which is not a rack middleware.
- [x] must change the response handling. websocket responses are now JSON-encoded messages of type: https://github.com/fastlane/ci/blob/master/protos/agent.proto#L54
- [x] migrate existing `FastlaneBuildRunner` to `RemoteRunner`.

KrauseFx:
What milestone should this be of? Beta?

snatchev:
@KrauseFx yes. Milestone added.

PR #1093, opened 2018-07-20
Upgrade all FastlaneBuildRunner -> RemoteRunner

snatchev:
Fixes: #1089


app/features-json/api_controller.rb
```diff
@@ -98,6 +98,12 @@ def jwt
         authorization = request.env["HTTP_AUTHORIZATION"]
         bearer_token = authorization && authorization.slice(7..-1) # strip off the `Bearer `
 
+        # give the option to pass the bearer token as a query param.
+        # this is used when making websocket connections which do not allow us to set http headers
+        if bearer_token.nil?
```
KrauseFx: If we check for `nil` you can use `bearer_token ||= something`

app/features-json/build_json_controller.rb
```diff
@@ -191,7 +190,9 @@ def current_project
 
     # convert .log files that include the color information as ANSI code to plain text
     def convert_ansi_to_plain_text(data)
-      return data.gsub(/\e\[[0-9;]*m/, "")
+      data.scan(ANSI_PATTERN).inject("") do |str, match|
+        str << (match[3] || match[4])
```
KrauseFx: Could this potentially crash? If so, let's make sure to `rescue` the exception and fallback to plain text

app/features/build_runner/remote_runner.rb
```diff
@@ -48,6 +49,7 @@ def initialize(project:, git_fork_config:, trigger:, github_service:, grpc: Agen
 
     # TODO: consider the implications for users of this method that make the assumption
     # that the build status has been persisted or not. `start` is an async operation via the build_runner_service
+    # TODO(snatchev): pass the sha of the required checkout from the git_fork_config
```
KrauseFx: Let's also create a GH issue for this and link it here

app/services/build_runner_service.rb
```diff
@@ -44,7 +44,7 @@ def build_runner_task_queue
     def find_build_runner(project_id:, sha: nil, build_number: nil)
       return build_runners.find do |build_runner|
         build_runner.project.id == project_id &&
-        (build_runner.sha == sha || build_runner.current_build.number == build_number)
+        (build_runner.current_build.sha == sha || build_runner.current_build.number == build_number)
```
KrauseFx: Why access `.current_build` directly? Do you think it's the better approach? If so, should we remove the `.sha` helper method in the other build_runner?

KrauseFx (approved):
This is so great 👍

app/features-json/api_controller.rb
AndrewMcBurney: We could write it as this and get rid of the `if` condition here:

```ruby
bearer_token = authorization&.slice(7..-1) || request.params["bearer_token"]
```

app/services/build_runner_service.rb
snatchev: This was a bug. I don't believe we set the sha on the build_runner, when we `prepare_build_object`. This caused `find_build_runner` to match against ANY build runner with the same `project_id` only (since `build_runner.sha == nil` and `sha == nil`)

web/app/build/build.component.ts
```diff
@@ -65,7 +56,10 @@ export class BuildComponent implements OnInit, OnDestroy {
         this.buildLogSocketService.connect(projectId, buildNumber)
             .subscribe((message) => {
               // TODO: define a log line model.
-              this.logs.push(JSON.parse(message.data));
+              const response = JSON.parse(message.data);
+              if(response.log) {
```
AndrewMcBurney: This is a huge `nit`, but could you please add a space after the `if`?

app/services/build_runner_service.rb
snatchev: You'll notice that `sha` is an attribute on `BuildRunner` independent of the Build. It was possible for them to by out of sync.

app/features-json/api_controller.rb
KrauseFx: It would make the code harder to read though

AndrewMcBurney: I think this is easier to read personally, since you're only setting the `bearer_token` in one place in the code.

```ruby
bearer_token = authorization && authorization.slice(7..-1)

if bearer_token.nil?
  bearer_token = request.params["bearer_token"]
end
```
> Current Implementation

___

```ruby
bearer_token = authorization&.slice(7..-1) || request.params["bearer_token"]
```
> Suggested Implementation

___

The way I read this is: slice the `authorization` if it exists, else `request.params["bearer_token"]`.

Is this what you're suggesting as more readable alternative?

```ruby
bearer_token = authorization && authorization.slice(7..-1)
bearer_token ||= request.params["bearer_token"]
```

web/app/services/build-log-websocket.service.ts
```diff
@@ -36,7 +40,7 @@ export class BuildLogWebsocketService {
   }
 
   private createSocketUrl(projectId: string, buildNumber: number) {
-    return `${this.API_ROOT}?project_id=${projectId}&build_number=${
-        buildNumber}`;
+    const token = this.authService.token();
```
nakhbari: Note: We're going to be removing our own login and be using GitHub's authentication. So maybe here we can send GitHub's token and ensure they have access to the Build's target repo.

I'll take this as a note while working on the GitHub OAuth stuff.

web/app/services/build-log-websocket.service.spec.ts
```diff
@@ -12,6 +13,7 @@ describe('BuildLogWebsocketService', () => {
       providers: [
         BuildLogWebsocketService,
         {provide: DOCUMENT, useValue: {location: {host: 'host'}}},
+        {provide: AuthService, useValue: new AuthService(null)},
```
nakhbari: It's best if you mock out the `.token` method of the authService and use that as the value instead.

ex. https://github.com/fastlane/ci/blob/master/web/app/login/login.component.spec.ts#L28

PR #1099, opened 2018-07-23
Add OAuth Section to the Onboarding

nakhbari:
Issue: #784 

Note: i haven't been able to run rspec on this. @snatchev might need you help with the tests at some point.

#### Added Tests
image

#### Screenshot
image

```
```

### DIFF
--- a/app/features-json/setup_json_controller.rb
+++ b/app/features-json/setup_json_controller.rb
@@ -53,7 +53,7 @@ module FastlaneCI
         )
       end
 
-      unless FastlaneCI.dot_keys.oauth_client_id.nil? || FastlaneCI.dot_keys.oauth_client_secret.nil?
+      if !FastlaneCI.dot_keys.oauth_client_id.nil? && !FastlaneCI.dot_keys.oauth_client_secret.nil?
         json_error!(
           error_message: "The OAuth client has already been set",
           error_key: "OAuth.AlreadyExists",
--- a/app/features-json/setup_json_controller.rb
+++ b/app/features-json/setup_json_controller.rb
@@ -40,5 +40,33 @@ module FastlaneCI
         }
       )
     end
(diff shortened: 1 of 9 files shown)


app/features-json/setup_json_controller.rb
```diff
@@ -40,5 +40,33 @@ class SetupJSONController < APIController
...
       )
     end
+
+    post "#{HOME}/oauth", authenticate: false do
+      client_id = params["client_id"]
+      client_secret = params["client_secret"]
+
+      if client_id.nil? || client_secret.nil?
+        json_error!(
+          error_message: "Must provide both the OAuth client ID and client secret",
```
KrauseFx: Ohh, sooo to avoid having to create a personal access token, we now ask people to create their own GitHub OAuth app? And we can't create one for everyone, as the callback URL won't work?

app/features-json/setup_json_controller.rb
```diff
@@ -40,5 +40,33 @@ class SetupJSONController < APIController
...
+          error_key: "OAuth.Missing",
+          error_code: 400
+        )
+      end
+
+      unless FastlaneCI.dot_keys.oauth_client_id.nil? || FastlaneCI.dot_keys.oauth_client_secret.nil?
```
KrauseFx: I'd try avoiding `unless` together with `||` and similar, as it's confusing when it comes to priorities, can we re-org this to make it easy to read, including non Ruby devs

app/features-json/setup_json_controller.rb
nakhbari: >we now ask people to create their own GitHub OAuth app? 

yea!

>we can't create one for everyone, as the callback URL won't work?

the callback url won't work and the user won't own the oauth app.

PR #1100, opened 2018-07-24
Include Roboto font and font-family fallback

vjanssens:
- [ ] If there is an existing issue, make sure to add `Fixes ...` as part of the PR body to reference the issue you're solving

Roboto does not seem to be included by default in Angular Material, as seen in the image below and [here](https://material.angular.io/guide/typography#usage).

without-roboto

Included Roboto font from Google Fonts and properly set-up fallbacks if Google Fonts stopped working, resulting in:

with-roboto

**Alternative** would be to remove all references to Roboto Font and just use the Google Sans font everywhere.

### DIFF
--- a/web/global/theme.scss
+++ b/web/global/theme.scss
@@ -1,7 +1,6 @@
 /* You can add global styles to this file, and also import other style files */
 
 @import '~@angular/material/prebuilt-themes/deeppurple-amber.css';
-@import url('https://fonts.googleapis.com/css?family=Roboto:300,400,500');
 
 // TODO: using theme functions to set the themes instead of overriding manually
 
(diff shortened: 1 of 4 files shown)


web/global/theme.scss
```diff
@@ -1,8 +1,10 @@
 /* You can add global styles to this file, and also import other style files */
 
 @import '~@angular/material/prebuilt-themes/deeppurple-amber.css';
+@import url('https://fonts.googleapis.com/css?family=Roboto:300,400,500');
```
nakhbari: Is this recommended in comparison to having a link in the `index.html`?

vjanssens: There seems to be a difference in the way a link-element loads the font, because @import will block parallel downloads ([source](https://stackoverflow.com/a/7199377/1323800)). I'll see if I can change it up and load it from index.html.

PR #1102, opened 2018-07-25
use Procfile for development as well

snatchev:
This PR adds a separate Procfile for development. The workflow `rake dev` is updated to do the right thing. This is required because now we have 2 processes required for usage.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,9 +1,15 @@
 task :dev do
-  sh "bundle exec rackup --host 0.0.0.0 --port 8080 --env development"
+  unless system("which foreman", out: File::NULL)
+    raise "`foreman` is now required for process management. `sudo gem install foreman`"
+  end
+  sh "foreman start -f Procfile.dev"
 end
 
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -1,9 +1,15 @@
 task :dev do
-  sh "bundle exec rackup --host 0.0.0.0 --port 8080 --env development"
+  unless system("which foreman", out: File::NULL)
+    raise "`foreman` is now required for process management. `sudo gem install foreman`"
```
KrauseFx: Any reason why we recommend `sudo gem install` over `bundle install`?

snatchev: `foreman` doesn't (and shouldn't according to the foreman Readme, and https://github.com/ddollar/foreman/pull/678#issuecomment-398211757) live in the Gemfile.

KrauseFx: That makes sense, let's add a comment here with a reference to the GH issue

PR #1077, opened 2018-07-12
Details button now links to JSON URL

KrauseFx:
Fixes https://github.com/fastlane/ci/issues/1075

PR #1106, opened 2018-07-26
update the build_url thats associated with the build

snatchev:
Same change as https://github.com/fastlane/ci/pull/1077/files but now its on the `RemoteRunner` too.


KrauseFx:
Let's ship this 🚀

